PR #44, opened 2021-05-08
fix/issue-26: Backtracking change for allowing QA process because we were not allowing experiment creation unless the point/ID were valid

ppratikcr7:
We merged it but then realized that it will slow down the QA process because we were not allowing experiment creation unless the point/ID were valid. So we are raising this backtracking PR which will just comment out one line in the frontend that does the validation. We can think about the validation and whether we want to allow experiments to go into enrolling mode with invalid points and IDs etc. later


SritterCL:
On 12/4, we decided that we wanted strong validation, if metadata was ever provided (see note at the end of https://docs.google.com/document/d/13SvJajnCcH0cqMD7gPqb5MWmynzJrVqmD5a5tooPeNo/edit#) but also allowing the experiment to start with any points/IDs if no metadata was provided.

It's worth revisiting this decision, given the implementation. Maybe validation would be a good code review topic.

amurphy-cl:
@ppratikcr7 Is this still an issue with Playpower's QA process?

Issue #29, opened 2021-01-26
Log errors, warnings and info


SritterCL:
We should distinguish three different types of logging.

An Error means that UpGrade is unable to continue normally. For example, if an experiment requires assignment by group and no group has been set for the user, that's an error. The user gets the "default" condition.

A Warning means that UpGrade has detected a strange situation but can continue. If a user's working group is specified but the general group membership isn't, that reflects some kind of misconfiguration, but it's clear what UpGrade should do - assign the user according to the specified working group.

An Info message is used for debugging. It allows a user to trace what's going on UpGrade. For example, in verbose mode, we should log any API calls.

amurphy-cl:
This is done.

PR #49, opened 2021-05-13
Adding error log for experiments which are exclude because of invalid group or working group information

VivekFitkariwala:


VivekFitkariwala:
For issue #29 
Added error log when group/working group is invalid for experiment

msandbothe (approved):

PR #51, opened 2021-05-13
Fix/issue 23: Decimal condition assignment weights allowed with a checkbox to assign equal weights to all conditions

ppratikcr7:
Below is the screenshot for the UI change:

Screenshot 2021-05-13 at 4 39 13 PM


msandbothe (approved):

VivekFitkariwala (dismissed):

VivekFitkariwala:
Experiment need to be added again.

nirmalpatel:
@mfugate1 this PR is going to make a DB schema change (condition weight type from int to float)

We are not merging it now as we have not tested the data migration piece. This commit is modifying the type of an existing column, rather than adding a new column. @VivekFitkariwala do you know how the existing migration script will behave in this case? Maybe all of the ints get cast as floats...

msandbothe (dismissed):
Change looks good, but hold off until we know the ramifications of the migration.

amurphy-cl:
@msandbothe can we merge this one?

msandbothe:
@nirmalpatel or another Playpower dev Can you give a comment on whether we can or cannot merge this? It's unclear whether this change will function properly with migration.

ppratikcr7:
@msandbothe We shouldn't merge this PR as of now due NOT NULL constraint on the field and migration would generate NULL values. We can merge it when we deploy fresh DB on production.

danoswaltCL (approved):

PR #64, opened 2021-06-03
Fix/issue 53 Condition code as 'default' not allowed from frontend and backend

ppratikcr7:
In this PR I have made changes in the frontend code to not allow users to enter conditionCode as 'default' (case-insensitive) whether they create a new experiment or update an existing experiment and they try to move forward clicking on "Next" or "Save" button. Also, if someone tries to create a new experiment or update an existing experiment from backend api calls with conditionCode as 'default' (case-insensitive), they would get a 400 BadRequestError with error message.

Following validation message would be shown on frontend:

Screenshot 2021-06-03 at 4 23 04 PM
Screenshot 2021-06-03 at 4 23 31 PM

Following is the 400 BadRequestError with error message during backend api call for creating and updating an experiment:

Screenshot 2021-06-03 at 4 19 01 PM
Screenshot 2021-06-03 at 4 19 17 PM


ppratikcr7:
@VivekFitkariwala @JD2455 Can anyone of you just do the code review.
@msandbothe Meanwhile you can verify and test the change at your end.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -445,19 +446,31 @@ export class ExperimentService {
         } catch (error) {
           throw new Error(`Error in updating experiment document "updateExperimentInDB" ${error}`);
         }
-
+        let defaultconditionCodeError = false;
```
VivekFitkariwala, line 449: This variable is not required. Throw the error inside if condition

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -727,6 +740,23 @@ export class ExperimentService {
         uniqueIdentifiers = response[1];
       }
       const { conditions, partitions, ...expDoc } = experiment;
+      
+      let defaultconditionCodeError = false;
```
VivekFitkariwala, line 744: Same as above

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Okay will update code.

SavvasJaydipHirapara: Wouldn't it be better to make a separate private function which is responsible for handling this part only?

```private hasConditionCodeDefault(conditions: Conditions[]): boolean {}```

If the function returns true, throw the error, otherwise proceed with further logic.

cc: @VivekFitkariwala

File: frontend/projects/abtesting/src/assets/i18n/en.json [hunk 3]
```diff
@@ -96,6 +97,7 @@
   "home.new-experiment.design.add-experiment-partition.text": "Add Experiment Site",
   "home.new-experiment.design.assignment-weight-validation.text": "Assignment weights' sum must be 100",
   "home.new-experiment.design.condition-unique-validation.text": "Condition should be unique",
+  "home.new-experiment.design.condition-name-validation.text": "Condition name cannot be 'default'",
```
SavvasJaydipHirapara, line 100: Shouldn't message be like **Condition name can not be 'default'**. Space between cannot.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
JD2455: Wouldn't it be better to make a separate private function which is responsible for handling this part only?

private hasConditionCodeDefault(conditions: Conditions[]): boolean {}

If the function returns true, throw the error, otherwise proceed with further logic.

cc: @VivekFitkariwala

frontend/projects/abtesting/src/assets/i18n/en.json
JD2455, line 100: I think we can have a space between _cannot_ in the message

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -243,11 +243,20 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   validateConditionCodes(conditions: ExperimentCondition[]) {
+    let defaultKeywordError = false;
     const conditionCodes = conditions.map(condition => condition.conditionCode);
+    conditionCodes.forEach(conditioncode => {
+      if (conditioncode.toUpperCase() === this.translate.instant('home.new-experiment.design.condition.invalid.text')) {
```
JD2455, line 249: It seems two different validations are going on in this function. Can we have a separate method for the validation for default keyword? may be hasConditionCodeDefault()?

Another thing is instead of translating the text in each iteration, take this out from the forEach loop.

frontend/projects/abtesting/src/assets/i18n/en.json
ppratikcr7: ok

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: ok

ppratikcr7:
@JD2455 @VivekFitkariwala I have made all the necessary changes to this PR. Please check and we can merge this.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -405,6 +406,20 @@ export class ExperimentService {
     }
   }
 
+  private hasConditionCodeDefault(conditions: ExperimentCondition[]): any {
+    // Check for conditionCode is 'default' then return error:
+    const hasDefaultConditionCode = conditions.filter(
+      condition => condition.conditionCode.toUpperCase() === 'DEFAULT'
+    );
+    if (!!hasDefaultConditionCode.length) {
```
JD2455, line 414: Don't we need just a single ! instead of !!?

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -243,11 +243,24 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
-      this.conditionCodeError = this.translate.instant('home.new-experiment.design.condition-unique-validation.text')
-    } else {
-      this.conditionCodeError = null;
+      this.conditionCodeError = conditionUniqueErrorText;
+    }  else {
+        this.conditionCodeError = null;
+    }
+  }
+
+  validatehasConditionCodeDefault(conditions: ExperimentCondition[]) {
```
JD2455, line 256: method name should be camelCase. Rename it to validateHasConditionCodeDefault.

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -243,11 +243,24 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
+    let defaultKeyword = this.translate.instant('home.new-experiment.design.condition.invalid.text');
+    let defaultConditionCodeErrorText = this.translate.instant('home.new-experiment.design.condition-name-validation.text')
+    if (conditions.length >= 1 ) {
+      const conditionCodes = conditions.map(condition => condition.conditionCode);
+      conditionCodes.forEach(conditioncode => {
+        conditioncode && conditioncode.toUpperCase() === defaultKeyword ? this.conditionCodeError = defaultConditionCodeErrorText : this.conditionCodeError = null;
```
JD2455, line 262: This logic seems wrong. Can we reverify this?

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: okay

backend/packages/Upgrade/src/api/services/ExperimentService.ts
VivekFitkariwala: I agree. This is better way to do.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: `!` returns true if the object is null, undefined or an empty string and false otherwise. If you negate it again using `!!` you get true for values that exist and false for the ones that do not

ppratikcr7: I feel this is fine as when we have a condition filtered with 'default' name, then `!!` this would allow it to enter if condition and throw badrequesterror

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: You are correct. Nice catch. I did take care of this in the backend check but forgot the same in frontend one. Now it is handled properly.
Screenshot 2021-06-17 at 7 01 34 PM

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Done accordingly

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: Done now

frontend/projects/abtesting/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: Done

backend/packages/Upgrade/src/api/services/ExperimentService.ts
JD2455: @ppratikcr7 Can you check this logic again by keeping only `hasDefaultConditionCode.length` only because filter function will return back the array always and if you have anything in the array then it will throw error. So it should work by removing !!.

ppratikcr7: okay let me check

ppratikcr7: Code changes done

JD2455 (approved):

msandbothe (approved):

VivekFitkariwala (approved):

PR #103, opened 2021-07-16
[DO NOT MERGE] Fix/issue-69: not creating user if doesn't exist except init call

YashilDepani:
In this PR, I have removed user creation if the user doesn't exist and now instead error message: `user not defined` is thrown. The user can only be created via `/init` call. The following calls are now devoid of user creation( if not exist).

1.  /mark,
2. /assign,
3. /useraliases,
4. /datalog,
5.  /bloblog, 
6. /failed,
7. /groupmembership,
8. /workinggroup

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -292,20 +292,19 @@ export class ExperimentClientController {
    *            description: Log blob data
    */
   @Post('bloblog')
-  public blobLog(@Req() request: express.Request): any {
+  public blobLog(@Req() request: express.Request): Promise<Log[] | any> {
     return new Promise((resolve, reject) => {
       request.on('readable', async (data) => {
         const blobData = JSON.parse(request.read());
         try {
-          // The function will throw error is iserId doesn't exist
+          // The function will throw error if userId doesn't exist
           const response = await this.experimentAssignmentService.blobDataLog(blobData.userId, blobData.value);
           resolve(response);
-        }
-        catch(error) {
+        } catch (error) {
           // The error is rejected so promise can now handle this error
-          reject(error); 
+          reject(error);
         }
-      })
+      });
     }).catch(error => {
       throw new Error(
         JSON.stringify({
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -7,7 +7,7 @@ import { ExperimentUser } from '../models/ExperimentUser';
 import { ExperimentUserService } from '../services/ExperimentUserService';
 import { UpdateWorkingGroupValidator } from './validators/UpdateWorkingGroupValidator';
 import { MonitoredExperimentPoint } from '../models/MonitoredExperimentPoint';
-import { IExperimentAssignment, ISingleMetric, IGroupMetric } from 'upgrade_types';
+import { IExperimentAssignment, ISingleMetric, IGroupMetric, SERVER_ERROR} from 'upgrade_types';
 import { FailedParamsValidator } from './validators/FailedParamsValidator';
 import { ExperimentError } from '../models/ExperimentError';
 import { FeatureFlag } from '../models/FeatureFlag';
@@ -292,13 +292,27 @@ export class ExperimentClientController {
    *            description: Log blob data
    */
   @Post('bloblog')
-  public blobLog(@Req() request: express.Request): Promise<Log[]> {
-    return new Promise((resolve) => {
+  public blobLog(@Req() request: express.Request): any {
+    return new Promise((resolve, reject) => {
       request.on('readable', async (data) => {
         const blobData = JSON.parse(request.read());
-        const response = await this.experimentAssignmentService.blobDataLog(blobData.userId, blobData.value);
-        resolve(response);
-      });
+        try {
+          // The function will throw error is iserId doesn't exist
+          const response = await this.experimentAssignmentService.blobDataLog(blobData.userId, blobData.value);
+          resolve(response);
+        }
+        catch(error) {
+          // The error is rejected so promise can now handle this error
+          reject(error); 
+        }
+      })
+    }).catch(error => {
+      throw new Error(
+        JSON.stringify({
+          type: SERVER_ERROR.EXPERIMENT_USER_NOT_DEFINED,
+          message: error.message,
+        })
+      );
     });
   }
 
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -22,7 +22,6 @@ import { IndividualAssignment } from '../models/IndividualAssignment';
 import { GroupAssignment } from '../models/GroupAssignment';
 import { IndividualExclusion } from '../models/IndividualExclusion';
 import { GroupExclusion } from '../models/GroupExclusion';
-import { ExperimentUserRepository } from '../repositories/ExperimentUserRepository';
 import { Experiment } from '../models/Experiment';
 import { ExplicitIndividualExclusionRepository } from '../repositories/ExplicitIndividualExclusionRepository';
 import { ExplicitGroupExclusionRepository } from '../repositories/ExplicitGroupExclusionRepository';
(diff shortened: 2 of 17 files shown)


backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -292,13 +292,27 @@ export class ExperimentClientController {
    *            description: Log blob data
    */
   @Post('bloblog')
-  public blobLog(@Req() request: express.Request): Promise<Log[]> {
-    return new Promise((resolve) => {
+  public blobLog(@Req() request: express.Request): any {
```
JD2455, line 295: @YashilDepani We should not use any if we already know the type of it which was `Log` previously

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -292,13 +292,27 @@ export class ExperimentClientController {
...
+    return new Promise((resolve, reject) => {
       request.on('readable', async (data) => {
         const blobData = JSON.parse(request.read());
-        const response = await this.experimentAssignmentService.blobDataLog(blobData.userId, blobData.value);
-        resolve(response);
-      });
+        try {
+          // The function will throw error is iserId doesn't exist
```
JD2455, line 300: Can you please update the comment by fixing the typo?

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -513,9 +515,9 @@ export class ExperimentAssignmentService {
     let userDoc = await this.experimentUserService.getOriginalUserDoc(userId);
     const keyUniqueArray = [];
 
-    // create user if user does not exist
+    // throw error if user not defined
     if (!userDoc) {
-      userDoc = await this.userRepository.save({ id: userId });
+      throw new Error(`User not defined: ${userId}`);
```
JD2455, line 520: @YashilDepani  Shouldn't we throw the same error for all apis? In the `getAllExperimentConditions()`, we are throwing with `type as EXPERIMENT_USER_NOT_DEFINED`.

JD2455:
@YashilDepani I can still see the user is being created from the below 2 calls
```
1. /groupmembership
2. /workinggroup
```

Are we going to keep user creation for this calls?
We also need to add `/init` API in the client library to call backend.

CC: @VivekFitkariwala  @nirmalpatel

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
YashilDepani: If I use any then the code is not allowing me to add `.catch ()` for the promise rejection

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
YashilDepani: Sorry about this, will fix it.

YashilDepani:
I am not sure about these two calls, can anyone let me know what to do for these calls?
1. groupmembership
2. workinggroup

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
JD2455: @YashilDepani Can you look more into this to fix this?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
JD2455: thanks!

JD2455:
> I am not sure about these two calls, can anyone let me know what to do for these calls?
> 
> 1. groupmembership
> 2. workinggroup

This 2 calls are also from the client library and as per discussion, we will allow to create user from the `/init` call only and hence we should throw error in this 2 APIs also. Here is the link of all APIs being called from the library - https://github.com/CarnegieLearningWeb/UpGrade/blob/main/clientlibs/js/src/UpgradeClient.ts

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
YashilDepani: I was not able to find anything related to this. Just had a thought of doing `Promise<Log[] | any>`.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
YashilDepani: The same type of error is also thrown by this /bloblog. Reference code: 
`throw new Error(
        JSON.stringify({
          type: SERVER_ERROR.EXPERIMENT_USER_NOT_DEFINED,
          message: error.message,
        })
      );`

JD2455: @YashilDepani Please update the code here to send the same error which we are throwing in other client APIs

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -61,6 +64,11 @@ export default class UpgradeClient {
         }
     }
 
+    async init(userId: string, group?: Map<string, Array<string>>, workingGroup?: Map<string, string>): Promise<Interfaces.IUser> {
```
JD2455, line 67: @YashilDepani I think we shouldn't anticipate userId in the argument because we already have that at the time of initializing  `UpgradeClient` class. We are also validating it in the function `this.validationClient` that if the userId exist or not so IMO, we can remove it from the argument.

@VivekFitkariwala  @nirmalpatel What are your thoughts here?

clientlibs/js/src/functions/init.ts
```diff
@@ -0,0 +1,52 @@
...
+        workingGroupObj
+      }
+    }
+
+    const response = await fetchDataService(url, token, data, Types.REQUEST_TYPES.POST);
+    if (response.status) {
+      return {
+        id: userId,
+        group,
+        workingGroup
```
JD2455, line 44: @YashilDepani  Can we use the group and workingGroup which are getting in the response to forward it to caller of this function? I think that would be better instead of sending what we got in the argument.

clientlibs/js/src/UpgradeClient.ts
YashilDepani: Yes, I got that, so I instead use instantiated user id (this.userId) right?

JD2455: Right. But also need opinion of @VivekFitkariwala or @nirmalpatel here

nirmalpatel:
@msandbothe there are 3 things combined in this PR:

- only init() creates new users in UpGrade
- JS client supports init()
- Java client supports init()

JD2455:
@YashilDepani  I think You also need to bump up the JS library version.

clientlibs/js/src/UpgradeClient.ts
YashilDepani: Code updated

clientlibs/js/src/functions/init.ts
YashilDepani: I am not able to use group and workingGroup of response as the function return expects IUSER interface and getting error as non assignable.

msandbothe (approved):

VivekFitkariwala (approved):

clientlibs/js/src/functions/init.ts
YashilDepani: I have updated the code.

JD2455 (approved):

amurphy-cl:
No current version of MATHia makes an `init` call, so we may not want to merge this now. @kschaefe can you let us know the deadline for the next Portal release? We would normally ask @jerith666 to make the change of adding the `init` call but he is on vacay until ~Aug13.

kschaefe:
It shouldn't matter if we merge this as long as the currently-used client library versions will continue to work (or at least work as well as they did prior to these changes).

If these changes are going to force the clients to update, then we can only merge them, if we don't release and use that version of the server. 

Can anyone tell me what the client responsibilities are for these changes?

msandbothe:
This is a major change that will cause our implementation of the client library to return an error response. We currently don't call init() so all calls on new users should fail (which I hope we handle gracefully). The necessary change is to call init() in both portal and launcher before calling setGroup/setWorkingGroup.

At first I thought we just needed to call init() from portal, but we should also call init() for users who launch without portal. The init() call in launcher is lower priority than the one in portal.

jerith666:
I've entered https://carnegielearning.atlassian.net/browse/PORTAL-620 to have portal make the `init()` call.  I think we have to merge only the client API change first, though, so that it's available to make the portal update, before the server starts requiring it.

nirmalpatel:
@jerith666 @msandbothe

@YashilDepani created two new PRs. One for SDK changes and one for Backend changes.

SDK changes #124 
Backend changes #125

PR #108, opened 2021-08-01
Removing extra query in analysis function

VivekFitkariwala:
Removing extra query in analysis function
Changing console.log to logger


backend/packages/Upgrade/src/env.ts
```diff
@@ -54,6 +54,7 @@ export const env = {
     database: getOsEnvOptional('TYPEORM_DATABASE') || getOsEnvOptional('RDS_DB_NAME'),
     synchronize: toBool(getOsEnvOptional('TYPEORM_SYNCHRONIZE')),
     logging: getOsEnv('TYPEORM_LOGGING'),
+    maxQueryExecutionTime: toNumber(getOsEnvOptional('TYPEORM_MAX_QUERY_EXECUTION_TIME')),
```
JD2455, line 57: @VivekFitkariwala Can we add it to a `.env.example` or `.env.test` file so that it doesn't get missed?

msandbothe (approved):

PR #127, opened 2021-08-25
Added sample responses in swagger (such as 200, 404, 500) and created API documentation

YashilDepani:
In this PR,  I have added sample responses for Experiment and Experiment-client APIs. These includes responses such as 200, 404, 500.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/AnalyticsController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/AnalyticsController.ts
@@ -43,6 +43,25 @@ export class AnalyticsController {
    *       responses:
    *          '200':
    *            description: Analytics For Experiment Enrollment
+   *            schema:
+   *              type: array
+   *              description: ''
+   *              minItems: 1
+   *              uniqueItems: true
+   *              items:
+   *                type: object
+   *                required:
+   *                  - users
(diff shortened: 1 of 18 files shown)


RidhamShah:
Things to change:

1. `/mark` - not working in swagger
2. `/log` - giving 500 errors 
3. `/bloblog` - something is going in infinite loop
4. `/failed` - not working
5. `/feature flag` - not showing data even if there is
6. `/condition{id}` - giving 500 error instead of 404 for wrong UUID

YashilDepani:
@msandbothe as we had a discussion regarding adding test cases for cases such as 500, 404, etc. I looked into the test cases and as we are mocking the services it is not possible to generate such cases. I have added a few possible test cases giving 500 errors when the ```experiment id is not of type UUID``` as this was checked before in the code and not related to service.

RidhamShah (changes requested):
Things to change:

1. `/mark` - not working in swagger
2. `/log` - giving 500 errors
3. `/bloblog` - something is going in infinite loop
4. `/failed` - not working
5. `/feature flag` - not showing data even if there is
6. `/condition{id}` - giving 500 error instead of 404 for wrong UUID

RidhamShah (approved):

PR #138, opened 2021-09-16
Fix/issue-121 All metric queries will not fail if one or more fails

YashilDepani:
In this PR, I have decoupled the metric queries. By this we can get response of successful queries even if one or more query fails. Also we will have database entry of failed queries and id of those failed query can be seen in log.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/QueryService.ts
+++ b/backend/packages/Upgrade/src/api/services/QueryService.ts
@@ -63,7 +63,7 @@ export class QueryService {
       await Promise.all(failedQuery);
     }
 
-    modifiedResponse = modifiedResponse.filter(res => res.length).map((res, index) => {
+    modifiedResponse = modifiedResponse.map((res, index) => {
       return { id: queryIds[index], result: res };
     });
 
--- a/backend/packages/Upgrade/src/api/services/QueryService.ts
+++ b/backend/packages/Upgrade/src/api/services/QueryService.ts
@@ -6,6 +6,7 @@ import { Query } from '../models/Query';
 import { LogRepository } from '../repositories/LogRepository';
 import { SERVER_ERROR } from 'upgrade_types';
 import { ErrorService } from './ErrorService';
+import { ExperimentError } from '../models/ExperimentError';
 
 @Service()
 export class QueryService {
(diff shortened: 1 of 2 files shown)


backend/packages/Upgrade/src/api/services/QueryService.ts
```diff
@@ -34,13 +37,35 @@ export class QueryService {
...
+        return query.value;
+      } else {
+        this.log.error('Error in Query Id ', queryIds[index]);
+        failedQuery.push(this.errorService.create({
+          endPoint: '/api/query/analyse',
+          errorCode: 500,
+          message: `Query Failed error: ${JSON.stringify(queryIds[index], undefined, 2)}`,
+          name: 'Query Failed error',
+          type: SERVER_ERROR.QUERY_FAILED,
+        } as any));
```
JD2455, line 55: Can you please update the `any` with `ExperimentError`?

backend/packages/Upgrade/src/api/services/QueryService.ts
```diff
@@ -34,13 +37,35 @@ export class QueryService {
...
+      }
     });
-    return response;
+
+    if (failedQuery.length) {
+      await Promise.all(failedQuery);
+    }
+
+    modifiedResponse = modifiedResponse.map((res, index) => {
+      return queryIds[index] ? { id: queryIds[index], result: res} : null;
```
JD2455, line 66: I think. the condition `queryIds[index]` will always have value. In what can null value will be returned to map function?  Also fix the formatting issue.

backend/packages/Upgrade/src/api/services/QueryService.ts
```diff
@@ -34,13 +37,35 @@ export class QueryService {
...
+    }).filter(query => query);
```
JD2455, line 67: Update the variable name from query -> queryResult.

JD2455 (changes requested):

VivekFitkariwala (approved):

backend/packages/Upgrade/src/api/services/QueryService.ts
YashilDepani: Thanks, code updated

backend/packages/Upgrade/src/api/services/QueryService.ts
YashilDepani: modifiedResponse had empty array as response (in case the promise was rejected). I have updated the code wherein the response is first filtered so we won't have empty array.

backend/packages/Upgrade/src/api/services/QueryService.ts
YashilDepani: Thanks, code updated.

JD2455 (approved):

PR #140, opened 2021-09-17
Add Zack and Kyle as admin users

zackcl:
This commit adds Zack Lee (zlee@carnegielearning.com) and Kyle Weldon (kweldon@carnegielearning.com) as admins users.


ppratikcr7 (approved):
These changes seems fine, although we wont be using `config.json` and would be removed in next PR, still you can keep the changes in this file for now.

VivekFitkariwala (approved):

PR #142, opened 2021-09-19
Fix/issue 134

VivekFitkariwala:
- Added swc transpiler
- Removed type checking from typescript transpilation and added separate command for transpilation
- Added husky for pre commit hook for lint and type check during development
- Added type checking during build


JD2455 (approved):

VivekFitkariwala:
Husky is added at the root level of the project. So for husky to work properly you need to npm install at the root level.

frontend/package-lock.json
```diff
@@ -1,8 +1,16112 @@
 {
     "name": "ab-testing",
     "version": "1.0.52",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
```
JD2455, line 4: @VivekFitkariwala Do we really need this change as we don't have any changes in the package.json file?

msandbothe (approved):

frontend/package-lock.json
VivekFitkariwala: This is due to different npm version we are using. I think if we are npm > ^7.0.0 it uses `"lockfileVersion": 2`. I have reverted the change for now but I think we can force user to use minimum version of npm to avoid this conflict in future

jerith666:
I just got:

```
$ git commit

> ab-testing@1.0.69 test
> ng test

sh: line 1: ng: command not found
husky - pre-commit hook exited with code 127 (error)
```

I don't like husky sideloading itself into my local git config without any warning just because I ran `npm install`.  Especially since I use a multi-remote local repo so now husky is going to try to run on a bunch of other repos where it's not wanted.  :-1:

PR #160, opened 2021-10-01
fix for issue 119 jsonb null and test case added to handle null values

ppratikcr7:
cannot cast jsonb null error for accuracy metrics resolved and test case added to handle null values in metrics.


backend/packages/Upgrade/src/api/repositories/LogRepository.ts
```diff
@@ -291,6 +291,7 @@ export class LogRepository extends Repository<Log> {
   ): SelectQueryBuilder<Experiment> {
     // get experiment repository
     const experimentRepo = getRepository(Experiment);
+    const nullstring = 'null';
```
VivekFitkariwala, line 294: Is the null value in json is converted to string 'null'? So the question is what if someone uses undefined?

VivekFitkariwala (approved):

PR #171, opened 2021-11-01
Consistent init call response

ppratikcr7:
We have removed ['createdAt', 'updatedAt', 'versionNumber'] keys from the init call response as this is not expected in Java clientlib. Also, we only return the keys that are passed during the init call.

We have verified the init call response for the below cases:
1. Only userId is passed.
2. User with group and working group is passed.
3. User already exist.
4. With correct group/working group info.
5. UserId with updated group and working group info.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -131,7 +131,14 @@ export class ExperimentClientController {
   ): Promise<ExperimentUser> {
     request.logger.addFromDetails(__filename, 'init');
     request.logger.info({ stdout: 'Starting the init call for user', stack_trace: 'null' });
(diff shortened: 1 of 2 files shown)


msandbothe (approved):

ppratikcr7:
I have updated the PR with a 3 field JSON with these fields: ['userId', 'group', 'workingGroup'] No matter how we call the API, the 3 fields would be present as keys. See below init calls and their corresponding consistent response:

Screenshot 2021-11-02 at 1 12 13 PM
Screenshot 2021-11-02 at 1 12 29 PM
Screenshot 2021-11-02 at 1 12 54 PM
Screenshot 2021-11-02 at 1 13 22 PM
Screenshot 2021-11-02 at 1 13 49 PM

VivekFitkariwala (approved):

PR #174, opened 2021-11-02
add sessionId to java client lib request header

kawcl:


kawcl:
I did not add the sessionId to the constructor of the `ExperimentClient` so not to break any current implementations using it. Also because it's an optional field.  Instead I provided a setter if a user wanted to optionally set it.  If it's thought to be better to include it with the constructor that can be added.

jerith666:
You need to rev the version in the pom file.

Since clients will also need to rev the version in their pom files, I think it's fine to make breaking API changes like adding a c'tor argument.

I'd rather see the session ID as a required c'tor argument, since we believe it's best practice to provide one.  The caller can always fill it in with `UUID.randomUUID().toString()` if their environment doesn't provide anything better.  :)

kawcl:
@jerith666 you want the setter completely removed with the constructor update?

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -50,13 +50,12 @@
      *            Properties to permit users to control how the underlying JAX-RS
      *            client behaves. These are passed through to
      *            {@link javax.ws.rs.core.Configurable#property(String, Object)}. */
-	public ExperimentClient(String userId, String authToken, String baseUrl, Map<String, Object> properties) {
+	public ExperimentClient(String userId, String authToken, String sessionId, String baseUrl, Map<String, Object> properties) {
```
VivekFitkariwala, line 53: In Javascript client we have kept sessionId optional so if it is not provided explicitly we can generate a random one.

jerith666 (approved):
This looks good to me -- passing the session ID in to the `APIService` ctor is a better approach.

I prefer requiring the argument for the reasons I gave earlier, but if others feel strongly about it, I'm okay with a second ctor that doesn't require it.

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
kawcl: I added in a second constructor without the session id argument, making it optional.  If that one is used then we will generate a random one

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -55,8 +56,19 @@ public ExperimentClient(String userId, String authToken, String baseUrl, Map<Str
 			throw new IllegalArgumentException(INVALID_STUDENT_ID);
 		}
 		this.userId = userId;
+		this.apiService = new APIService(baseUrl, authToken, UUID.randomUUID().toString(), properties);
```
jerith666, line 59: you don't need to copy-paste the whole body of the constructor.  Just do:

```
    this(userId, authToken, UUID.randomUUID().toString(), baseUrl, properties);
```

jerith666 (changes requested):

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
kawcl: good catch, it's been updated

VivekFitkariwala (approved):

jerith666 (approved):

jerith666:
Looks good now, thanks!

PR #177, opened 2021-11-05
Update Java Client Lib assign response with order field

kawcl:
add the "order" field in the response object

### DIFF
--- a/clientlibs/java/pom.xml
+++ b/clientlibs/java/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.upgradeplatform</groupId>
 	<artifactId>upgrade-client</artifactId>
-	<version>0.8</version>
+	<version>0.10</version> <!-- adds the order field to the response object -->
 
 	<build>
 		<plugins>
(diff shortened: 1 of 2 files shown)


jerith666:
looks good except it needs a `pom.xml` version increment.

jerith666 (changes requested):
looks good except you need a `pom.xml` version increment

kawcl:
> looks good except you need a `pom.xml` version increment

🤦  .. is there a pattern we follow for these version increments? what happens when two or more PRs are updating to the same version? is that fine?

jerith666:
Alas, we don't have good answers to either of those questions.

We should probably rev to 1.0 at some point, and then start following https://semver.org, but I'm not sure when's the right time for that.  Though honestly, now is probably fine; semver FAQ says:

> How do I know when to release 1.0.0?
> If your software is being used in production, it should probably already be 1.0.0.

... and we are using it in production.  :)

If you put a comment (like `<version>0.10</version> <!-- 0.10 adds the order field to the response object -->`), then it'll produce a conflict unless the other branch has added that exact same comment, which is highly unlikely.  That's probably the best strategy I can recommend.

kawcl:
I updated the version and added in the a comment for now

jerith666 (approved):

jerith666:
Looks good now, thanks!

ppratikcr7 (approved):

VivekFitkariwala (approved):

PR #186, opened 2021-11-12
increment java client version

kawcl:
increment version so Jenkins builds correctly


jerith666 (approved):
that looks like a version increment to me!  ;-)

msandbothe (approved):

PR #191, opened 2021-11-17
ExperimentClient: provide more detail on ProcessingException errors

jerith666:


clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -524,4 +524,32 @@ public void failed(Throwable throwable) {
...
+    /**helper method to include more useful information in the ErrorResponse for a ProcessingException*/
+    private static <T> void readResponseToCallback(Response response, ResponseCallback<T> callback, Class<T> clazz) {
+        readResponseToCallback(response, callback, clazz, UnaryOperator.identity());
+    }
+
+    /**helper method to include more useful information in the ErrorResponse for a ProcessingException*/
+    private static <T> void readResponseToCallback(Response response, ResponseCallback<T> callback, Class<T> clazz, UnaryOperator<T> finisher) {
+	    response.bufferEntity();
+	    try {
+	        callback.onSuccess(finisher.apply(response.readEntity(clazz)));
```
kawcl, line 550: What is this goal of applying the UnaryOperator's identity function to the response entity?

kawcl: The markExperimentPoint function looks to be using a different `finisher`. Is passing the identity function just for getting around that unique case?

kawcl (approved):

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
jerith666: Yes, exactly.  Otherwise I'd have to repeat the rest of the implementation of `readResponseToCallback()`.

Indeed, I probably should've taken it one step further so that all _three_ overloads of that method could flow through one common code path.  That would've required even more functional programming-style abstraction, though, and would've made the code harder to follow.

A bit of a judgment call.  :)

kawcl: ok I just wanted to make sure I was following that correctly. 👍

Issue #194, opened 2021-11-18
Decouple data export from experiment export


msandbothe:
Data export is an expensive operation that is only useful for a handful of users. We need a way to export the experiment json and other properties without the expensive DB query.

PR #198, opened 2021-11-20
Fix for issue #194: Adding new api to decouple export of experiment

ppratikcr7:
Following are the changes done in this PR:
Adding a new api `export` which will download Experiment Design in JSON file to user's system.
Updated the existing api `stats/csv`, in which only the Monitored experiment point data (CSV file) is sent in the email.

Screenshot 2021-11-20 at 10 30 49 AM

As seen in the above image, the `Export Data` button will trigger a mail which will have s3 link to only one file that is the Monitered Experiment points data csv. The `Export Design` button will not send any mail, just download the Experiment Design JSON file on user's system.


VivekFitkariwala (approved):

nirmalpatel:
Merging so that Pratik can continue solving the Metric Queries import export.

PR #202, opened 2021-11-23
fix for issue #199: api to clear db tables in demo app

ppratikcr7:
In this PR, we have added a DELETE API (`/api/clearDB`) to truncate the db tables excluding ['user', 'metric', 'setting', 'migrations'].
For this to be protected, you need to set an environment variable named `APP_DEMO=true` in the `.env` file to work.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -1,4 +1,4 @@
-import { JsonController, Post, Body, UseBefore, Get, BodyParam, Req, InternalServerError } from 'routing-controllers';
+import { JsonController, Post, Body, UseBefore, Get, BodyParam, Req, InternalServerError, Delete } from 'routing-controllers';
 import { ExperimentService } from '../services/ExperimentService';
 import { ExperimentAssignmentService } from '../services/ExperimentAssignmentService';
 import { MarkExperimentValidator } from './validators/MarkExperimentValidator';
@@ -20,6 +20,8 @@ import { ExperimentUserAliasesValidator } from './validators/ExperimentUserAlias
 import { Metric } from '../models/Metric';
 import * as express from 'express';
 import { AppRequest } from '../../types';
+import { env } from '../../env';
+import { toBool } from '../../lib/env';
 
 /**
  * @swagger
@@ -654,5 +656,14 @@ export class ExperimentClientController {
   @Post('useraliases')
   public setUserAliases(@Body() user: ExperimentUserAliasesValidator): Promise<ExperimentUser[]> {
     return this.experimentUserService.setAliasesForUser(user.userId, user.aliases);
+  }
+
+  @Delete('clearDB')
+  public clearDB(): Promise<string> {
+    // if DEMO mode is enabled, then clear the database:
+    if(toBool(env.app.demo)) {
+      return this.experimentUserService.clearDB();
+    }
+    return Promise.resolve('DEMO mode is disabled. You cannot clear DB.');
   }
 }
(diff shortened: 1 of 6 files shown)


backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -655,4 +657,13 @@ export class ExperimentClientController {
   public setUserAliases(@Body() user: ExperimentUserAliasesValidator): Promise<ExperimentUser[]> {
     return this.experimentUserService.setAliasesForUser(user.userId, user.aliases);
   }
+
+  @Delete('clearDB')
+  public clearDB(): Promise<string> {
+    // if DEMO mode is enabled, then clear the database:
+    if(toBool(env.app.demo)) {
```
VivekFitkariwala, line 664: Move toBool operation inside env.ts

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -178,4 +178,19 @@ export class ExperimentRepository extends Repository<Experiment> {
...
+      const entities = entityManager.connection.entityMetadatas;
+      for (const entity of entities) {
+        if(!(['user', 'metric', 'setting', 'migrations'].includes(entity.tableName))) {
+          const repository = await entityManager.connection.getRepository(entity.name);
+          await repository.query(`TRUNCATE ${entity.tableName} CASCADE;`);
+        }
+      }
+      return 'DB truncate successful';
+    } catch (error) {
+      throw new Error('DB truncate error. DB truncate not allowed');
```
VivekFitkariwala, line 193: Just throw existing error with details or type
`
error.type = .....;
error.detail = ....,
throw error;
`

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -393,4 +393,11 @@ export class ExperimentUserService {
       await this.individualAssignmentRepository.deleteExperimentsForUserId(userId, assignedExperimentIds);
     }
   }
+
+  public clearDB(): Promise<string> {
+    getConnection().transaction(async (transactionalEntityManager) => {
```
VivekFitkariwala, line 398: Don't you need `await` before getConnection()?

backend/packages/Upgrade/src/env.ts
```diff
@@ -39,6 +39,7 @@ export const env = {
       middlewares: getOsPaths('MIDDLEWARES'),
       interceptors: getOsPaths('INTERCEPTORS'),
     },
+    demo: getOsEnvOptional('APP_DEMO'),
```
VivekFitkariwala, line 42: Add toBool here itself

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
ppratikcr7: okay

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
ppratikcr7: okay

backend/packages/Upgrade/src/env.ts
ppratikcr7: okay

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
ppratikcr7: Inside we are using await, and it didn't gave me any error on this. Still for safer side we can put await. I will check and update accordingly

VivekFitkariwala: It won't give error but your http request will send response before the DB cleanup. I think you should not await then as awaiting might lead to request timeout.
But just keep in mind that DB cleanup might not be complete even though you received successful http response.

ppratikcr7: Oh okay got it. Okay, so should I remove `await` from the latest commit?

VivekFitkariwala (approved):

Issue #200, opened 2021-11-22
Allow any email account domain name for UpGrade


zackcl:
For the demo app, we'll need to allow any email account domain name for UpGrade and not just playpowerlabs.com and carnegielearning.com so anyone with a Google account can log in to UpGrade.

PR #203, opened 2021-11-23
fix for issue #200: Any domain allowed to login to upgrade with creator access

ppratikcr7:
In this PR, we have set domainName to blank, so that any domain's email ID can login to upgrade.

Also, we have provided `Creator` access to the new user for the demo. We have to revert this migration after the demo, so that we give the default `Reader` access to the new users.

You have to set `DOMAIN_NAME =` as blank in the` .env` file as mentioned in the `.env.example` file when you want Upgrade to not restrict to any specific domain to access the application. Else, if you want Upgrade to be accessible to any particular domain only, then please set `DOMAIN_NAME = domainName` in the` .env` and `environment.ts` files. For example: `DOMAIN_NAME = playpowerlabs.com` and `domainName: 'playpowerlabs.com'`

### DIFF
--- a/backend/packages/Upgrade/.env.example
+++ b/backend/packages/Upgrade/.env.example
@@ -60,7 +60,7 @@ MONITOR_PASSWORD=1234
 # Google Client Id
 #
 GOOGLE_CLIENT_ID=google_project_id
-DOMAIN_NAME = playpowerlabs.com
+DOMAIN_NAME =
 AUTH_CHECK=false
 SCHEDULER_STEP_FUNCTION=arn:aws:states:us-east-1:656500730782:stateMachine:ExperimentSchedular-development
 AWS_REGION=us-east-1
--- a/backend/packages/Upgrade/.env.test
+++ b/backend/packages/Upgrade/.env.test
@@ -60,7 +60,7 @@ MONITOR_PASSWORD=1234
 # Google Client Id
 #
 GOOGLE_CLIENT_ID=google_project_id
-DOMAIN_NAME = playpowerlabs.com
+DOMAIN_NAME =
 AUTH_CHECK=false
 SCHEDULER_STEP_FUNCTION=arn:aws:states:us-east-1:656500730782:stateMachine:ExperimentSchedular-development
 AWS_REGION=us-east-1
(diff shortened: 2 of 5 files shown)


backend/packages/Upgrade/.env.example
```diff
@@ -60,7 +60,7 @@ MONITOR_PASSWORD=1234
 # Google Client Id
 #
 GOOGLE_CLIENT_ID=google_project_id
-DOMAIN_NAME = playpowerlabs.com
+DOMAIN_NAME =
```
VivekFitkariwala, line 63: You should give example in comment for different variations. Updating it actually also removes the variation when specific domain is required.

backend/packages/Upgrade/.env.test
```diff
@@ -60,7 +60,7 @@ MONITOR_PASSWORD=1234
 # Google Client Id
 #
 GOOGLE_CLIENT_ID=google_project_id
-DOMAIN_NAME = playpowerlabs.com
+DOMAIN_NAME =
```
VivekFitkariwala, line 63: Same as above

frontend/projects/upgrade/src/environments/environment.ts
```diff
@@ -14,7 +14,7 @@ export const environment = {
   i18nPrefix: '',
   appVersion: require('../../../../package.json').version,
   gapiClientId: '135765367152-pq4jhd3gra10jda9l6bpnmu9gqt48tup.apps.googleusercontent.com',
-  domainName: 'playpowerlabs.com',
+  domainName: '',
```
VivekFitkariwala, line 17: Same as above

VivekFitkariwala (approved):

backend/packages/Upgrade/.env.example
ppratikcr7: Okay. I will do that.

backend/packages/Upgrade/.env.test
ppratikcr7: Okay.

frontend/projects/upgrade/src/environments/environment.ts
ppratikcr7: done

PR #204, opened 2021-11-28
Fix/issue 201: export and import experiment will now support queries and adding seeding of metrics

ppratikcr7:
This PR will export any metric queries if added on the experiment upon clicking the `Export Design` button.
User can then import the same experiment design to create an experiment with metric queries as well.

This PR also provides provision to seed initial list of metrics to be populated to the database. For this you need to add the metrics in the `.env` file as shown in the `.env.example` file. 
For example, I have added one single metric and one grouped metrics:
`METRICS = [{"metric": "totalTimeSeconds","datatype": "continuous"}, { "groupClass": "masteryWorkspace", "allowedKeys": [ "calculating_area_various_figures", "Compare_functions_diff_reps_quadratic" ], "attributes": [{ "metric": "timeSeconds", "datatype": "continuous"}]}]`


backend/packages/Upgrade/src/api/repositories/QueryRepository.ts
```diff
@@ -36,6 +36,26 @@ export class QueryRepository extends Repository<Query> {
     return result.raw[0];
   }
 
+  public async insertQueries(queryDoc: Array<Partial<Query>>, entityManager: EntityManager): Promise<Query> {
```
VivekFitkariwala, line 39: Why haven't you used `upsertQuery`?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -521,7 +526,11 @@ export class ExperimentService {
           (queries &&
             queries.length > 0 &&
             queries.map((query: any) => {
-              promiseArray.push(this.metricRepository.findOne(query.metric.key));
+              if (query.metric == null) {
+                promiseArray.push(this.metricRepository.findOne(query.query.metric.key));
```
VivekFitkariwala, line 530: This looks incorrect. What is this for?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -798,15 +807,43 @@ export class ExperimentService {
...
+            promiseArray.push(this.metricRepository.findOne(query.metric));
+            // tslint:disable-next-line:no-shadowed-variable
+            const { createdAt, updatedAt, versionNumber, metric, ...rest } = query;
+            rest.experiment = experimentDoc;
+            rest.id = rest.id || uuid();
+            return rest;
+          })) ||
+        [];
+      if (promiseArray.length) {
+        const metricsDocs = await Promise.all([...promiseArray]);
```
VivekFitkariwala, line 826: `[...promiseArray]` why you creating new array by spreading existing array?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -798,15 +807,43 @@ export class ExperimentService {
...
           return partition;
         });
-
+      
+      // creating queries docs
+      const promiseArray = [];
+      let queryDocsToSave =
+        (queries &&
+          queries.length > 0 &&
+          queries.map((query: Query) => {
```
VivekFitkariwala, line 816: Use `forEach` if you don't need the return value

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -798,15 +807,43 @@ export class ExperimentService {
...
+            const { createdAt, updatedAt, versionNumber, metric, ...rest } = query;
+            rest.experiment = experimentDoc;
+            rest.id = rest.id || uuid();
+            return rest;
+          })) ||
+        [];
+      if (promiseArray.length) {
+        const metricsDocs = await Promise.all([...promiseArray]);
+        queryDocsToSave = queryDocsToSave.map((queryDoc, index) => {
+          queryDoc.query.metric = metricsDocs[index];
```
VivekFitkariwala, line 828: You should take care of scenario where metric doc doesn't exist

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -798,15 +807,43 @@ export class ExperimentService {
...
       // saving conditions and saving partitions
       let conditionDocs: ExperimentCondition[];
       let partitionDocs: ExperimentPartition[];
+      let queryDocs: Query[];
       try {
         [conditionDocs, partitionDocs] = await Promise.all([
           this.experimentConditionRepository.insertConditions(conditionDocsToSave, transactionalEntityManager),
           this.experimentPartitionRepository.insertPartitions(partitionDocsToSave, transactionalEntityManager),
         ]);
+        if (queryDocsToSave) {
```
VivekFitkariwala, line 842: Add this in existing `Promise.all`
`queryDocsToSave.length > 0 ? Promise.all([this.queryRepository.insertQueries(queryDocsToSave, transactionalEntityManager)]) : Promise.resolve([]);`

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -798,15 +807,43 @@ export class ExperimentService {
...
+          queryDocs = await Promise.all([this.queryRepository.insertQueries(queryDocsToSave, transactionalEntityManager)]);
```
VivekFitkariwala, line 843: Why have you used `Promise.all` if in the array only one element is present?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -819,7 +856,11 @@ export class ExperimentService {
         const { experimentId, ...restDoc } = partitionDoc as any;
         return restDoc;
       });
-      return { ...experimentDoc, conditions: conditionDocToReturn as any, partitions: partitionDocToReturn as any };
+      const queryDocToReturn = 
+        !!queryDocs && queryDocs.map((queryDoc) => {
+        return queryDoc;
+      });
+      return { ...experimentDoc, conditions: conditionDocToReturn as any, partitions: partitionDocToReturn as any, queries: (queryDocToReturn[0] as any) || []};
```
VivekFitkariwala, line 863: Why the `[0]`?

backend/packages/Upgrade/src/env.ts
```diff
@@ -89,6 +89,7 @@ export const env = {
   initialization: {
     contextMetadata: JSON.parse(getOsEnv('CONTEXT_METADATA')),
     adminUsers: parseAdminUsers(getOsEnv('ADMIN_USERS')),
+    metrics: JSON.parse(getOsEnv('METRICS')),
```
VivekFitkariwala, line 92: `getOsEnv` means that it should not be null. This will throw error if in environment variable is not defined. And also what if the format is incorrect `JSON.parse` will throw error. Please take care of those scenarios

ppratikcr7: okay.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: On `queryDocs.map()` it was returning array of array [[{},{}]]. So, had to give the first indexed Doc.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -128,6 +128,12 @@ export class ExperimentService {
     };
   }
 
+  public getMetricQueries(): object {
```
VivekFitkariwala, line 131: Where are you using this? And as this only return value store in `env` you should directly use that.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Oh yes, nice catch. Will update this.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: okay

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: ok

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: In the `queries`, we get already created query and newly added query to be updated in this `updateExperimentInDB` function. The already created query has different doc format then the newly added one's as shown below:

`[
  {
    createdAt: '2021-11-28T09:54:41.740Z',
    updatedAt: '2021-11-28T09:54:41.740Z',
    versionNumber: 1,
    id: 'fedaff16-9654-408b-9ccc-bd55a92ed58d',
    name: 'total time',
    query: { metric: [Object], operationType: 'sum' },
    repeatedMeasure: 'MOST_RECENT',
    metricKey: null,
    experimentId: 'be3ae74f-370a-4015-93f3-7761d16f8b17'
  },
  {
    createdAt: '2021-11-28T09:54:41.740Z',
    updatedAt: '2021-11-28T09:54:41.740Z',
    versionNumber: 1,
    id: '9ba5b350-d257-4655-91f1-7ecaa360c6ab',
    name: 'time',
    query: { metric: [Object], operationType: 'avg' },
    repeatedMeasure: 'MEAN',
    metricKey: null,
    experimentId: 'be3ae74f-370a-4015-93f3-7761d16f8b17'
  },
  {
    name: 'avg time',
    query: { operationType: 'avg' },
    metric: { key: 'totalTimeSeconds' },
    repeatedMeasure: 'MOST_RECENT'
  }
]`

But I have changed this now as per what you want.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Screenshot 2021-11-28 at 4 21 38 PM
 I am keeping it out separate.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Done. Removed it. Not required.

backend/packages/Upgrade/src/env.ts
ppratikcr7: done

backend/packages/Upgrade/src/api/repositories/QueryRepository.ts
ppratikcr7: As in upsert query it will check for conflict each time in db through all queries and its not required for inserting fresh queries as it wont be present in db. I think why to add more computation on db. Even for conditions and partitions we have separate queries for upsert and insert.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: We are taking it in new metricDocs array to use ahead. You have also used the same in the same `ExperimentService.ts` file in `updateExperimentInDB()` function as shown below:

Screenshot 2021-11-29 at 12 30 37 AM

Let me know if there is something wrong, I would change at both places.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -542,7 +533,7 @@ export class ExperimentService {
         if (promiseArray.length) {
           const metricsDocs = await Promise.all([...promiseArray]);
           queriesDocToSave = queriesDocToSave.map((queryDoc, index) => {
-            queryDoc.metric = metricsDocs[index];
+            metricsDocs ? queryDoc.metric = metricsDocs[index]: queryDoc.metric = null;
```
VivekFitkariwala, line 536: You should not save the query with no valid metric.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
VivekFitkariwala: Yup that is incorrect correct at both the places

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -825,24 +816,24 @@ export class ExperimentService {
       if (promiseArray.length) {
         const metricsDocs = await Promise.all([...promiseArray]);
         queryDocsToSave = queryDocsToSave.map((queryDoc, index) => {
-          queryDoc.query.metric = metricsDocs[index];
+          metricsDocs ? queryDoc.metric = metricsDocs[index]: queryDoc.metric = null;
```
VivekFitkariwala, line 819: Change accordingly

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -825,24 +816,24 @@ export class ExperimentService {
...
-      let queryDocs: Query[];
+      let queryDocs: any;
       try {
         [conditionDocs, partitionDocs] = await Promise.all([
           this.experimentConditionRepository.insertConditions(conditionDocsToSave, transactionalEntityManager),
           this.experimentPartitionRepository.insertPartitions(partitionDocsToSave, transactionalEntityManager),
         ]);
-        if (queryDocsToSave) {
-          queryDocs = await Promise.all([this.queryRepository.insertQueries(queryDocsToSave, transactionalEntityManager)]);
+        if (queryDocsToSave.length > 0) {
```
VivekFitkariwala, line 833: You should add this code in the above await statement with Promise.all else they will run sequentially

backend/packages/Upgrade/src/init/seed/initMetrics.ts
```diff
@@ -1,12 +1,19 @@
...
 export function InitMetrics(): Promise<any> {
   const metricService: MetricService = Container.get(MetricService);
   // Init default metrics in system
   if (env.initialization.metrics) {
-  return metricService.saveAllMetrics(env.initialization.metrics);
+    try {
+      return metricService.saveAllMetrics(JSON.parse(env.initialization.metrics));
+    } catch (error) { 
+      error = new Error('Error while initializing metrics');
+      (error as any).type = SERVER_ERROR.QUERY_FAILED;
```
VivekFitkariwala, line 14: This is not a query failed error.
It's not necessary to assign type to all the errors. The error with type is shown on the UI. So remove the type if we don't have any appropriate error code.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: okay

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: okay. I tried that as you mentioned earlier, but I was not able to put condition on this queryDoc.

backend/packages/Upgrade/src/init/seed/initMetrics.ts
ppratikcr7: Okay. This can be due to JSON parsing or QUERY_FAILED ON DB. We dont have type for JSON parsing error, so have used QUERY_FAILED. Ok, I will remove the type.

PR #216, opened 2021-12-18
Fix/issue 164 PR for all other controllers then ExperimentClientController

ppratikcr7:
PR-2: This includes logger format change for all controllers except ExperimentClientController which is covered in PR-1.

This will replace all the old logger instances throughout the codebase.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/AnalyticsService.ts
+++ b/backend/packages/Upgrade/src/api/services/AnalyticsService.ts
@@ -24,12 +24,12 @@ import { MonitoredExperimentPointLogRepository } from '../repositories/MonitorEx
 import { In } from 'typeorm';
 import fs from 'fs';
 import { SERVER_ERROR } from 'upgrade_types';
-import { Logger, LoggerInterface } from '../../decorators/Logger';
 import { env } from '../../env';
 import { METRICS_JOIN_TEXT } from './MetricService';
 import { ErrorService } from './ErrorService';
 import { ExperimentAuditLogRepository } from '../repositories/ExperimentAuditLogRepository';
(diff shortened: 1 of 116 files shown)


backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -81,12 +80,12 @@ export class AnalyticsService {
       groupExclusion,
     ] = promiseArray[1];
 
-    this.log.info('individualEnrollmentByCondition', individualEnrollmentByCondition);
-    this.log.info('individualEnrollmentConditionAndPartition', individualEnrollmentConditionAndPartition);
-    this.log.info('groupEnrollmentByCondition', groupEnrollmentByCondition);
-    this.log.info('groupEnrollmentConditionAndPartition', groupEnrollmentConditionAndPartition);
-    this.log.info('individualExclusion', individualExclusion);
-    this.log.info('groupExclusion', groupExclusion);
+    logger.info({ message : `individualEnrollmentByCondition ${individualEnrollmentByCondition}` });
```
VivekFitkariwala, line 83: This will get logged as [Object Object]. Instead log this as another property of the object `data`
```
{
     message: "individualEnrollmentByCondition",
     data: individualEnrollmentByCondition
}
```

VivekFitkariwala (approved):
Because we are logging object you are free to create other properties within a object don't try to add all the details in the message itself

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
ppratikcr7: okay

VivekFitkariwala (approved):

PR #217, opened 2021-12-28
Fix/issue 195: RDS read replica DB for export data functionality

ppratikcr7:
In this PR, I have made changes in terraform scripts to allow creation of multiple read replicas of RDS DB of current environment DB. I have also made changes to use separate db connection during `Export Data`. 

Once terraform would create the read replica DB's and deploy the backend in elastic beanstalk with the new environment variable named `RDS_HOST_REPLICAS`, the Upgrade will use a multiple connections to db instances and trigger export data related queries on first read replica DB which in turn would not create any load or blockage to our master DB. Any extra read replicas would be used as a stand by read replica db for the main connection.


backend/packages/Upgrade/src/loaders/typeormLoader.ts
```diff
@@ -6,14 +6,55 @@ import { SERVER_ERROR } from 'upgrade_types';
 
 export const typeormLoader: MicroframeworkLoader = async (settings: MicroframeworkSettings | undefined) => {
   const loadedConnectionOptions = await getConnectionOptions();
```
VivekFitkariwala, line 8: Create common objects here.

backend/packages/Upgrade/src/loaders/typeormLoader.ts
```diff
@@ -6,14 +6,55 @@ import { SERVER_ERROR } from 'upgrade_types';
...
+      slaves: [],
+    },
+    synchronize: env.db.synchronize,
+    logging: env.db.logging,
+    maxQueryExecutionTime: env.db.maxQueryExecutionTime,
+    entities: env.app.dirs.entities,
+    migrations: env.app.dirs.migrations,
+  });
+  const slaveConnectionOptions: ConnectionOptions = Object.assign(loadedSlaveConnectionOptions, {
+    name: 'export',
```
VivekFitkariwala, line 35: Can you create an enum for the connection name because having hardcoded string will create trouble in the future

VivekFitkariwala (changes requested):
I have added some comments can you fix those.
Thanks

ppratikcr7:
> I have added some comments can you fix those.
> Thanks

Sure

backend/packages/Upgrade/src/loaders/typeormLoader.ts
ppratikcr7: ok

backend/terraform/environments/staging/tfvars.sample
```diff
@@ -13,6 +13,8 @@ identifier              = "dev-postgres"
 instance_class          = "db.t2.small"
 storage_type            = "gp2"
 multi_az                = "false"
+replica_count           = 2
```
VivekFitkariwala, line 16: One of these two seems redundant. Can we remove count and use array length to generate count

ppratikcr7: done

backend/packages/Upgrade/src/loaders/typeormLoader.ts
ppratikcr7: done

VivekFitkariwala (approved):

VivekFitkariwala (approved):

PR #218, opened 2022-01-02
Fix/issue 143

VivekFitkariwala:
The PR contains fixes for
- Taking types directly from the types folder instead of having a dependency in package.json
- Updating boilerplate of frontend, backend and clientLib to remove upgrade_types as a dependency.
- Updating github actions accordingly


types/package.json
```diff
@@ -2,11 +2,8 @@
   "name": "upgrade_types",
   "version": "1.0.5",
   "description": "",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
-  "files": [
-    "dist/*"
-  ],
+  "main": "src/index.ts",
+  "types": "src/index.ts",
```
ppratikcr7, line 6: Just to confirm, types would also be using the same `index.ts` file and not the `index.d.ts` as earlier?

ppratikcr7 (approved):
Just check for merge conflicts, rest looks fine.

types/package.json
VivekFitkariwala: Yup now removing the build process from the types and consuming typescript files directly by other projects.

ppratikcr7: okay got it.

PR #229, opened 2022-01-20
locust tests follow user flow

jreddig:


backend/locust/cleanup_experiments.py
```diff
@@ -0,0 +1,15 @@
+from  upgrade_mathia_data import modules, workspaces
+import deleteExperiment
+
+protocol = "http"
+host = "localhost:3030"
+
+option = int(input("Enter the number of  experiments to delete: "))
```
msandbothe, line 7: This file should be improved. Dangerous to use even in QA. Perhaps an interactive prompt to create/delete experiments

msandbothe: Splitting it out from main script was appropriate. Create a separate task to improve create/delete process.

backend/locust/load_test_upgrade_mathia.py
```diff
@@ -0,0 +1,356 @@
...
+            retClassData.append(schools[schoolId]["classes"][classId])
+
+    return retClassData
+
+# Main Locust API calls for enrolling students in an experiment:
+class UpgradeUserTask(SequentialTaskSet):
+
+    # each User represents one Student
+    def on_start(self):
+        self.student = initStudent()
```
msandbothe, line 143: No changes here except removing the lock, correct?

backend/locust/readme.md
```diff
@@ -11,16 +11,53 @@ Install locust:
...
+
+## changing from individual level to group level experiment
+
+On line 20 in load_test_upgrade_mathia.py, change False to True
+
+    groupExp = False
+
+False represents an individual level experiment and is the default value.
+
+True represents a group level experiment.
```
msandbothe, line 63: This could be rolled into a createExperiment process (separate task). We could provide some example experiment jsons that the script can choose from.

msandbothe (commented):
Create issue for improving the creation/deletion process. Other changes look good.

jreddig:
Main changes:
 - user driven instead of api driven, each user represents a student
 - test sequence models student flow
 - experiments created by script marked as "LocustLoadTest"
 - group vs individual experiment set on creation interactively
 - deleteExperiment only grabs marked experiments
 - deletion of experiments done separately from tests
 - user certifies choice to delete experiment
 - Added comments to explain test/method functionality
 - More explanation in ReadMe

msandbothe (approved):

PR #241, opened 2022-02-02
Filtering users with null ID

RidhamShah:
Solved the issue by filtering out all the users with Null ID and null Users.


ppratikcr7 (approved):

VivekFitkariwala (changes requested):
Instead of filtering the users, log error in this scenario.

YashilDepani:
Not to filter the users but instead, throw an informative error.

PR #266, opened 2022-02-16
Run UpGrade Developer Environment with Docker

kawcl:
Any and all feedback is welcome.
Feel free to add anyone as a reviewer who you think would be interested.

This is kind of a proof of concept.  
Started doing this when trying to reproduce the QA environment in order to look into a bug.
This will run UpGrade with a frontend, backend, and postgres container linked through a docker network.

I created 2 new dockerfiles in both the frontend and backend directories, both named `dev.Dockerfile`
Both dockerfiles are used in conjunction with docker-compose, where I use bindmounts to link the host machine and the containers running the respective applications.  This bindmount allows for hot reloading when doing edits to the code on your machine.

I also included a Makefile with what I thought were desirable docker and compose commands.
It includes commands to run the network of containers, build the container images from the`dev.Dockerfile`, display the container log history and follow them live, enter a bash terminal inside the frontend & backend containers, take down and clean up the docker environment, as well as backup and restore the postgres db

The backend also uses a new `.env` file passed to it in the compose file.

This would also reduce initial setup of the application for local development down to just a few steps.
```
1. Install Docker
2. `make setup-local` --- this runs a script to do all npm installs required on the application
3.  `make` --- first go at this will take a little while because it'll need to download the base images, build our images, start the containers, and then let the apps startup
```


backend/dev.Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:12.7.0
```
ppratikcr7, line 1: @kawcl Can you try once with nodejs 16 image using `FROM node:16.3.0-alpine`. I have checked our Upgrade for nodejs 16 locally and I works fine, just we need to upgrade the "pg" package in backend/packages/Upgrade/package.json from "^7.12.1" to "^8.3.0". Can you do these two changes for backend and the `FROM node:16.3.0-alpine` change for frontend and test if the local development works well?

kawcl: I can make these changes and try getting things to run.  I'll be using a non-alpine node image just because of how I set up docker locally for development. I don't want to use the minimized OS.

mfugate1: I would strongly recommend using the same OS in your dev environment and in deployments. There are so many times where I run into trouble because something in alpine works very differently than ubuntu and I waste a lot of time trying to track down the problem.

Is there something you need for development that alpine doesn't have?

kawcl: no not necessarily. it's just easier to not work with alpine since I'm not as familiar with it but I can try to get it running with alpine, it hopefully shouldn't be too much work to make the switch.

kawcl: update to 12.7.0-alpine was more flawless than I expected. I'll try the update with 16.3.0-alpine

jerith666:
Hope I'm not overstepping by pushing bfc9b8f.  :)

Running `make setup-local` fails for me with:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: ab_testing_backend@1.0.86
npm ERR! Found: class-transformer@0.3.1
npm ERR! node_modules/class-transformer
npm ERR!   class-transformer@"^0.3.1" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer class-transformer@"^0.2.3" from routing-controllers@0.8.1
npm ERR! node_modules/routing-controllers
npm ERR!   routing-controllers@"^0.8.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! See /home/mmchenry/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/mmchenry/.npm/_logs/2022-02-17T22_25_06_986Z-debug.log
```

I'm at npm version 8.1.2 if that matters.

kawcl:
No problem! This makes sense to have it nix compatible. I'm all for collaborative updates.

I've tried to update the node version in the docker images to 16.3.0 and started running into this same error you are getting with `make setup-local`.  I've been trying to figure out what's going on with it.  Using `npm install --legacy-peer-deps` does seem to solve it for me but I'm not sure if it's the right solution.  It's something I'm looking into.

jerith666:
I don't fully grok `--legacy-peer-deps`, but my shallow understanding of it makes me wary that it's the right solution.

kawcl:
Added a commit to update to the `node:12.7.0-alpine` image.
I tried migrating to a node version of `16.3.0` but ran into numerous npm dependency issues that I couldn't resolve.
The migration task will probably need someone more familiar with the project and npm in general.

@jerith666 The only way I've been able to get passed the dependency tree issue is to be on a node version of `12.7.0` and a npm version of `6.13.4`.  I have been managing the different node & npm versions with `nvm` : https://github.com/nvm-sh/nvm

danoswaltCL:
Since this PR is just going to handle local docker setup as-is and not solve _everything_ , I will add an issue specifically for the `api/query/analyze` issue and then once this merged, I will put in a PR to update docker images to one that will pull node 12.22+, which will solve it that.

A major node version jump to 14 or 16 seems like it will need a little more attention before we leap.

jerith666:
If the script needs some specific version of node/npm to run, I submit that the script should ensure it obtains that version itself.  See 1b398cb1 for how to do that with Nix.  (Not bold enough to push _that_ directly to your branch. :) )

docker-setup.sh
```diff
@@ -0,0 +1,58 @@
...
+    npm install
+
+    cp -R ../types packages/Upgrade
+    cd packages/Upgrade
+    rm -rf package-lock.json node_modules
+    npm install
+
+    cd ../../../frontend
+    rm -rf package-lock.json node_modules
+    npm install
```
jerith666, line 25: All these `npm install`s mean that after I run `make setup-local`, `git status` is not clean.  IMO that's not a great developer experience -- we should run `npm ci` in this script instead.

Choosing to consume new minor/patch/etc. versions of dependencies should be an explicit thing that a developer plans/chooses to do, not something that can happen by accident during local environment setup.

kawcl: ok I can update the script and look into how to obtain the right version with nix

jerith666:
`make setup-local` succeeded for me on NixOS!  :)

However:

```
$ make
---- Launching Environment ----
docker-compose up -d
Creating network "upgrade_localdev" with driver "bridge"
Creating volume "upgrade_upgradedbdata" with default driver
Building frontend
Sending build context to Docker daemon  1.017GB
Step 1/8 : FROM node:12.22.10-alpine
12.22.10-alpine: Pulling from library/node
59bf1c3509f3: Pull complete 
8769eb813ad5: Pull complete 
7025e9ac362e: Pull complete 
1efe07d207fa: Pull complete 
Digest: sha256:dfa564312367b1a8fca8db7ae4bae102b28e68b39ebcb7b17022c938f105846b
Status: Downloaded newer image for node:12.22.10-alpine
 ---> 1b156b4c3ee8
Step 2/8 : WORKDIR /usr/src/app
 ---> Running in 7aeebc5e4718
Removing intermediate container 7aeebc5e4718
 ---> 856dc7e5a1be
Step 3/8 : COPY ./../types ./types
COPY failed: forbidden path outside the build context: ../types ()
ERROR: Service 'frontend' failed to build : Build failed
make: *** [Makefile:18: default] Error 1
```

kawcl:
Hmm, I'll try to clean up my docker environment and start from scratch to try and reproduce this error.
I'm thinking I might need Linux to test it out.

kawcl:
So I'm wasn't able to reproduce this issue and this confuses me..
```
COPY failed: forbidden path outside the build context: ../types ()
ERROR: Service 'frontend' failed to build : Build failed
```
because the build context is specified in the `docker-compose.yml` file for both the the frontend & backend images separately. If you look at both of them the build context is `.` or the directory of the compose file, which is the root of the project. I did that specifically because of this `types` folder being where it's at.  I'm not sure if this is an issue differences between how docker is setup on Linux vs Mac.

danoswaltCL (approved):
If this is mostly there pending some tweaks that can be safely addressed later, I vote to merge this into dev branch as soon as we can so we can have it our current branches and steamline testing. @jerith666 are the issues you're having blocking us from being able to merge this as-is currently?

jerith666:
I'm not doing day-to-day development on UpGrade, so in that sense, no, it shouldn't prevent a merge if it's working fine for others.

But, we should open another issue to figure it out later.

FWIW, I have:

```
$ docker --version
Docker version 20.10.12, build v20.10.12
```

(maybe this is another place where Nix could help by making sure we're all using the same versions of the same tools?)

kawcl:
> I'm not doing day-to-day development on UpGrade, so in that sense, no, it shouldn't prevent a merge if it's working fine for others.
> 
> But, we should open another issue to figure it out later.
> 
> FWIW, I have:
> 
> ```
> $ docker --version
> Docker version 20.10.12, build v20.10.12
> ```
> 
> (maybe this is another place where Nix could help by making sure we're all using the same versions of the same tools?)

This is the docker version that I'm running as well but being on Mac I need to use Docker Desktop so I'm not sure if that has an affect. I did notice yesterday looking around the internet about nix that it might be possible to have the makefile run with nix.  

In the meantime, I'm fine with getting this in and then I can focus on resolving the issue Matt is having through another issue.

danoswaltCL (approved):

Issue #267, opened 2022-02-18
`api/query/analyze` fails in QA


danoswaltCL:
`Promise.allSettled` does not work in node 12.7, fix is to upgrade docker image to 12.22.

danoswaltCL:
will create PR once dev updated with docker-local https://github.com/CarnegieLearningWeb/UpGrade/pull/266 so I can accurately test the update in local

Issue #260, opened 2022-02-10
Upgrade Docker base image to latest 12 version


danoswaltCL:
Current image in Dockerfile (`node:12.7.0-alpine`) is outdated https://hub.docker.com/_/node, and is likely cause of bugs seen in `api/query/analyse`, which fail because of outdated version of node on server (`Promise.allSettled` not available on 12.7, needs 12.9 and up).

Can we create an issue to update the docker image to `12-alpine3.14` or `12-alpine` in dev and get it into QA asap if all looks good there? (these will pull node 12.22^).

In near future upgrading to 14+ will be necessary also, as 12 will be no longer officially in next couple months. https://nodejs.org/en/about/releases/

nirmalpatel:
Should we just try upgrading to 14+ now instead of later? @ppratikcr7 not sure how much work will be required to move to v14 or v16.

ppratikcr7:
@nirmalpatel Not sure, will have to look into it. Let's discuss about this in our internal scrum meet.

danoswaltCL:
closing this as 12.22.10 is being used now, separate issue should be created to move dev branch to node 16 image

PR #269, opened 2022-02-22
update docker image to 12.22.1

danoswaltCL:
#260 #267

Updating node image 12.7 -> 12.22 in docker. This will ensure code will work for `api/query/analyze` endpoint.

Pushing this now because it is blocking ability to push QA code to production. Linking issue #267 for upgrading the docker image @ppratikcr7, node version 16 was not working smoothly for us in local yet (npm install / dependency version issues raise questions), still needs to be looked into further.

Tested via @kawcl's local-docker branch off of latest dev with recent replica of QA db.

@amurphy-cl

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.7.0-alpine
+FROM node:12.22.1-alpine
 
 ENV NEW_RELIC_NO_CONFIG_FILE=true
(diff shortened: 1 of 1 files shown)


jreddig (approved):

VivekFitkariwala (approved):

danoswaltCL:
updated for correct patch version (12.22.10 instead of 12.22.1), please reapprove!

kawcl (approved):

PR #271, opened 2022-02-24
use logic for single replica for now

danoswaltCL:
trying to revert code to state that seems to work before multiple replicas logic was added to unblock dev environment: https://github.com/CarnegieLearningWeb/UpGrade/commit/726dcb8ccdb34149ee8f7907f25a105993b8f8e0

@mfugate1 will need either TYPEORM_HOST_SLAVE or RDS_HOSTNAME_SLAVE to be set to whatever they need to be set to, @ppratikcr7 and @VivekFitkariwala please let me know if this looks like it will work in order to get dev environment working.

Will this spin up a replica (so we can test export or whatever else is tied to that?), or will that only happen in higher envs?


mfugate1:
There are currently no replicas in any of our environments, and I'd prefer to get dev working without any replicas before adding one.

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/blob/1.0.87/backend/packages/Upgrade/src/loaders/typeormLoader.ts is what is in QA/Prod currently, AFAIK. I was assuming that a single read replica was already a working thing, have multiple connections like this been tested?

@ppratikcr7 @VivekFitkariwala @nirmalpatel

In dev branch, we have code that looks for two separate connections by name, 'export' and 'default', else I would revert to the above single connection, but that would be further code surgery. Perhaps multiple connection names can all point to the same db host for now, which basically how I get it to boot up successfully in local (TYPEORM_HOST_SLAVE = postgres just like the default), but will that actually work?

danoswaltCL:
@ppratikcr7 will be opening different PR

PR #276, opened 2022-02-25
temp force export to default host

danoswaltCL:


danoswaltCL:
Temporary fix to unblock the dev env. This appears to need the db details to be provided in master.host, and when it sees an empty string `''` it throws the `127.0.0.1:5432 connection refused error`. I suspect if we force export function connection to be the same as the default for now, we won't get this issue.

Dev env still doesn't connect, and when I pull the code down it doesn't connect for me in local docker setup. Something doesn't seem right with the replica connection stuff yet, not sure how it works on your env @ppratikcr7.

Issue #301, opened 2022-03-07
Backend using Alias IDs instead of Root User IDs


ppratikcr7:
During regression testing we found a bug in the user doc in which instead of pointing the root user id, we are taking the alias user id. This change was introduced in logger format [commit](https://github.com/CarnegieLearningWeb/UpGrade/commit/12b23ded25d9ac8166765e0d40bd9290eec866f9) where we tried to have a single level fetching of userDoc and pass it though the logger instance.

This was not caught in test cases as its a controller level change and we currently don't have controller level integration test cases. We are planning to soon have test cases for controllers as well.

Note: We will have to clear the db.

danoswaltCL:
Can we add a catch block to this function if it will still be used? There's a question of lambda calls timing out when `api/log` is being called and no logging to show any errors from Upgrade. I am wondering if as a symptom of the id issue is that this call is happening but the Promise either never resolves and breaks or the rejection is swallowed without any logging.

@jreddig @msandbothe do you think this could be a place where a failure would result in the behavior noticed in the lambda failures to `api/log`?

```
public async getUserDoc(experimentUserId, logger) {
    const experimentUserDoc = await this.experimentUserService.getOriginalUserDoc(experimentUserId, logger);
    if ( experimentUserDoc ) {
      var aliasUserId = experimentUserId;
        if (!experimentUserDoc.aliases) {
          aliasUserId = experimentUserId;
        }
        const userDoc = {
            id: aliasUserId,
            originalUserId: experimentUserDoc.id,
            group: experimentUserDoc.group,
            workingGroup: experimentUserDoc.workingGroup
        };
        return userDoc;
    } else {
      return null; 
    }
  }
```

ppratikcr7:
@danoswaltCL I have added error handling and more error logs and also logs inside api/log call to give you more idea on this issue.

PR #302, opened 2022-03-07
Fix/issue 301

ppratikcr7:
Things covered in this PR are as below:

1. Fix for setting root userId instead of alias user id in userDoc
2. Adding error handling try catch blocks in getUserDoc()
3. Adding missing error logger instances to better debug errors in splunk
4. Added logs in `api/log` to trace performance issue.
5. Made change in test log format to pretty print JSON format logs instead of [ Object ]


backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -764,19 +764,15 @@ export class ExperimentClientController {
 
   public async getUserDoc(experimentUserId, logger) {
     const experimentUserDoc = await this.experimentUserService.getOriginalUserDoc(experimentUserId, logger);
```
danoswaltCL, line 766: had comment on the issue that I posted at same time you put up the PR:

https://github.com/CarnegieLearningWeb/UpGrade/issues/301

danoswaltCL: what happens if `this.experimentUserService.getOriginalUserDoc` errors? It looks like it needs a try catch block.

danoswaltCL (changes requested):
This may need error-handling / logging in scenarios where this internal network call errors, otherwise it will be hard to trace.

ppratikcr7:
@danoswaltCL Okay. Will do it.

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
ppratikcr7: Adding the try catch block.

danoswaltCL: `const experimentUserDoc = await this.getUserDoc(logData.userId, request.logger);`

I have a question on this line also, what is the performance implication of including this internal network call? It will be an extra network call for each endpoint in this controller?

danoswaltCL (dismissed):

danoswaltCL:
I'm okay with this so we can solve user id issues, but I'm wondering if there's a better way to handle `getUserDoc`. When a user logs in, this network call to get same user doc from db will get called for every endpoint (i guess potentially adding 8-9 calls for each student login). I don't know if that is what is causing the production servers to overheat, but it has to be a performance hit. Can we cache this via user session or something?

YashilDepani (approved):

nirmalpatel:
@danoswaltCL can we get this fix in without optimization first since it affects most of the enrollment counts in MATHia and then do optimizations in the next PR?

danoswaltCL (approved):

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
ppratikcr7: We will track this performance optimisation in another ticket.

VivekFitkariwala (approved):

Issue #310, opened 2022-03-17
Condition name cut off when setting preview user


amurphy-cl:
When setting a preview user, the condition names are cut off so that you can't tell which condition you're assigning

![image](https://user-images.githubusercontent.com/65311369/158884925-f8ab12ae-0393-4140-846a-cf8896f689eb.png)

danoswaltCL:
this seems like a quick thing i can look into, along with the other truncate issue

Issue #308, opened 2022-03-16
Tooltip under Conditions in dashboard truncates


amurphy-cl:
The tooltip under Conditions in the experiment dashboard/detail page truncates the name of the condition. It should show the full condition name.

![image](https://user-images.githubusercontent.com/65311369/158649717-6f7a0def-3792-437d-a6aa-106fdea07939.png)

danoswaltCL:
i can take a look at this

PR #313, opened 2022-03-18
adds global css to override tooltip/selection option truncation

danoswaltCL:
This takes care of #308 #310

For both the [MatTooltip](https://material.angular.io/components/tooltip/api) and [MatSelect Options](https://material.angular.io/components/select/overview) widgets, truncation for long content is automatic. No param is exposed that allows you to change this, so this is a basic css override to accommodate long text values, but note, this would be _**global**_ change for all tooltips and dropdowns.

Tooltips will just wrap, which look pretty natural, no issues.
Dropdowns will extend horizontally without wrapping, see notes below:

Screen Shot 2022-03-18 at 3 10 32 PM
nt.com/97542869/159071524-46aa15aa-b820-46f8-8fe6-dd7a77286c9e.png">
Screen Shot 2022-03-18 at 3 10 45 PM
Screen Shot 2022-03-18 at 3 09 59 PM
Screen Shot 2022-03-18 at 3 09 35 PM

A quick look around the UI seemed pretty good, tooltips shouldn't cause any issues with long texts. These dropdown select option widths could get pretty wide depending on the length of the text, but it looks like this will work well.

If long horizontal text turns out to cause some real estate issues, there are some other ways: you could have a dropdown of 'normal' width that allows overflow of text, which would create a horizontal scrollbar, which is kind of gross:

image

Or some word-wrap, which I only messed around with a little bit, it could improved but still, kinda gross, and could grow pretty long vertically:

image

### DIFF
--- a/frontend/projects/upgrade/src/styles.scss
+++ b/frontend/projects/upgrade/src/styles.scss
@@ -196,3 +196,17 @@
     opacity: 0 !important;
   }
 }
+
+// Material library doesn't expose any params to prevent content truncation for these widgets.
+// Therefore we override the css. If option appears in future, these can go away.
+
+// Added to globally prevent truncation of content dropdowns
+.mat-select-panel-wrap .mat-select-panel {
+  max-width: fit-content;
+}
+
+// Added to globally prevent truncation in tooltips
+.mat-tooltip {
+  word-break: break-all;
+  white-space: normal;
+}


danoswaltCL:
@amurphy-cl let me know if this looks good to you

ppratikcr7 (approved):

PR #321, opened 2022-03-25
fix for off metric data due to excluded users data pulled in

ppratikcr7:
This PR contains fix for metric data pulled for excluded users as query didn't had join with monitored experiment point table to check metric data only for enrolled users. I have verified this on the prod_clone db shared by @jreddig and now we see metric data for only enrolled users using this fix. Test cases where also updated to mark the experiments so as to enroll users in the experiment and we get only enrolled user metric data.


jreddig:
This looks great, Pratik! Thanks so much for working on this! @danoswaltCL are we still on the code freeze or are we good to merge?

danoswaltCL:
@jreddig we should be fine to merge to dev, my Q is if we want this pulled in as a hotfix to production (1.0.95)? @amurphy-cl?

danoswaltCL (approved):

PR #332, opened 2022-03-30
add some coverage badges

danoswaltCL:
This will create some code coverage badges that will show up in the frontend `README.md`.

image

It's not automatic in the commit hook when running tests, as husky seems to not like that very much.

But if either of these is run while testing, the badges will get automatically generated in `README.md`, which can then be committed along with your branch changes if you so please.

`npm run test:badges`
`npm run test:dev`


jreddig:
These are cute! I want to add coverage badges for the backend too

danoswaltCL:
yeah, should be exact same process for backend if we like this. potentially the integration tests can be included also. Want me to try that out? @jreddig

Issue #294, opened 2022-03-02
Update docker image to node 16


danoswaltCL:
Creating new issue to keep separate from a bugfix where this was discussed and solved by updating to 12.22.10 and @ppratikcr7 already did some work to analyze the impact. Mainly: `npm install` on our app with node 16 gives peer dependency errors / warnings.

ppratikcr7:
Yeah, I am working on this further.

PR #346, opened 2022-04-07
Fix/issue #294: Upgrading Node and npm versions

ppratikcr7:
This PR will upgrade the Node version from v12 to v16 and npm version from v6 to v8 for backend and frontend
Code and test cases are also updated according to same.
Angular version would be updated in separate PR


danoswaltCL (commented):
1. The main docker file still says 12.22.0 node image, should be updated to 16.3.0

2. Running in local docker setup I get js heapspace error in the frontend build.... running outside of docker all appears to work fine. Did you run into this @ppratikcr7 ?

danoswaltCL:
> 1. The main docker file still says 12.22.0 node image, should be updated to 16.3.0
> 2. Running in local docker setup I get js heapspace error in the frontend build.... running outside of docker all appears to work fine. Did you run into this @ppratikcr7 ?

I increased the memory settings in my local docker app (via the docker GUI > settings gear > resources > advanced) from 2GB to 4GB, and it started.

It could be very well that the app uses resources differently now with npm / angular versions updated; many suggestions online point to including a command in the `npm scripts` or in the Dockerfile to add something like `"export NODE_OPTIONS=--max_old_space_size=4096` or whatever is needed. If this is an issue that pops up for others or in the pipeline builds, this may be something that needs to be added.

ppratikcr7:
Thanks @danoswaltCL for pointing this out, actually I had mentioned in the meeting notes that I am yet testing this for docker setup. Earlier I did check on docker when I managed to get frontend peer dependencies to be installed using a `force` flag. But, I am yet to test this now when I have managed to get frontend packages installed without a `force` flag. Thanks for the finding, this will help me out in case I also get into this issue.

danoswaltCL (approved):

ppratikcr7:
@danoswaltCL I have added the env variable to set NODE_OPTIONS for heap memory to 4GB in the latest [commit](https://github.com/CarnegieLearningWeb/UpGrade/pull/346/commits/8b103785c3595173af020c270bbed62bf7156cc0). Still, we need to update the memory unto 4Gb in Docker GUI for it to work. We still need to check it once when this runs on dev branch deployment. The extra memory used probably is due to Node16 version and buildOptimizer value set to true during this PR change. Angular is still kept to v8 as moving to higher versions was causing a lot of issues, I will do this in a new PR later.

danoswaltCL:
I've pulled latest, app runs as expected (and in docker, looks good), but the jest tests won't run for me when I test a commit:

`TypeError: Jest: a transform must export a `process` function.`

The suggestion for this error is to make sure `jest` and `ts-jest` are matching and 27+.

When i tried this, it wants typescript 3.8 at least: `peer typescript@">=3.8 <5.0"`.

Are you able to run tests @ppratikcr7 ? Could just be an issue on my side if so.

ppratikcr7:
@danoswaltCL Actually I downgraded jest from v27 as they require angular versions above v9 and this was getting populated as npm warnings in node v12, but the same gets logged as npm error in node v16. I am able to run the jest tests, also they do run successfully on all commits in this PR. Also, I had to keep typescript version in frontend to v3.4 due to few other packages that doesn't support typescript above 3.4

Maybe someone else can also verify once and we may find out what's exactly going on,

danoswaltCL (approved):
ok, could be just my local setup that is not working with jest, shouldn't block this PR if the tests are running fine elsewhere

ppratikcr7:
@YashilDepani Can you try it out once?

VivekFitkariwala (approved):

danoswaltCL (dismissed):

amurphy-cl:
Are we waiting on the enrollment table to merge this?

ppratikcr7:
Actually we were waiting for the production to be up and work normal and later merge this in dev. Now, we can wait for enrollment table changes to be merged first, then will merge this one.

ppratikcr7:
@amurphy-cl We are planning to merge this PR. Let us know in case we need to hold.

danoswaltCL (approved):
I pulled this morning and this time Jest tests worked fine when i ran `npm run test` so I'm good. Looks good to me.

ppratikcr7:
Updating a test case for providing awsservice mock. So cahnged title to DO NOT MERGE

VivekFitkariwala (dismissed):

jreddig (approved):

VivekFitkariwala (approved):

Issue #27, opened 2021-01-26
Update enrollment codes to enums


SritterCL:
It looks like the enrollment codes are strings, rather than numbers (enums would be OK).

It also looks like enrollment code was set to null on an error (rather than an error code). Here's my suggestion:
0 - participant was included in experiment
1 - participant excluded due to unspecified error
2 - participant reached experiment point prior to experiment enrolling
3 - participant was on exclusion list
4 - participant’s group was on exclusion list
5 - participant excluded due to group assignment logic
6 - participant excluded from group-assignment experiment - no group specified

(we should try not to use CarnegieLearningWeb/educational-experiment-service#1 - any "common" errors should get their own code).

nirmalpatel:
@SritterCL we had some more discussion on this. The first step towards solving this is to finalize our first list of codes, then publish those as enums in the `upgrade_types` NPM package, and then go on to change the backend code.

Here is your last list of codes where I split out the INCLUDE into ASSIGN and ENROLL:

1. **ASSIGN** - participant was assigned a condition in the experiment
2. **ENROLL** - participant was enrolled in the experiment (as a result of the mark() call)
3. **ERROR** - participant excluded due to unspecified error
4. **REACHED_PRIOR** - participant reached experiment point prior to experiment enrolling
5. **PARTICIPANT_ON_EXCLUSION_LIST** - participant was on the exclusion list
6. **GROUP_ON_EXCLUSION_LIST** - participant’s group was on the exclusion list
7. **EXCLUDED_DUE_TO_GROUP** - participant excluded due to group assignment logic
8. **NO_GROUP_SPECIFIED** - participant excluded from the group-assignment experiment - no group specified

We don't know when we will assign the ERROR as the enrollment code, but we can publish it as an enum for now and find out the logic for it later. Let us know if you think the list is good to go.

@msandbothe if we change the `upgrade_types` NPM package and merge the PR, will it get published automatically in the public NPM repository?

msandbothe:
>     7\. **EXCLUDED_DUE_TO_GROUP** - participant excluded due to group assignment logic

Can we have EXCLUDED_DUE_TO_GROUP_LOGIC

msandbothe:
@jerith666 Do you know the answer to the publishing changes to `upgrade_types` NPM package? (@msandbothe if we change the upgrade_types NPM package and merge the PR, will it get published automatically in the public NPM repository?)

jerith666:
[...]

PR #358, opened 2022-05-11
Fix/issue 342

VivekFitkariwala:
This pull request has some major changes and it is not backward compatible.
### Changes
- Earlier we had assignments tables -> Individual Assignment and Group Assignment these tables are replaced with Individual Enrollment and Group Enrollment. From functionality new enrollment table will store the data of enrollment which means when the user mark the experiment point an enrollment document will be created based on the experiment and experiment user.
- Now the Enrollment table also stores the partition for which it got enrolled so in the future for implementing factorial/partial factorial it will be used.
- Indiviudla Enrollment table now also stores the workingGroupId for experiment with assignment type group. This will remove the Join cost of tables to calculate number of groups enrolled and number of users per group. This will also improve query from the ending criteria of per user per group.
- Queries are simplified because of the above design changes.
- Individual Enrollment also stores the enrollment codes.
- Exclusion table also stores the enrollment codes details [here](https://github.com/CarnegieLearningWeb/UpGrade/issues/27#issuecomment-1122076029)
- Earlier to calculate the enrollment details two tables Monitored Experiment Point and Assignment Table need to be joined to get the data. Now with the above change of Enrollment Table only single table will be queried to get those details so it will improve the database performance overall.
### Things which are left
- Export data part is commented for now because the format is going to get changed so I will be implemented at that time.
- Exclusion code `PARTICIPANT_ON_EXCLUSION_LIST` and `GROUP_ON_EXCLUSION_LIST` is left to be added as this will get affected by segment implementation.


backend/packages/Upgrade/.env.test
```diff
@@ -19,7 +19,7 @@ LOG_OUTPUT=dev
 # PostgreSQL DATABASE
 #
 TYPEORM_CONNECTION=postgres
-TYPEORM_HOST=postgres
+TYPEORM_HOST=localhost
```
ppratikcr7, line 22: Can you revert this back to 'postgres'. This is just required for local testing.

danoswaltCL:
i will try and pull this down and spend some time looking into this today in local docker. before this gets merged, since it is a major breaking change, I think it would make sense to bump base version to 2.0?

danoswaltCL (dismissed):
This looks good to me, I've pulled this down and done some of the locust load tests / manual testing like how @ppratikcr7 demo'd. I'm eager to get this code in and tested in QA, will be easier to spot edge cases. thx @VivekFitkariwala for suggesting and implementing this change, I think it will make a big difference.

ppratikcr7 (dismissed):

ppratikcr7:
Looks good to me as well.

danoswaltCL (dismissed):

ppratikcr7 (approved):

jreddig (approved):
Checked this out locally and ran manual/locust tests. Looks good!

ppratikcr7:
> Checked this out locally and ran manual/locust tests. Looks good!

@jreddig Great. We want to merge this PR but would need a clean db for this. Can you do that for dev env and let us know.

jreddig:
> @jreddig Great. We want to merge this PR but would need a clean db for this. Can you do that for dev env and let us know.

dev db cleared and ready for merge

PR #362, opened 2022-05-13
add http cancel interceptor on user navigating away from page

danoswaltCL:
Currently if user navigates away from a page that has a pending API call, the call still completes in the background even though the payload is no longer needed. If we are having latency issues or slower queries, this could compound issues with performance (as it does currently with the enrollment query in production).

This is a simple interceptor for HTTP requests that will listen for router event that signals that a user has navigated away from page. It will cancel any requests that are pending so we aren't wasting resources.

Attached is short video showing current behavior in QA region (requests always attempting to complete), vs this change in local (pending requests will get cancelled if they are still running when user navigates away.)

Note: In dev console, I use the feature to throttle requests to make them extra slow (simulates 3G connection).

Uploading cancelcompare.mov…


danoswaltCL:
#365

jreddig (approved):

Issue #213, opened 2021-12-09
Auto-refresh experiment data


zackcl:
It would be nice if the (enrolling) experiments can be auto-updated.

Currently, in order to see the updates on enrollment, the user needs to either refresh the page or click on another tab (e.g. Feature Flags) and then click on the "Experiments" tab.
<img width="1145" alt="Screen Shot 2021-12-08 at 7 44 51 PM" src="https://user-images.githubusercontent.com/90279765/145314015-3360c7dc-4e41-451a-88be-89b7a551442b.png">

Also, in order to see the updates on the experiment detail page, the user needs to click on the "Experiments" tab and click on the experiment again to see the change.
<img width="1464" alt="Screen Shot 2021-12-08 at 7 47 43 PM" src="https://user-images.githubusercontent.com/90279765/145314273-ec6a12c3-9a5a-442d-a598-beda0cebeaf1.png">

I'm not sure what would be the best way to implement this in UpGrade, but I believe this will improve the user experience.

danoswaltCL:
This could be accomplished by refetching/polling for new data on every x seconds while the component page is open. Should I pick this up @amurphy-cl ?

amurphy-cl:
That'd be great!

danoswaltCL:
Restating it in 'user story' language helps me out, I'm fully brainwashed in this part of agile. is this what we are saying:

As a user on the details page,
I want to see `/enrollment` and `/details` data updated automatically,
So that I don't have to click refresh.

As a user on the home page,
I want to see `/enrollment` data updated automatically,
So that I don't have to click refresh.

@amurphy-cl @zackcl

jerith666:
I'm pretty sure we have some internal code in `cli-services` that shares requests to avoid re-sending them.  @kschaefe can probably point you to it.

kschaefe:
Yes, the `@UseInFlightCache`[1] decorator in TypeScript will prevent creating a new call while the current one is active.  It is currently limited to calls that take no parameters, but we really need it to work will all idempotent calls.  If you have a need for us to hash and cache calls with params, just let us know.


[1] https://github.dev.carnegielearning.com/CarnegieLearning/cli-services/blob/master/libs/cli-services/src/lib/rest/common/use-in-flight-cache.ts

danoswaltCL:
[...]

PR #368, opened 2022-05-19
Feature/213 autorefresh enrollment details page

danoswaltCL:
In order to enable the experiment detail page to poll every X sec for new data, I need to straighten out the initial data-fetch code, especially for refreshing page. It had a tendency to call the ngrx store in unexpected ways, I believe this will stabilize that behavior so that there are no surprises / bugs when trying to do the next steps.

Wanted to put this in initial PR by itself as it's not totally related to #213.


danoswaltCL:
Closing, had other changes, will just make one PR

PR #369, opened 2022-05-20
Feature/213 autorefresh enrollment details page

danoswaltCL:
This an enhancement to the details page, per issue #213. The user will see almost-live updates to enrollments while on the details page, via polling the `/details` and `/date` endpoints every n seconds in background if the experiment is in `enrolling` status. This polling only occurs while they are on the details page, and will stop if the user changes the status of the experiment to anything other than `enrolling`.

The polling for this is determined by environment variables:

`pollingEnabled` flag to toggle this feature
`pollingInterval` ms between each call (I currently set randomly to 10s, but we can discuss).
`pollingLimit` max number of calls so it doesn't run forever in a forgotten tab

Notes:
I removed table-row animation for opening expanded row, as it looked pretty janky when the data refreshed.
I tested this with refreshing page as well as opening new tabs, etc, it is working appropriately to start/stop the polling.

I set to draft just because I'd like to demo and get feedback before merging, but please still review!


jreddig:
I checked out this branch and ran the locust load tests with the experiment detail page up. Looks great!

jreddig (approved):

PR #387, opened 2022-06-10
Frontend enhancements

ppratikcr7:
This PR contains below changes:

Equally weighted checkbox (O)
Remove number column from tables (O)
Place decision points above conditions (O)
Required checkbox UI for decision points (O)
Show weights with % rounded to 1 decimal place (e.g. 33.3%) (O)
Export experiment popup (one export button instead of two) (O)
Experiment stepper "update" (instead of "save") button when editing (O)
Within-stepper metrics definition (O)
Disable required checkbox for decision points
Remove old Monitor Metrics from Experiment Page
Stepper Dialog box uniform size
Stepper Enhancements (Mostly Design Page)
Partition changed to Decision Point
Upgrade Logo Centric with single version
Server Error Enum spelling correction "CONDITION_NOT_FOUND"
Metrics Description required validation and error message


danoswaltCL:
I am wondering @zackcl @amurphy-cl  if you have some design pictures or figma links and specs for the various things contained within this PR. Seeing some unexpected things (forgive me if it's just my local, I am switching between branches frequently and some cache funkiness could certainly happen).

In the stepper, step 2, 
- The checkbox for equal weight is floating above the table on that column, is this the design?
- The content expands below the modal when things are added and cause a vertical scrollbar on the right.

Screen Shot 2022-06-10 at 11 52 42 AM

- Some notes in the metric step (3) also, some spacing and missing lines under controls:
image

- An error in the console when adding a metric
image

- A valid metric selection is looked greyed out as if 'disabled':
image

- The stepper won't let me skip adding a metric. It seems like this should be optional?

For both step 2 and 3:
- The autocomplete bug has returned (doesn't pop up initially, was something fixed in [363](https://github.com/CarnegieLearningWeb/UpGrade/issues/363))
- The numbering is removed from the tables, but the space remains, would expect that to be adjusted.
- Spacing in rows and between the sections / buttons feels large or inconsistent

In below pic, it's a mix of preexisting bugs (at least, feels off to me): If you click "next" too soon, we get an error message saying we need to fill in some details. It doesn't go away once you've entered something valid, which may be confusing for user.

Also "atleast" is smashed together, and I think we're no longer using the word 'Partition' right?
image

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -448,11 +455,15 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
             (condition, index) => {
+              if (isNaN(condition.assignmentWeight) && condition.assignmentWeight.endsWith("%")) {
+                condition.assignmentWeight = Number(condition.assignmentWeight.slice(0,-1));
+              }
               return this.experimentInfo
                 ? ({ ...this.experimentInfo.conditions[index], ...condition, order: order++ })
                 : ({ id: uuid.v4(), ...condition, name: '', order: order++ });
             }
           );
+          console.log(experimentDesignFormData.conditions);
```
danoswaltCL, line 466: oops console log snuck in

zackcl:
The checkbox for equal weight should be positioned below the "WEIGHT" label.

Experiments - Add Experiment-02-09

In the current design, the texts in the Design step are center-aligned whereas they're left-aligned in the Metrics step. I think maybe it's a good idea to make them all left-aligned to make them look more consistent.

Experiments - Add Experiment-04-16

I think adding more things shouldn't cause the vertical scrollbar on the right side of the modal. I think we can fix this by reducing the maximum height of the internal tables.

I think the metrics table should be more vertically compact like the original Figma design.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
ppratikcr7: Hey @danoswaltCL thanks for looking in depth in this PR. Actually I was still working on various fixes and few commits are yet to be pushed as I wanted to create a draft PR. Still a lot of your detailed feedback would help me a lot. Thanks again. 

Clearing a few things:
1. I just pushed a commit that will allow an empty metric list in stepper to create an experiment and later add them in edit mode.
2. Removed the extra console log in my last commit
3. The floating "equal" checkbox is yet to be improved. I had looked into the figma design, but placing it in header cell below the titles for the mat-table was not working, still have an idea to try again. Just kept it above the title for now.
4.  Disabled "required" as it wont be used as of now. We need to implement this in coordinated site/target.
5. In Edit mode "Grouped and Single" metric keys are kept as disabled and user needs to delete the complete metric if they want to edit it and make it group to single or vice-versa as now we don't have explicit sections for two different metric types and it became very complex to handle change of type in edit mode. I have to comment this in figma design. Rest fields are editable.

Rest things I will make changes accordingly.

ppratikcr7:
> The checkbox for equal weight should be positioned below the "WEIGHT" label.
> 
> Experiments - Add Experiment-02-09
> 
> In the current design, the texts in the Design step are center-aligned whereas they're left-aligned in the Metrics step. I think maybe it's a good idea to make them all left-aligned to make them look more consistent.
> 
> Experiments - Add Experiment-04-16
> 
> I think adding more things shouldn't cause the vertical scrollbar on the right side of the modal. I think we can fix this by reducing the maximum height of the internal tables.
> 
> I think the metrics table should be more vertically compact like the original Figma design.

The floating "equal" checkbox is yet to be improved. I had looked into the figma design, but placing it in header cell below the titles for the mat-table was not working, still have an idea to try again. Just kept it above the title for now.

I will look into the scrollbar stuff.

danoswaltCL:
Yeah adding a checkbox there for equal weight would be challenging with the `mat-table` element, a header row typically wouldn't have something inside or in-between, it's not made to accommodate that.

What about this: I think maybe UX-wise the reason why it feels like there's no natural place for it is because we've neglected to give these two tables any heading/label! 

If we give each table a proper left-aligned heading, it would visually make a lot more sense to have any number of form field controls between the heading and the table.

Something like this (I couldn't figure out how to get into figma, so I slapped this together basically in a Microsoft Paint-like program, ignore the colors/font-sizes....):
Stepper2

It may help also to use the standard normal Mat-Table borders instead of these floating rows with indefinite shape. I like the minimalism of it, but these tables may feel better to interact with if we roll with the Material Design norms: https://material.angular.io/components/table/examples

zackcl:
@danoswaltCL I like the design! I've previously tried adding headings to tables ([Slack Post](https://carnegielearning.slack.com/archives/C02QA168MNW/p1651770885694969)) but I didn't apply this change to the latest design yet because I was a little concerned that there might be too little space for the tables and require too much scrolling. But maybe it's OK since most experiments will only have up to 2 decision points and conditions. I would like to ask @ppratikcr7 if adding the headings would be OK because maybe we're already struggling with the lack of space.

I also like having the checkbox below the heading. I've previously explored ways to position the checkbox ([Slack Post](https://carnegielearning.slack.com/archives/C02QA168MNW/p1645195389641559)) and ended up with the current design, but putting the checkbox below the heading also looks good to me. 

I will create your design in Figma and share it on Slack to ask what people think of it.

ppratikcr7:
> @danoswaltCL I like the design! I've previously tried adding headings to tables ([Slack Post](https://carnegielearning.slack.com/archives/C02QA168MNW/p1651770885694969)) but I didn't apply this change to the latest design yet because I was a little concerned that there might be too little space for the tables and require too much scrolling. But maybe it's OK since most experiments will only have up to 2 decision points and conditions. I would like to ask @ppratikcr7 if adding the headings would be OK because maybe we're already struggling with the lack of space, and I think at least 2 decision points and 2 conditions should be visible without scrolling.
> 
> I also like having the checkbox below the heading. I've previously explored ways to position the checkbox ([Slack Post](https://carnegielearning.slack.com/archives/C02QA168MNW/p1645195389641559)) and ended up with the current design, but putting the checkbox below the heading also looks good to me.
> 
> I will create your design in Figma and share it on Slack to ask what people think of it.

I will check if the space is fine with 2-2 items.

ppratikcr7:
@amurphy-cl @nirmalpatel What is to be used instead of the term "partition" in validation error messages as @danoswaltCL  had rightly pointed out? "Site" or "Target" or any other term?

nirmalpatel:
Yes Decision Point is good 👍

zackcl:
@ppratikcr7 Thanks for trying this out. Here are some things we can consider updating:

1) I think the space below the Back/Close/Next buttons should be consistent across all steps and regardless of the size of the modal. (The second screenshot looks good to me)
2) I think it would be better to center-align the "Required" checkboxes (And maybe the weight values too).
3) I think the Trash icons can be pushed closer to the right side of the table, ideally give an equal space for the left and right side of the table.
4) I think we can consider moving the "Description" column a little to the left so we can give more area for the description.

ppratikcr7:
@zackcl Sticking the button container at button isn't working as it should be. It first shows its initial position and then arranges it in bottom at absolute position, which seems to be very visible transition. Will look for other options. Meanwhile if we need to test other things we can get this PR merged if required.

danoswaltCL:
> @zackcl Sticking the button container at button isn't working as it should be. It first shows its initial position and then arranges it in bottom at absolute position, which seems to be very visible transition. Will look for other options. Meanwhile if we need to test other things we can get this PR merged if required.

Yeah we can review that separately and decide if we want an issue for that after these changes merged in.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -565,7 +565,7 @@ export class ExperimentService {
         // creating queries docs
         const promiseArray = [];
         let queriesDocToSave =
-          (queries &&
+          (queries[0] &&
```
VivekFitkariwala, line 568: What if queries is not defined?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -849,7 +849,7 @@ export class ExperimentService {
       // creating queries docs
       const promiseArray = [];
       let queryDocsToSave =
-        (queries &&
+        (queries[0] &&
```
VivekFitkariwala, line 852: Same as above

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: It will store queries as [] blank array in the experiment json object

ppratikcr7:
> As you might already know, a table with the "Except" heading should show up when the inclusion criteria is set to "Include Specific" in the Participants step. And there should be a "Defined Metrics" heading for the table in the Metrics step for consistency in design (all tables have headings). Also, I think the Metrics table should be as compact as the below design.
> 
> Experiments - Add Experiment-03-01
> 
> Experiments - Add Experiment-04-16

Participants Page changes would be in segments PR and it has that "Except" heading. I will add "Defined Metrics" Heading on Metrics Page. Also about the compactness. I was not able to set variable row height as per the number of operation fields. So, I have kept the max height to accommodate max fields as shown below:
Screenshot 2022-06-14 at 2 18 37 PM

I will try to look for more solutions later.

danoswaltCL:
@ppratikcr7 is this ready for review or are more updates coming?

ppratikcr7:
@danoswaltCL Yes, it's done. You can review now.

PR #394, opened 2022-06-15
missing await on delete

jreddig:


jerith666:
FWIW, I've found https://typescript-eslint.io/rules/no-floating-promises/ to be of great help in preventing this kind of mistake.

danoswaltCL:
interestingly, this only popped up as an actual code issue when we tried unit testing a scenario in node 16 vs node 12, the unit test breaks in 16, maybe node 12 both allowed and 'fixed' accidental implicit awaits? can't find the reason in brief googling. actual code may be edge-case enough that it never was found previously.

but yes, we ought to up our linting in general, the codebase is still on tslint

danoswaltCL (approved):

PR #395, opened 2022-06-15
Fix/GitHub runner version

danoswaltCL:
I did have these changes included in Angular branch but better to have them in own PR first.


danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/pull/394 should the solve backend unit-test runner fail; this should get merged before then angular branch update to be sure all CI/CD and backend test-check things are good-to-go with node 16.15.0. That will make sure the angular PR can just represent frontend changes alone without these things.

PR #400, opened 2022-06-22
env var and base-url refactor

danoswaltCL:
This update is intended to resolve a few awkward things regarding env vars and how they are being used on the frontend. It looks like a lot of files, but mostly it's same small change in all of them for importing `environment` in a more flexible way...

Problems that needed addressed for angular 13 update. These used to be possible, but the compilation happens differently now:
- Use of `window` in `environment.ts` prod builds.
- String interpolation of `endpointApi` value.

What these changes do:
- Eliminate use of jQuery on index.html to fetch `environment.json` details
    - Uses `APP_INITIALIZER` to accomplish that data fetch during app bootstrap hook at runtime.

- Creates interface declarations for strong `Environment` typing

- Creates an `InjectionToken` for environment object that essentially turns the `environment.ts` object into a "normal" injectable Angular dependency.
    - Enables ability to remove use of global `window` object.
    - Makes mocking `environment` object much more straightforward when needed for conditionals.
    - Generally normalizes how `environment` is used as dependency in class constructors instead of a direct import.
    
- Creates a Base-url http interceptor
    - Prepends api-host url to outgoing API requests
    - Allows to remove string interpolation of `endpointApi` in environment.ts
    - I changed the term `endpointApi` to `apiBaseUrl`, as the term `entpointApi` seemed inaccurate to me. Let me know if you disagree, I can change it back!
    
- Updates all specs to use mockEnvironment strategy
- Updates all `environment.ts` type files to remove string interpolation and update usage of `apiBaseUrl`.

Lastly, this will allow a future possibility of getting moving env var fetching to the build or deployment process directly from host OS or job-runner. At that point, we'd be able to forgo fetching details at bootstrap and just delete the `APP_ITIALIZER` provider and discard use of `enviroment.json`.


jreddig (approved):
Looks good to me!

Issue #413, opened 2022-06-28
Stepper Buttons / Modal height


danoswaltCL:
**Describe the bug**
This is somewhere between a bug and a feature request!

The Create Experiment modal has been given a fixed height, which is probably a good thing for consistency between steps, but now the nav buttons aren't sticky to the bottom as might be expected, which creates some UX awkwardness with buttons floating around wherever the content ends.

**To Reproduce**
1. Create a new experiment and move through the steps and observe the bottom nav buttons

**Expected behavior**
The expectation would be that these buttons stay put. 

**Screenshots**
For instance:
<img width="981" alt="image" src="https://user-images.githubusercontent.com/97542869/176287768-da7d4688-328d-4ac1-9620-33d4cd8dc41f.png">

danoswaltCL:
i can take this if we agree that sticking with a fixed modal height is fine with buttons aligned to stay put at the bottom?

i'd say though that this change may take a little time due to refactoring some stuff, there's potential for it being more complicated than it appears on surface. if there are more important things to tackle I can look elsewhere.

@amurphy-cl @zackcl @mswartzCL

PR #442, opened 2022-07-11
Feature/consistent modal button styles

danoswaltCL:
#413

As a user, one will now see the buttons always at bottom of experiment-stepper.

Part of this fix is a small refactor to generalize modal styles / basic html layout and shared naming.

You'll see a new `shared-modal.scss` stylesheet, which at this point is imported into the global `style.scss` so the following classes can be made to all modals automatically at the component-level.

These were only applied to experiment-stepper to start with as it was most important. I will look into expanding this where needed for other modals where it makes sense so all is consistent. FYI, you will see some file changes where `modal-btn` was already a global style class, I just renamed it to `shared-modal--modal-btn` and tucked it into this new stylesheet because it was already serving this purpose.

Basic layout classes:
```
div.shared-modal--step-container
   section.shared-modal--form-container
   section.shared-modal--buttons-container
```

Bottom modal button:
```
shared-modal--buttons-left
shared-modal--buttons-right
shared-modal--modal-btn
```

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/components/flag-overview/flag-overview.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/components/flag-overview/flag-overview.component.html
@@ -49,7 +49,7 @@
 
   <div class="button-container">
     <button
-      class="modal-btn"
+      class="shared-modal--modal-btn"
       mat-raised-button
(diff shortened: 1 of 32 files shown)


jreddig (commented):
I like this! it looks really clean

jreddig (approved):

PR #446, opened 2022-07-13
Release1 Bug Fixes for Metrics Stepper Page

ppratikcr7:
Draft PR for metrics UI changes. It is still work in progress. Planning to resolve multiple tickets related to metric stepper in one PR. 

This PR includes:
1. Dynamic Vertical spacing
2. Font Size increased on metrics stepper
3. Validation error messages for required statistics values
4. Metric, Key, Metric - Placeholder logic
5. Editing mode change for equal weight checkbox got reverted in some old PR
6. Metrics will not be disabled and grayed out from now on
7. Metric Key Hierarchy and repeated measure fields will be shown at once when we select a repeated Metric
8. Editing Metric-Key should clear future selections
9. Editing Metric-Key should update Single and Repeated Metric Input format
10. Updating Metrics once saved should not have an issue opening auto-complete.
11. Added validation for metrics/key required once the user fills in stats or query name.
12. Removal of queries using `slice` instead of `delete` keyword to maintain proper query index.
13. Refactoring code and renaming variables and functions with proper naming and adding more comments for better code understanding. 

The changes in the ts file may seem too much but it's actually code blocks being refactored to proper locations as well. Sorry for that.

Screenshot 2022-07-15 at 8 20 01 PM


ppratikcr7:
@danoswaltCL Removing Zack from reviewer. I am still working on it. Only the above mentioned points in the PR description are covered. In case you will try to test few other bugs for e.g. editing the metric should update the other fields. That won't work for now. If you want to test the mentioned points in this PR, then go ahead and let me know in case of any issues there.

danoswaltCL:
oh okay, I just wanted to make sure @zackcl saw this

danoswaltCL:
hey @ppratikcr7, @zackcl and I have pulled down the code and tried it out, I think we had some comments but I assume you're still working on it, should we share things that we've noticed or wait until you're done? (It seems a lot closer but we observed a couple things with the mastery workspace selection, this seems tricky to implement for sure)

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -55,13 +57,18 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
   metricsDataSource = new BehaviorSubject<AbstractControl[]>([]);
 
   metricsDisplayedColumns = ['keys', 'operationType', 'queryName', 'removeMetric'];
-  queryIndex = 0;
+  queryIndex: number = 0;
+  editMode: Boolean = false;
```
danoswaltCL, line 61: we want lowercase `boolean`. `Boolean` refers to the non-primitive `new Boolean()` object, which isn't quite right. https://stackoverflow.com/questions/15487220/typescript-primitive-types-any-difference-between-the-types-number-and-numbe

ppratikcr7: Nice catch, will update this. Thanks

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -385,28 +321,170 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
...
   }
 
+  checkStatisticRequiredError(statistic: any) {
+    if (statistic == null || statistic === '') {
+      this.queryStatisticError.push(true);
+    }
+  }
+
+  checkComparisonStatisticRequiredError(compareFn: any, compareValue: any) {
+    if (compareFn == null || compareFn === '' || compareValue == null || compareValue === '') {
```
danoswaltCL, line 343: i'm seeing a number of places with null-checking specific things, but this is actually less type-safe than just using the existence-checking pattern of `if (!compareFn || !compareValue)`. Empty string `''` is a falsey value, just like `null`, they can both be covered this way. The current code will allow `undefined`, `NaN`, `0`, `false` values to slip through.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -385,28 +321,170 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
     return metric;
   }
 
-  checkQueryNameError(queryName){
+  checkMetricKeyRequiredError(metricKeyMissing: Boolean){
```
danoswaltCL, line 324: `boolean` here too

ppratikcr7:
@danoswaltCL Let me know once you are done with the review, so that I will make the changes in a single commit.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -385,28 +321,170 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
...
+
+  checkComparisonStatisticRequiredError(compareFn: any, compareValue: any) {
+    if (compareFn == null || compareFn === '' || compareValue == null || compareValue === '') {
+      this.queryComparisonStatisticError.push(true);
+    }
+  }
+
+  ManageKeysControl(queryIndex: number, keyIndex: number) {
+    let keysArray;
+    if (queryIndex == null) {
```
danoswaltCL, line 350: also seeing lots of double-equals used for comparison, usually it's technically fine but I think the advice of always using the strict triple-equals in javascript/typescript comes into play in examples like this spot.

Javascript does super weird stuff with double-equals, which one has to have it all memorized if you want to use double-equals. For instance in this spot, if what if `queryIndex` is `undefined`? Surprisingly, this would be true. What if it's `NaN` or empty string? It's false. Maybe that is fine for the code in this spot, but it's confusing. If you really are just checking for `null` and only `null`, you need `===`.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -385,28 +321,170 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
...
+      queryIndex = this.queryIndex;
+    }
+    // Prepare filteredMetrics for each query and its keys for experimentInfo while in edit Mode
+    if (this.experimentInfo) {
+      if (this.experimentInfo.queries.length > 0) {
```
danoswaltCL, line 355: This block looks like a duplicate on both sides of the `if/else`? I don't see the difference:

```typescript
keysArray = this.queries.at(queryIndex).get('keys') as FormArray;
        this.filteredMetrics$[keyIndex] = keysArray.at(keyIndex).get('metricKey').valueChanges
        .pipe(
          startWith<string>(''),
          map(key => {
            if (keyIndex - 1 >= 0) {
              const { metricKey } = keysArray.at(keyIndex - 1).value;
              this.options = metricKey.children;
            }
            return key ? this._filter(key) : this.options ? this.options.slice() : [];
          })
        );
```

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
ppratikcr7: At some places, I pass `null` as a param to ManageKeysControl to be able to take global queryIndex. But, just now observed that I can pass global queryIndex directly.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
ppratikcr7: Again nice catch. Actually, earlier there was a difference in both sides of the `if/else` but with time changes were made upon debugging and finally it would have turned out to be exact same. My bad to have skipped this refactoring. I will do that.  Thanks :)

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -385,28 +321,170 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
...
+        this.queryStatisticError = []; 
+        this.queryComparisonStatisticError = [];
+        this.queryNameError = [];
         const monitoredMetricsFormData = this.queryForm.getRawValue();
         monitoredMetricsFormData.queries = monitoredMetricsFormData.queries.map(
           (query, index) => {
             let { keys, operationType, queryName, compareFn, compareValue, repeatedMeasure } = query;
             if (keys) {
-              keys = keys.filter((key) => key.metricKey !== null).map(key => key.metricKey.key);
+              keys = keys.filter((key) => key.metricKey !== null).map(key => key.metricKey.key ? key.metricKey.key : key.metricKey);
```
danoswaltCL, line 480: what if `key.metricKey` is `undefined` or other falsey values? It will only filter out if it's exactly `null` here.

ppratikcr7: We push `null` for metrickey on adding a new query, so I would have filtered with `null`

jreddig:
Select a group metric,

Screen Shot 2022-07-22 at 10 24 26 AM

Delete the metric and add a new metric. The "Most recent" selection remains. But future metrics added do not have this drop down.
Screen Shot 2022-07-22 at 10 24 52 AM

If you delete and re-add the metrics, the first metric will always have this "Most recent" selection on creation. This doesn't appear to affect metric calculations or stick around for future edits. If you save the experiment and then close the stepper, the "Most recent" option doesn't appear on future edits. Not sure how important this is since it doesn't affect the experiment, but I wanted to make a note of it.

ppratikcr7:
> Select a group metric,
> 
> Screen Shot 2022-07-22 at 10 24 26 AM
> 
> Delete the metric and add a new metric. The "Most recent" selection remains. But future metrics added do not have this drop down. Screen Shot 2022-07-22 at 10 24 52 AM
> 
> If you delete and re-add the metrics, the first metric will always have this "Most recent" selection on creation. This doesn't appear to affect metric calculations or stick around for future edits. If you save the experiment and then close the stepper, the "Most recent" option doesn't appear on future edits. Not sure how important this is since it doesn't affect the experiment, but I wanted to make a note of it.

Okay, will look into this.

danoswaltCL:
hi @ppratikcr7 , what is the status of this PR?

ppratikcr7:
> hi @ppratikcr7 , what is the status of this PR?

Hi @danoswaltCL, I have made a commit to resolve the above code review points. I have also made a flex change to see the dropdown arrows. Also, trying to refactor the code even further.
Screenshot 2022-07-26 at 5 30 41 AM

danoswaltCL (approved):

PR #450, opened 2022-07-19
Increasing limit for segment import

RidhamShah:
Increased segment import limit and removed unnecessary 'status' checks while importing a segment


jreddig (approved):
This looks good in my local! I imported a very large user segment and ran some initial load tests. On my local environment, initial load test results seem promising but I'd like to see how the caching behaves in region.

VivekFitkariwala:
New environment variable is added `CACHING_ENABLED`. If the value is true app will use in memory caching to cache the segment data.
We have used [cache manager](https://www.npmjs.com/package/cache-manager) as caching manager and the store can be switched with other services if required. The idea is to use in-memory cache at present and might replace with redis in future

VivekFitkariwala (approved):

PR #480, opened 2022-07-29
fix for metrics validation on count statistic

ppratikcr7:
This PR contains the fix for count statistic validation fix on metrics stepper. It also contains a deep validation for metric/key required for moving forward.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
@@ -321,9 +321,11 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
     return metric;
   }
 
-  checkMetricKeyRequiredError(metricKeyMissing: boolean){
-    if (metricKeyMissing) {
-      this.queryMetricKeyError.push(true);
+  checkMetricKeyRequiredError(metricKeys: any){
+    for (let i = 0; i < metricKeys.length; i++) {
+      if (metricKeys[i].metricKey === null) {
+        this.queryMetricKeyError.push(true);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -457,11 +459,14 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
           (query, index) => {
```
danoswaltCL, line 459: if i'm reading this right, this isn't a `map` it's just a `forEach` because we never return anything.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -492,7 +497,6 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
                     : ({...this.removeMetricName(queryObj)})
                   );
               } else {
```
danoswaltCL, line 499: I don't think these `else` blocks are doing anything, can they both be removed?

danoswaltCL (changes requested):

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -321,9 +321,11 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
     return metric;
   }
 
-  checkMetricKeyRequiredError(metricKeyMissing: boolean){
-    if (metricKeyMissing) {
-      this.queryMetricKeyError.push(true);
+  checkMetricKeyRequiredError(metricKeys: any){
+    for (let i = 0; i < metricKeys.length; i++) {
+      if (metricKeys[i].metricKey === null) {
```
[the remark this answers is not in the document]
ppratikcr7: Actually we insert null for empty nodes when a repeated metrics is selected. But, still if you say so we can catch all the falsely cases using your condition. I will make the change.

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
ppratikcr7: correct

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
ppratikcr7: correct, I will update

ppratikcr7: correct

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
danoswaltCL: okay my bad, I think I got confused because sometimes it doesn't return anything when `keys` is falsey (well, it does, it returns `undefined` since the `return` is still implied even when removed), so we would be potentially building an array of `undefined`s. Maybe that just works though, the code has lots of conditionals in it, trying my best to follow, I wonder if in my mind I'm trying to turn it into a `forEach` where `if (keys)` and `keys.length` are both true after filtering, then `push` onto the `monitoredMetricsFormData.queries` array, otherwise do nothing... but if this works in test cases, i'm okay with it.

danoswaltCL (approved):

PR #487, opened 2022-08-03
fixed broken repeated metrics reselection behaviour

ppratikcr7:
This PR also contains code refactoring for proper autocomplete list for each row separately. I didn't came across any good variable name to differentiate the 3 filteredMetrics Observables (each for Metric/key/Metric of a repeat Metrics). If someone can suggest one, I will change that. Thanks :)


danoswaltCL (changes requested):
The issue with the ghost fields appearing is fixed, but still has issues other when re-selecting after a `masterWorkspace` type node has been chosen. For instance, if I select `masteryWorkspace` and then select a simple metric like `totalSeconds`, and then select `masteryWorkspace` again, it breaks:

image

Also try selecting `masteryWorkspace` then deleting it, it autofills with the workspace metric-keys afterward:

image

ppratikcr7:
@danoswaltCL Checking on this.

ppratikcr7:
@danoswaltCL Updated the PR. Can you please check now.

danoswaltCL:
i'm still seeing this after selecting `masteryworkspace` and then deleting, the autofill options are incorrect

https://user-images.githubusercontent.com/97542869/182926318-d09673d7-643e-4064-a6b2-5ac754301741.mov

:

danoswaltCL:
sorry I'm also seeing this, haven't seen this one, if selecting `masteryWorkspace` and picking the key and metric, then delete the metric and key. on deleting the key, then suddenly the metrics data disappears or something and the screen goes blank:

https://user-images.githubusercontent.com/97542869/182929532-70ea1d5b-fa12-499a-b706-2ed3bba80e5d.mov

danoswaltCL:
for the second issue, I believe that the reason is just that the `ngIf` for displaying that message should be `allMetrics` and not `options`.

```html
<ng-container *ngIf="allMetrics?.length; else noAvailableMetricsMessage">
```
instead of 
```html
<ng-container *ngIf="options?.length; else noAvailableMetricsMessage">
```

ppratikcr7:
@danoswaltCL Can you please review it once, I have updates in this PR. It also has two test case imports changes done on behalf of Jenn's test cases. :)

danoswaltCL (changes requested):
@ppratikcr7 ty for going through this so much, I think the bugs that I have reported have been fixed, there are a couple edge-case things I see that cause error in console but are not important for the functionality, not worried about that.

So looking good I think from the requirements, I hate to ask for any more changes, but this is just a refactor for readability request: the `selectedOption` method is growing near 100 lines with all the tricky nested cases, can you break out some of these code blocks into smaller functions with descriptive names? I think that would help out a lot for reading through all of it. I appreciate all the comments though, that helps too.

ppratikcr7:
> @ppratikcr7 ty for going through this so much, I think the bugs that I have reported have been fixed, there are a couple edge-case things I see that cause error in console but are not important for the functionality, not worried about that.
> 
> So looking good I think from the requirements, I hate to ask for any more changes, but this is just a refactor for readability request: the `selectedOption` method is growing near 100 lines with all the tricky nested cases, can you break out some of these code blocks into smaller functions with descriptive names? I think that would help out a lot for reading through all of it. I appreciate all the comments though, that helps too.

Sure, no issues. Will do that. I felt the same, it was growing to long.

danoswaltCL (approved):

PR #492, opened 2022-08-03
disable weights sum check on equalweight checkbox selection

ppratikcr7:
This PR also handles sum of weights on manual values for more conditions as well.

Screenshot 2022-08-04 at 1 23 47 AM


danoswaltCL:
This solves for validation, but I think there's a bigger issue, can we hold off on this one? I'd like to ask about this in thurs mtg. We're sending the rounded, single-decimal UI display values instead of the true equally-weighted precise(ish) floating point numbers we need to send, and as you can see in the pic, for some cases, the weights aren't actually equal, which will mess up the experiment.

We can definitely handle this differently in the UI, the form control would just be an object instead of a number, and that object would have a `display` property we use to round to 1 decimal, and `rawValue` for the actual floating point decimal that gets posted to backend (with some number of decimals).

BUT this feels like a thing the UI typically wouldn't be calculating in the first place really. `Weighted equally` should be business logic for the backend to deal with.

@nirmalpatel @SritterCL @amurphy-cl @mswartzCL can we make this a topic in mtg?

nirmalpatel:
@danoswaltCL IMO we can display rounded-off values and store up to 5 decimal points in the frontend and send the floats to backend. The equally weighted option with an odd number of conditions means that the last condition will have weight = 100 minus sum(rest of the conditions). I don't think we need to split the condition weight calculation code across frontend and backend...

ppratikcr7:
> This solves for validation, but I think there's a bigger issue, can we hold off on this one? I'd like to ask about this in thurs mtg. We're sending the rounded, single-decimal UI display values instead of the true equally-weighted precise(ish) floating point numbers we need to send, and as you can see in the pic, for some cases, the weights aren't actually equal, which will mess up the experiment.
> 
> We can definitely handle this differently in the UI, the form control would just be an object instead of a number, and that object would have a `display` property we use to round to 1 decimal, and `rawValue` for the actual floating point decimal that gets posted to backend (with some number of decimals).
> 
> BUT this feels like a thing the UI typically wouldn't be calculating in the first place really. `Weighted equally` should be business logic for the backend to deal with.
> 
> @nirmalpatel @SritterCL @amurphy-cl @mswartzCL can we make this a topic in mtg?

The one different condition weight is due to manual weight entry to make the total sum to be 100. When we select Equal Weight checkbox, all conditions remain 7.1 and no validation is done. This is as per nirmal's last comment.

ppratikcr7:
@zackcl @danoswaltCL Maybe we can merge this if the current PR resolves this 14 condition for now. We can track the relative weights approach in a separate [issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/500).

danoswaltCL:
ok we can do that

danoswaltCL (approved):

PR #494, opened 2022-08-03
backend partition renaming in comments and locally scoped variable names

danoswaltCL:
This change just updates `partition` to `decision point` in backend .ts files, focusing on the easy ones, comments and locally scoped variables that don't affect other things.


zackcl (approved):

danoswaltCL:
@VivekFitkariwala @ppratikcr7 please take a look and make sure this won't cause issues

PR #499, opened 2022-08-08
Make the Target field required

zackcl:
Changes:
  * Make the Target field required in the design step
  * An error message will show up if the user clicks "Next" while any of the required fields are empty

Screen Shot 2022-08-08 at 3 49 23 PM

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
@@ -239,7 +239,7 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
   addPartitions(site = null, target = null, description = '', order = null) {
     return this._formBuilder.group({
       site: [site, Validators.required],
-      target: [target],
+      target: [target, Validators.required],
       description: [description],
       order: [order]
     });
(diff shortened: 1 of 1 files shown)


zackcl:
@ppratikcr7  If I just enter spaces for the Site/Target/Name fields and then click the "Next" button, it shows the error messages but still goes to the next step which it shouldn't. I would appreciate it if you could give me a hint on how to fix this (Or you can fix it if it's simple enough).

Screen Shot 2022-08-08 at 3 58 29 PM

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -388,25 +388,19 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
   validateConditionCount(conditions: ExperimentCondition[]) {
     const conditionCountErrorMsg = this.translate.instant('home.new-experiment.design.condition-count-new-exp-error.text');
-    if (conditions.length >= 0) {
-      if(conditions.length == 0) {
-        this.conditionCountError = conditionCountErrorMsg;
-      } else if (conditions.length >= 1) {
-        const conditionWeight = conditions.map(condition => condition.assignmentWeight);
-        !conditionWeight[0] ? this.conditionCountError = conditionCountErrorMsg : this.conditionCountError = null;
-      }
+    if (conditions.length === 0 || !conditions.every(condition => condition.conditionCode && condition.assignmentWeight && condition.conditionCode.trim())) {
```
danoswaltCL, line 391: 💯

zackcl:
I've also changed the `Weight` field's input type to `number` and removed the arrows/spinners (also disabled the autocomplete suggestions). After this fix, the validation code `condition.assignmentWeight !== null` works as expected (Previously, it didn't work if the user has entered an empty string).

ppratikcr7:
> @ppratikcr7 If I just enter spaces for the Site/Target/Name fields and then click the "Next" button, it shows the error messages but still goes to the next step which it shouldn't. I would appreciate it if you could give me a hint on how to fix this.
> 
> Screen Shot 2022-08-08 at 3 58 29 PM

Sorry saw this message a bit late. Hope it's resolved now.

zackcl:
> Sorry saw this message a bit late. Hope it's resolved now.

No, it's not resolved yet. Do you know how I can prevent from going to the next step after entering spaces in the Site/Target/Name fields? (like the above screenshot) I see error messages if I come back to the design step but it still goes to the next step when I click the "Next" button.

danoswaltCL:
> > Sorry saw this message a bit late. Hope it's resolved now.
> 
> @ppratikcr7 No, it's not resolved yet. Do you know how I can prevent from going to the next step after entering spaces in the Site/Target/Name fields? (like the above screenshot) I see error messages if I come back to the design step but it still goes to the next step when I click the "Next" button.

@zackcl Line 476 has `this.experimentDesignForm.valid`, which should be `false` if the inputs are actually empty. The `trim()` that you added helps determine that a message is needed, but it isn't modifying the existing string, so angular still sees the form as valid.

What may actually be more helpful for all of this is to save ourselves these checks later down the line by adding the `trim` functionality further up the chain on the actual value. Other solution is to manually invalidate the form, but that I don't like.

ppratikcr7:
> > Sorry saw this message a bit late. Hope it's resolved now.
> 
> @ppratikcr7 No, it's not resolved yet. Do you know how I can prevent from going to the next step after entering spaces in the Site/Target/Name fields? (like the above screenshot) I see error messages if I come back to the design step but it still goes to the next step when I click the "Next" button.

Oh okay. Actually, in the metrics stepper component, I check for errors and when a validation error occurs, I push a flag in an error list and at the end emit the event only when the list has no errors. This may not be the best way but serves the purpose.

Screenshot 2022-08-10 at 8 24 46 PM

zackcl:
@ppratikcr7 Thank you! I was able to fix this by making it emit the event when both `this.partitionCountError` and `this.conditionCountError` are `null`: https://github.com/CarnegieLearningWeb/UpGrade/pull/499/commits/bd992beb1b496ed8d5f373a2500e9cd0d1197aac

zackcl:
Please feel free to merge this.

danoswaltCL (approved):

PR #503, opened 2022-08-10
Refine the Metrics step UI

zackcl:
Refined the Metrics step UI so that it looks more consistent with the other steps (e.g. same font size) and more similar to the [Figma design](https://www.figma.com/file/uYxnCT2dXTftSZKFQ00Q0y/UpGrade-UI-Prototype-(latest)). 

Before:
before

After:
after2


zackcl:
I wanted to vertically align the rows within cells to the top instead of the center (like the screenshot) when `masteryWorkspace` is selected, but I'm not sure how to do this properly. I would appreciate it if anyone knows a simple way to do this. Otherwise, we can continue with the current design for now as it's not super important.

Experiments - Edit Experiment-04

danoswaltCL (approved):

Issue #275, opened 2022-02-25
Unify /api/mark and /api/failed


jerith666:
Right now Mathia will always make some callback to UpGrade when it hits the SelectSection experiment point: either `/api/mark` if the (a) no experiment was running or (b) an experiment was running and Mathia was able to carry out the condition that UpGrade asked it to; or `/api/failed` if (c) an experiment was running but Mathia was not able to carry out the condition that UpGrade asked it to.

This is at least bad ergonomics for developers integrating with UpGrade: it’s not clear if you should call `/api/mark` XOR `/api/failed`, vs. `/api/mark` AND MAYBE ALSO `/api/failed`.  It also seems to be the case that UpGrade isn’t interpreting a lone call to `/api/failed` as a student encountering the experiment point, since such students didn’t show up on the experiment dashboard after the testing done on Dec. 16.

So, we think we ought to instead do away with `/api/failed` as a separate API call, and add to `/api/mark` a way to pass the status of the user’s encounter with the experiment point.  That status is either:

1. no experiment running
1. experiment running, experiment condition successfully applied
1. experiment running, experiment condition could not be applied

This both makes it clear to devs integrating with UpGrade that `/api/mark` should always be called; and makes sure that UpGrade has clear information about the state of each user WRT each experiment point.

cc @SritterCL @amurphy-cl

amurphy-cl:
Ok, this makes sense to me.

SritterCL:
Yeah, I like that idea.

VivekFitkariwala:
I agree on removing `/api/failed`.
But why we should get information which we already have or which is not Upgrade specific via `/api/mark`?

Upgrade already knows if an experiment is running or not or if we even have created an experiment for that site.
And is condition applied successfully or not is not Upgrade specific data. It's client specific data and it should be logged on client application.

jerith666:
In condition 2, the user should be considered successfully enrolled in the experiment: it makes sense for UpGrade to include that user in the metrics it displays for the experiment, etc.

[...]

PR #507, opened 2022-08-15
Adding status and removing failedExperimentPoint from clientlib

VivekFitkariwala:
- Removed `failedExperimentPoint` from js client side SDK
- Added `status` parameter inside `markExperimentPoint` with default value `marked`. Accepted values are 
  - applied
  - not-applied
  - marked
- Experiment user is excluded if status is `not-applied`


jreddig (commented):
The changes look good in my local! @mswartzCL are we good to merge this in? status isn't required so I believe MathIA will be unaffected. We'll then need to organize with engineering to make their changes.

mswartzCL:
@jreddig Sounds like it should be fine to merge it in. We can always roll it back out if needed. Normally, I'd suggest we check with @jerith666 , but he's on vacation for the rest of the week.

types/src/Experiment/enums.ts
```diff
@@ -44,6 +44,12 @@ export enum SERVER_ERROR {
   CONDITION_NOT_FOUND = 'Condition not found',
 }
 
+export enum MARKED_DECISION_POINT_STATUS {
+  APPLIED = 'applied',
+  NOT_APPLIED = 'not-applied',
+  MARKED = 'marked',
```
jerith666, line 50: I had to go back and read the discussion in #275 to remember what the difference is between `MARKED` and `APPLIED`.  To me, that indicates these are poorly named -- or at the very least need some JSDoc to tell future developers what they mean.

What about `CONDITION_APPLIED`, `CONDITION_FAILED_TO_APPLY`, and `NO_CONDITION_ASSIGNED`?  Is that too verbose?

jerith666 (changes requested):

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.ts
```diff
@@ -14,4 +15,8 @@ export class MarkExperimentValidator {
   @IsNotEmpty()
   @IsDefined()
   public condition: string | null;
+
+  @IsNotEmpty()
+  @IsDefined()
+  public status: MARKED_DECISION_POINT_STATUS;
```
jerith666, line 21: Isn't this going to require clients to supply a value?  That's contrary to the strategy outlined in https://github.com/CarnegieLearningWeb/UpGrade/issues/275#issuecomment-1122783575 -- since Mathia won't immediately start using the new Java client library (indeed, we don't even _have_ a new Java client library yet! :) ), it will be a while before most API calls include this parameter.

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -120,7 +120,7 @@ export default class UpgradeClient {
     async markExperimentPoint(
       experimentPoint: string,
       condition = null,
-      status: MARKED_DECISION_POINT_STATUS,
+      status: MARKED_DECISION_POINT_STATUS = MARKED_DECISION_POINT_STATUS.MARKED,
```
jerith666, line 123: I think the client library should require its users to supply a value for this parameter.

jerith666 (changes requested):

types/src/Experiment/enums.ts
VivekFitkariwala: This looks good to me. I will update the code accordingly

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.ts
VivekFitkariwala: I was thinking just in terms js Client and made it a required parameter. I will make it optional until we update Java client.

VivekFitkariwala:
@jerith666 PR now includes your feedback.

jerith666 (approved):
Thanks, looks good to me now!

PR #510, opened 2022-08-16
remove json-editor deps and manual add-metrics feature from profile

danoswaltCL:
Removing the barely json-editor module feature saves us a huge amount of space. Running `npm run analyze` in the frontend package.json generates a visual and statistical breakdown of module sizes.

In my amateur analysis using this tool, you can see how comically large the json-editor is, essentially the same size as both the Angular runtime and the rest of the app. Removing reduces our gzipped package size by 33% from 7.5kb to 5kb.

Screen Shot 2022-08-16 at 2 24 49 PM
Screen Shot 2022-08-16 at 2 24 59 PM


jreddig (approved):
:clap:

PR #511, opened 2022-08-16
adds options to experiment with making jest faster, ignore skipped te…

danoswaltCL:
Experimenting with Jest test runner to bring down test times on frontend


jreddig (approved):
checks passed in 2 minutes! i like that.

PR #517, opened 2022-08-19
add design type selector on overview with just simple experiment for now

danoswaltCL:


types/src/Experiment/enums.ts
```diff
@@ -4,6 +4,10 @@ export enum CONSISTENCY_RULE {
   GROUP = 'group',
 }
 
+export enum DESIGN_TYPE {
```
danoswaltCL, line 7: This is just placeholder, I want an enum for this value but I don't know if we'll need anything for this value that coordinates with backend, can be change later.

danoswaltCL:
i am going to fold this into other branch I'm working on.

PR #524, opened 2022-08-30
275 java clientlib

jreddig:
Added status parameter to `markExperimentPoint`. Overloaded the function so current use with no status parameter defaults to `CONDITION_APPLIED`.

Please check my enum :) Sending the String results in a `CONDITION_NOT_APPLIED` status to exclude the student, and the other two statuses enroll the student.


clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -274,8 +274,13 @@ public void markExperimentPoint(final String experimentPoint, String condition,
 
 	public void markExperimentPoint(final String experimentPoint, String experimentId, String condition,
 			final ResponseCallback<MarkExperimentPoint> callbacks) {
+		markExperimentPoint(experimentPoint, "", condition, MarkedDecisionPointStatus.CONDITION_APPLIED, callbacks);
```
jerith666, line 277: same as in #507 WRT the TS client, I think users of the Java client should be required to supply a status -- don't provide a default, make the developer who's integrating with UpGrade think this through.

jerith666 (changes requested):

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
jreddig: @jerith666 made the appropriate change! Take another look and let us know if you're comfortable with this code going to production :)

jerith666 (changes requested):
You need to bump the version in `pom.xml` so that a new artifact can be delpoyed to nexus and we can reference it in Mathia code.  Otherwise, this looks good to me.

jerith666 (approved):
A breaking API change, so major version change is appropriate.  Thanks!

PR #526, opened 2022-09-01
Bugfix/issue 245: Exclude If Reached and coordinated DP

ppratikcr7:
This PR contains frontend and baackend changes for exclude if reached checkbox param for decision points.


ppratikcr7:
@jreddig There is a unit test case related to the insertion of decision point. I tried including the new param in your test case but still the test case fails. If you can help with that, it would be great. :)

ppratikcr7:
@jreddig Thanks a lot Jenn for the quick fix. :)

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.html
```diff
@@ -75,31 +75,32 @@
...
               <mat-checkbox
-                      class="ft-8-100"
-                      color="primary"
-                      [disabled]="true"
-                      >
+              class="ft-8-100"
+              color="primary"
+              [disabled]= "experimentInfo && (experimentInfo.state == ExperimentState.ENROLLING || experimentInfo.state == ExperimentState.ENROLLMENT_COMPLETE)"
+              [checked]="this.partition.at(groupIndex).get('excludeIfReached').value"
+              (change)="setExcludeIfReachedFlag($event,groupIndex)"
```
danoswaltCL, line 93: Is this checkbox not registered as part of the form? we shouldn't have to manually handle setting the form value, something seems off with a (change) handler in a reactive form pattern. is there something special we're doing with this case?

danoswaltCL: meaning if you use `[formControlName] = "excludeIfReached"`, I don't think a `(change)` should be needed?

danoswaltCL:
let's make sure this gets merged first before the alias table UI PR and I can make sure to resolve any merge conflicts in my branch, I think it looks good just had the one comment

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.html
ppratikcr7: Right, we can do it that way to make it more simpler. I will update the PR. :)

danoswaltCL (approved):

PR #529, opened 2022-09-07
Fix/issue 457 - ConditionAlias

RidhamShah:


danoswaltCL:
As talked about in Thurs UpGrade mtg, requesting that the GET / POST / PUT calls return consistent object type for `conditionAliases` so the UI can know for sure how get the `decisionPoint` and `parentCondition` ids from an existing `conditionAlias`. Currently if `GET` was most recent API call in the UI, the UI needs to look for `decisionPoint.id`, whereas if `POST` or `PUT` was the most recent call, it will need to look for `decisionPointId`.

Other than that, I think this is ready to go.

RidhamShah:
@danoswaltCL pushed the required changes

danoswaltCL (approved):

Issue #532, opened 2022-09-09
App-Context and Weight equally no autofilling in some experiment 'Edit' flows like refresh


danoswaltCL:
**Describe the bug**
App-Context and Weight equally no autofilling in some experiment 'Edit' flows

**To Reproduce**
1. Click on an experiment
2. Click edit
3. Hit refresh on browser
4. Click edit again
5. App-Context value no longer filled in (also, weighted-equally will not be filled in on next page)

**Expected behavior**
All of these values should re-populate

**Screenshots**
https://user-images.githubusercontent.com/97542869/189373217-8ba067ad-4aff-45f3-b493-fe8257425865.mov

PR #534, opened 2022-09-09
Add pending state for stepper, wait for contextmetadata

danoswaltCL:
This change adds some `ngrx` actions for managing pending state for contextMetaData data fetch.
It will also force the experiment stepper to finish loading contextMetaData before proceeding.
In some circumstances they may now see a load-spinner briefly.

This change was necessary to solve for certain situations where the stepper form was loading before data was finished fetching, resulting in incomplete autofill details. #532

ngrx changes implicate a lot of files, but these changes are pretty straightforward i promise.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -4,13 +4,13 @@ A/B Testing Tool
 
 Frontend Unit Test Coverage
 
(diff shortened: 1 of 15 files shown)


jreddig (approved):
oh this is slick!

PR #535, opened 2022-09-09
Refine Aliases Table UI

zackcl:
Just an update on the style of the tables so that they can look like the Figma prototype and be consistent with other tables.

I didn't work on the scrolling transition yet because it is easier to compare the two tables this way.
I will work on the scrolling transition after this is merged.

Before:
before1

After: (Updated the gap between the header and the table, button/icon position, check boxes position)
after1

Before:
before2

After: (Updated the gap between the header and the table, column header height, column widths, icon size/color, row height)
after2

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.html
@@ -1,16 +1,16 @@
 <section class="aliases-table-component">
   <div class="aliases-header-container">
-    <span class="ft-18-700">{{ 'home.new-experiment.design.aliases.header.text' | translate }}</span>
+    <span class="title">{{ 'home.new-experiment.design.aliases.header.text' | translate }}</span>
     <button
       type="button"
       mat-flat-button
       color="primary"
-      class="ft-14-600"
+      class="ft-14-600 toggle-button"
       (click)="handleHideClick()"
       [disabled]="(isAliasTableEditMode$ | async)"
     >
       <mat-icon class="icon icon-bump-left">expand_less</mat-icon>
-      <span>{{ 'home.new-experiment.design.toggle-actions.conditions' | translate }}</span>
+      <span class="toggle-text">{{ 'home.new-experiment.design.toggle-actions.conditions' | translate }}</span>
     </button>
   </div>
 
@@ -90,7 +90,7 @@
     <!-- Actions Column -->
     <ng-container matColumnDef="actions">
       <mat-header-cell *matHeaderCellDef class="ft-14-700"></mat-header-cell>
-      <mat-cell *matCellDef="let rowData; let rowIndex = index" class="actions-cell">
+      <mat-cell *matCellDef="let rowData; let rowIndex = index" class="actions-cell" style="justify-content: flex-start;">
           <button
             type="button"
             class="row-action-btn"
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.scss
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.scss
@@ -5,15 +5,86 @@
     display: flex;  
     justify-content: space-between;
 
-    .icon-bump-left {
-      margin-left: -5px;
+    .title {
+      font-size: 18px;
+      font-weight: bold;
+    }
+    
+    .toggle-button {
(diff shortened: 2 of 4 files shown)


zackcl:
@danoswaltCL Also, one more thing is that I've disabled the icon's background changing on hover for now. I think it's a good idea to highlight an icon on hover but I just wanted to keep things consistent for now. Maybe we can discuss this soon and apply this to all icons including the trash icons.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.html
```diff
@@ -90,7 +90,7 @@
     <!-- Actions Column -->
     <ng-container matColumnDef="actions">
       <mat-header-cell *matHeaderCellDef class="ft-14-700"></mat-header-cell>
-      <mat-cell *matCellDef="let rowData; let rowIndex = index" class="actions-cell">
+      <mat-cell *matCellDef="let rowData; let rowIndex = index" class="actions-cell" style="justify-content: flex-start;">
```
danoswaltCL, line 93: can these styles be added to a class in scss file instead of inline?

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.scss
```diff
@@ -113,25 +127,36 @@
     }
   }
 
-  .simple-experiment-aliases-view {
-    height: 500px;
-    border: blue 1px solid;
+  // .simple-experiment-aliases-view {
```
danoswaltCL, line 130: if it's really not needed lets just remove instead of comment out

zackcl: Yes, I guess that's the code you've used temporarily before creating a separate component right? I just wanted to make sure it's no longer needed. I will remove it and commit again.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.scss
```diff
@@ -5,27 +5,98 @@
...
+      }
     }
   }
 
   .alias-table {
-    width: 100%;
     height: 475px;
     overflow: auto;
 
+    mat-header-row {
```
danoswaltCL, line 28: these `mat-` custom elements all get classes applied with the same name, like `.mat-header-row` and `.mat-row`, etc.  I've always read that classes are the preferred way to style things, and to avoid using element selectors and/or ids when possible to avoid confusing css rule-precedence issues. If you slap a dot in from of these style rules to make them class selectors instead of element selectors, I bet it will still work the same. can we try that?

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.html
danoswaltCL: yeah I think there's a lot of that inline styling for certain things, I think it's just due to lack of styling guidelines for the project.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.scss
danoswaltCL: yep i don't see that selector used anywhere, looks like something I left in there unused

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/aliases-table/aliases-table.component.scss
zackcl: FYI, I didn't initially add the `mat-` elements in the `scss` file. They were already all over the stepper-related `scss` files. I can rename them all to `.mat-` if that's a better practice.

danoswaltCL: it's up to you if you want to change it or not, there is a lack of style guidelines so it's definitely inconsistent, there are many places where `mat-` is used and many where `.mat-` is used. i just wanted to let you know about it.

zackcl:
I will update the inlined styling and the `mat-` selectors later in separate PRs because they should be applied to all other steps too. Please feel free to merge this one unless something specific to this step should be corrected/updated.

danoswaltCL (approved):

PR #536, opened 2022-09-09
Bugfix/equal weight checkbox on edit

danoswaltCL:
#532 also, currently the "weighted equally" checkbox is always false when editing an existing experiment. This is probably because the backend isn't sending a value for this, so I added some logic in edit flow to mark it as true manually if all the conditions are actually equal.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
@@ -141,7 +141,6 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
 
     // populate values in form to update experiment if experiment data is available
     if (this.experimentInfo) {
-      console.log('info', this.experimentInfo)
       this.equalWeightFlag = this.experimentInfo.conditions.every(condition => {
         return 100 / this.experimentInfo.conditions.length === condition.assignmentWeight;
       });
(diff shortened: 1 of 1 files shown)


zackcl:
I'm not sure if this is the bug you tried to fix but updating the "Weight Equally" checkbox still doesn't seem to work properly. 
I tried it in this branch.

https://user-images.githubusercontent.com/90279765/189432431-30a91965-a8a8-4607-8375-38af32b2c158.mov

danoswaltCL:
The scenario in your video would be if the user 1) unchecked weighted equally, then 2) weighted things equally anyway. If that user went in to edit, they would see 'weighted-equally' checked. True, that's not what they technically selected... but it _is_ weighted equally, so does it matter?

(The weighted-equally checkbox exists for the UI only, no value is sent to the backend for this, so it will never "remember" this selection. This seemed like a good workaround vs updating the API for something pretty small).

zackcl:
@danoswaltCL I think you are right, it doesn't matter and in some ways, it makes more sense to enable the checkbox when the weights are equal.

zackcl:
But one problem though is that it doesn't seem to work the same way when there are 3 equally-weighted conditions. The checkbox is disabled when I open the stepper the next time which I think is a little confusing (because the "Weight Equally" checkbox is disabled while actually, they are equally weighted - unlike when there are 2 conditions).

https://user-images.githubusercontent.com/90279765/189671081-17e66864-5e74-4672-a96a-28b91c29b3f4.mov

danoswaltCL:
@zackcl bah you're right, these darn decimals. I pushed a change.

zackcl (approved):

PR #537, opened 2022-09-09
533 missing validation errors

jreddig:
Triggered validation by touching all form elements on submit. This is what the page looks like now:
Screen Shot 2022-09-09 at 4 16 09 PM

To make all fields consistent, I added an error message to the name field but I could also remove all error messages.
Screen Shot 2022-09-09 at 4 09 43 PM


danoswaltCL:
This makes sense to me code-wise, @zackcl do you think the error messages under each field are necessary? They get crunched in the space, I think for a `required` error a message might be overkill, it's pretty obvious what got missed if it turns red and has that asterisk.

zackcl:
@danoswaltCL I agree, I think it's better without the error messages under each field.

danoswaltCL:
i love when PRs remove more lines than they add

danoswaltCL (approved):

PR #540, opened 2022-09-12
Apply scrolling transition when the "Aliases" button is clicked

zackcl:
I'm trying to add the scrolling transition animation when clicking the "Aliases" button.

One problem I'm having is that when I set the buttons container (`.shared-modal--buttons-container`) to `position: fixed;` so the buttons can always show up at the bottom of the stepper regardless of scrolling, the container no longer inherits the parent container (`.shared-modal--step-container`)'s width so the buttons look like the following as if no width is applied to the buttons container.

Screen Shot 2022-09-12 at 4 04 34 PM

And if I set the buttons container's width to `100%`, then the container width exceeds the stepper modal's width.

Screen Shot 2022-09-12 at 4 20 02 PM

And if I manually set the width of the parent container (e.g. 850px) and set the width of the buttons container to `inherit`, then the width is inherited so it looks OK, but since the width is fixed to `850px`, the stepper modal is no longer responsive.

Screen Shot 2022-09-12 at 4 08 17 PM

Screen Shot 2022-09-12 at 4 09 37 PM

I would appreciate any guidance on how to properly set the width of the buttons container. Thank you!


danoswaltCL:
hmm I can take a look, maybe tomorrow... did you see that these class names come from the `shared-modal.scss` file? maybe some funky conflicting rules if you're overriding in the component `scss` file.

danoswaltCL:
I played around with this a bit and I think I got what you're looking for with `position: sticky`, https://elad.medium.com/css-position-sticky-how-it-really-works-54cd01dc2d46 was very helpful, I don't think I've ever used it but it seemed like a good use-case.

I also tried putting the rules you created in `shared-modal.scss` as classes (`scrollable` and `sticky`) that can be applied to that element. if that works for what you end up doing, we could potentially apply those styles in other modal steps simply by popping those classes on there like this:

html:
image

`shared-modal.scss`
```scss

.shared-modal--step-container {
    display: flex;
    flex-direction: column;
    justify-content: space-between;
    height: 575px;
    position: relative;

    &.scrollable {
        -ms-overflow-style: none;  /* IE and Edge */
        scrollbar-width: none;  /* Firefox */
        overflow-y: scroll;
        margin-top: -5px;
        padding-top: 5px;
    }

    &.scrollable::-webkit-scrollbar {
        display: none;
    }
}

.shared-modal--buttons-container {
    display: flex;
    justify-content: space-between;

    .shared-modal--buttons-left {
        .btn-back {
            // TODO use parent container to position buttons
            margin-left: 0 !important;
        }
    }

    &.sticky {
        position: sticky;
        bottom: 0;
        width: 100%;
        background-color: white;
    }

    // .shared-modal--buttons-right {
    //     // TODO use parent container to position buttons
    // }
}

**Comment by zackcl** [comment]
@danoswaltCL Thank you so much! I was able to fix it following your suggestion!

**Comment by zackcl** [comment]
I've included a library called [smoothscroll-polyfill](https://www.npmjs.com/package/smoothscroll-polyfill) since the smooth scrolling transition didn't work on Safari for some reason. And not just for this one, the smooth scrolling didn't work when adding new rows to a table (try adding rows on Safari before this change and compare it with Chrome). So I google-searched and found this library which seems to work fine. 

So now the scrolling transition works fine on both Chrome and Safari although the transition speed seems a little slower on Chrome (I couldn't find a simple solution to change the scrolling speed). I don't think it's that bad though.

@danoswaltCL Please try this out and let me know what you think. Also, feel free to review the code.

**Comment by zackcl** [comment]
I found out that it's possible to change the scrolling speed using jQuery. 

```
toggleAliasTable(): void {
    if (!this.isAliasTableDisplayed) {
      $( ".scrollable" ).animate({scrollTop: this.stepContainer.nativeElement.scrollHeight / 2}, 300);
    }
    else {
      $( ".scrollable" ).animate({scrollTop: 0}, 300);
    }
    this.isAliasTableDisplayed = !this.isAliasTableDisplayed;
  }
```

Would it be OK to include and use jQuery?

**Comment by danoswaltCL** [comment]
hey @zackcl I was caught up in something else, I will take a look at this tomorrow, happy you got this to work! I would really personally like to avoid dependencies unless it's truly necessary, especially jQuery, but again, I haven't looked at it closely yet.

**Comment by zackcl** [comment]
I found another library called [easy-scroll](https://www.npmjs.com/package/easy-scroll) that works fine on both Chrome and Safari and that I can set the scrolling duration and easing. So I think it might be a better option than jQuery at least.

The code basically looks like this:
```
toggleAliasTable(): void {
    if (!this.isAliasTableDisplayed) {
      easyScroll({
        'scrollableDomEle': this.stepContainer.nativeElement,
        'direction': 'bottom',
        'duration': 500,
        'easingPreset': 'easeOutCubic',
        'scrollAmount': this.stepContainer.nativeElement.scrollHeight / 2
      });
    }
    else {
      easyScroll({
        'scrollableDomEle': this.stepContainer.nativeElement,
        'direction': 'top',
        'duration': 500,
        'easingPreset': 'easeOutCubic',
        'scrollAmount': this.stepContainer.nativeElement.scrollHeight / 2
      });
    }
    this.isAliasTableDisplayed = !this.isAliasTableDisplayed;
  }
```
And here's how it looks with the code:
https://user-images.githubusercontent.com/90279765/190170365-51a7d5c0-3f79-4e27-a950-95df176fa0ff.mov

**Comment by zackcl** [comment]
If we really don't want to have a dependency on this, there's a way to do this in vanilla JS using `requestAnimationFrame`. (https://stackoverflow.com/a/24559613) 

But if the resulting functions/code will be similar to the existing open-source libraries out there, I'm not sure if have a good reason to write our own from scratch instead of using them.

**Comment by zackcl** [comment]
I'm not sure how the sliding animation works when the "Next" or "Back" button is clicked. It seems like it's provided by Angular's stepper component? (https://material.angular.io/components/stepper/overview) which means we cannot use this animation function for our scrolling transition?

I also took a look at the animations provided by Angular (https://angular.io/guide/animations) but it seems like it can only animate CSS styles. I tried animating the `top` but then the whole page including the sticky buttons moved together. I don't think it's possible to completely mimic scrolling in CSS.

**Comment by zackcl** [comment]
OK, I was able to write a function that works just like [the video I posted](https://github.com/CarnegieLearningWeb/UpGrade/pull/540#issuecomment-1246788550) in vanilla JS based on the stackoverflow post ((https://stackoverflow.com/a/24559613).

I will clean up the code and share it soon.

**Comment by danoswaltCL** [comment]
hey Zack, yeah this looks great in the video, code makes sense to me; in my opinion, a small vanilla JS pollyfill is probably preferable to pulling in an extra dependency, so if you have it working like that, that's great.

I have found working with Angular animations by hand to be really unintuitive, but that's my lack of experience, there's probably a way to manage scrolling types of things that way also. But your approach is straightforward with a small, well-loved custom polyfill for cross-browser support, I like that the best, you're right, it's probably not much different than a small library but then again, [small libraries have caused big problems before](https://qz.com/646467/how-one-programmer-broke-the-internet-by-deleting-a-tiny-piece-of-code/), it's a slippery slope!

fyi, If you do end up finding that a dependency you mentioned is still better for whatever reason (now or for some future thing), we want to pull it in via `npm install <packagename>` to properly get the library registered Angular project (vs a script tag on index.html, that should only be for special cases). That way package updates will get managed by npm and all that nice stuff.

I'm always looking to keep projects as lean as they can be, because keeping the project up to date is always complicated by third-party code out of our control, but obviously sometimes libraries are awesome. The advice I've always been given on pulling in third-party code, so you know where I'm coming from: make sure any library we use is 1) popular on the npm repository, and 2) updated frequently, and 3) has no security vulnerability warnings. And also, 4) not very big and 5) absolutely necessary. If it's tiny and well-supported and saves us a lot of headache, then go for it.

**Comment by danoswaltCL** [comment]
also, in case you haven't seen this, angular out-of-the-box generates a `polyfills.ts` file if you're looking for a home for your polyfill snippets.

File: frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts [hunk 1]
```diff
@@ -617,6 +618,18 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
+        behavior: 'smooth'
+      });
+    }
+    else {
+      this.stepContainer.nativeElement.scroll({
+        top: 0,
+        behavior: 'smooth'
+      });
+    }
     this.isAliasTableDisplayed = !this.isAliasTableDisplayed;
```
danoswaltCL, line 633: `isAliasTableDisplyed` flag was probably only used for the `[hidden]` attribute in the html that you change removes, I think we can safely remove this line and it's declaration. would you be able to check and remove if so?

zackcl: `isAliasTableDisplayed` is still being used to toggle between the two pages, but I can also remove it and create 2 separate functions (e.g. `scrollToAliasesTable()` and `scrollToConditionsTable()`) instead of `toggleAliasTable()`. I can do it later.

danoswaltCL: oh okay nah don't worry about that, I thought it was no longer used, nevermind!

zackcl:
@danoswaltCL In the new commit, I'm still using the library (polyfill) that I mentioned before ([smoothscroll-polyfill](https://www.npmjs.com/package/smoothscroll-polyfill)) which seems to be the most popular one. But I've added options to set duration and easing function to a scrolling animation on top of it to serve our purpose.

So for example, I can now call the `scroll()` function this way:
```
this.stepContainer.nativeElement.scroll({
      top: 0,
      behavior: 'smooth',
      duration: 500,
      easing: 'easeOutCubic'
    });
```
The `duration` and `easing` can be passed optionally and without them, it will work just like the native `scroll()` function.

Also, I got rid of the functions that are currently unused by UpGrade (e.g. `scrollBy()`, `scrollIntoView()`) as I think we can add them anytime later when needed.

We do use the `nativeElement.scroll()` function in many places in UpGrade so this polyfill implementation will override the native function, and it will work identically on both Chrome and Safari.

Now I'll probably work on moving the code to the `polyfills.ts` file as you suggested. Please let me know if you have any questions or concerns.

danoswaltCL:
ok, try it there or in a new standalone angular service class (i can show what I mean by that if you're not sure how to implement that).

I think if it is added via assets folder outside of the angular source code it could potentially cause some issues, it's a little against the grain. if you run into typescript complaining, i might be able to help.

zackcl:
@danoswaltCL I've converted the previous JS code to TS and placed it in the `polyfills.ts` file. The code works fine but I'm not sure if this is the proper way to add polyfills in Angular. 

Do you think it's better to create an Angular service class for this, even if the code needs to run only once in the beginning? (maybe the class will be imported once in the `polyfills.ts` file?)

danoswaltCL:
I don't see any advantage of a service vs polyfills.ts really, it feels like the reason for it's this file's existence. If you are wanting to get that code out of there and modularized, you should be able to put that code in a ts file, export it, and import into polyfills. maybe that's cleaner.

I've not tried to find a home for custom polyfills in an angular project myself, so if you have found a better approach go for it.

zackcl:
Thank you! I agree that it would be cleaner to put the code in a separate ts file and have it imported into `polyfills.ts`.

So do you think there's no need to create a service? Should I just create a file called `smoothscroll.ts` then? (instead of `smoothscroll.service.ts`) 

And where do you think this file should be located? The same location where the `polyfills.ts` is in?

danoswaltCL:
yep, same directory, suppose could even have a new folder in the directory called polyfills and put it there, if we decided later to include some other custom polyfill scripts.

danoswaltCL:
> So do you think there's no need to create a service? Should I just create a file called smoothscroll.ts then? (instead of smoothscroll.service.ts)

Yup. I think the reason would be that services should be for app business logic, and this is strictly for overriding browser DOM, so really the app sort of doesn't even really 'know' this exists, it just reaches out to the browser to ask it to do scrolling, and the window takes care of that part itself. that's how i'm separating it in my head at least. and if I was looking in the project for polyfills, the first place i'd look would be in this file.

zackcl:
I've created a separate `polyfills/smoothscroll.ts` folder/file and had it imported in `polyfills.ts`.

Please feel free to try the app and review the code.

danoswaltCL:
I think this works, nice! I checked also to make sure that doing it this way with polyfills didn't put a bunch of extra variables on the global window scope, but it does not, I think typescript treats code in a file as if it is in local scope when it compiles it.

Anyway, nice work!

danoswaltCL (approved):

Issue #505, opened 2022-08-11
UI enhancements and bug fixes for the Experiment Stepper


zackcl:
## All Steps
- [x] Correct labels/placeholders ([#470]( https://github.com/CarnegieLearningWeb/UpGrade/pull/470))
- [x] Set equal font size and style for checkbox labels ([#482]( https://github.com/CarnegieLearningWeb/UpGrade/pull/482))
- [x] Remove default rows ([#498]( https://github.com/CarnegieLearningWeb/UpGrade/pull/498))
- [x] Disable autocomplete suggestions ([#501]( https://github.com/CarnegieLearningWeb/UpGrade/pull/501))

## Overview Step
- [x] Set the equal vertical spacing between the fields ([#506]( https://github.com/CarnegieLearningWeb/UpGrade/pull/506))

## Design Step
- [x] Change the title font size to 18 ([#486]( https://github.com/CarnegieLearningWeb/UpGrade/pull/486))
- [x] Correct the vertical spacing between the title and the table ([#486]( https://github.com/CarnegieLearningWeb/UpGrade/pull/486))
- [x] Correct the table column widths ([#484]( https://github.com/CarnegieLearningWeb/UpGrade/pull/484))
- [x] Right-align the weight field's text ([#484]( https://github.com/CarnegieLearningWeb/UpGrade/pull/484))
- [x] Change the weight field's input type to `number` ([#499]( https://github.com/CarnegieLearningWeb/UpGrade/pull/499))
- [x] Correct the `scrollHeight` so the entire row is visible when added ([#486]( https://github.com/CarnegieLearningWeb/UpGrade/pull/486))
- [x] Make validation error messages more vertically compact ([#486]( https://github.com/CarnegieLearningWeb/UpGrade/pull/486))
- [x] Make the `Target` field required ([#499]( https://github.com/CarnegieLearningWeb/UpGrade/pull/499))

## Participants Step
- [x] Change the title font size to 18 ([#493]( https://github.com/CarnegieLearningWeb/UpGrade/pull/493))
- [x] Correct the vertical spacing between the title and the table ([#493]( https://github.com/CarnegieLearningWeb/UpGrade/pull/493))
- [x] Correct the table column widths ([#493]( https://github.com/CarnegieLearningWeb/UpGrade/pull/493))
- [x] Remove underline from dropdowns ([#493]( https://github.com/CarnegieLearningWeb/UpGrade/pull/493))
[...]

PR #541, opened 2022-09-13
Front-end changes left for issue #505

Yagnik56:
- Show a validation message if there's any unfilled row
- Vertically align the rows within a cell to the top for any metrics selected


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -44,7 +44,9 @@ export class ExperimentParticipantsComponent implements OnInit {
   segmentNameId = new Map();
   subSegmentTypes: any[];
   subSegmentIdsToSend = [];
-  membersCountError: string = null;
+  IncludeEmptyRowError = false;
+  ExcludeEmptyRowError = false;
+  EmptyRowErrorMsg = "Please fill or remove empty row fields before moving forward"
```
ppratikcr7, line 49: We have validation message's in en.json file. Please use that for this message.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -197,29 +199,30 @@ export class ExperimentParticipantsComponent implements OnInit {
...
-    const excludedMembersFiltered = this.removeEmptyRows(excludedMembers);
+    const includedMembersFiltered = this.returnEmptyRowsCount(includedMembers);
+    const excludedMembersFiltered = this.returnEmptyRowsCount(excludedMembers);
 
-    if (includedMembersFiltered.length === 0) {
-      this.members1.clear();
-      this.updateView1();
+    if (includedMembersFiltered.length > 0) {
+      this.IncludeEmptyRowError = true;
+    }else{
```
ppratikcr7, line 207: spaces are required around `else`:

} else {

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -197,29 +199,30 @@ export class ExperimentParticipantsComponent implements OnInit {
...
+      this.IncludeEmptyRowError = false;
     }
 
-    if (excludedMembersFiltered.length === 0) {
-      this.members2.clear();
-      this.updateView2();
+    if (excludedMembersFiltered.length > 0) {
+      this.ExcludeEmptyRowError = true;
+    }else{
```
ppratikcr7, line 213: spaces are required around `else`:

} else {

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -197,29 +199,30 @@ export class ExperimentParticipantsComponent implements OnInit {
...
       return; // form will be invalid
     }
 
     return members.filter((memberRow) => {
-      // only return false if both type and id are falsey, which indicates empty/unneeded row
-      // otherwise, if one is false and other is truthy, don't remove row, handle as invalid form
-      return !(!memberRow.type && !memberRow.id)
+      // only return if any of type and id are falsey, which indicates empty/unneeded row
+      return (!memberRow.type || !memberRow.id)
     })
```
ppratikcr7, line 226: add semicolon after each statement:  return xyz;
and });

ppratikcr7 (changes requested):
Added few comments, please make those changes

danoswaltCL:
Wait let's talk about this one....`Show a validation message if there's any unfilled row` @zackcl what is the any design for what this change is supposed to be and what the message should look like? I don't think this change is necessary as far as I understand it, we already get a message if the row is incomplete (see screenshot), although it's maybe not the most clear message for every scenario.

Like our convo in @jreddig 's PR yesterday, a nicer way to handle this may be to make sure if there are empty required fields when clicking 'NEXT', that the form fields that are unfilled get their natural form validation triggered with `markAsTouched`. (aka, no message, just let angular turn it red to point out the unfilled parts).

(Side note, the function that @Yagnik56 has modified should actually probably not exist anymore, it was a function i put in there to automatically remove unused rows for some user story that is no longer relevant. I would not recommend using or modifying it to do something fundamentally different like what these changes are doing).

image

zackcl:
@danoswaltCL The screenshot you attached is the Design step and I believe this PR is for showing the validation message in the Participants step. 

And yes, I think we can consider turning the unfilled parts to red without showing the message. But I'm thinking, what about other validation messages? Should we still show messages about duplicated names and weights not summing up to 100?
182390749-e0fcec30-781d-4be4-be18-07c51afdeacb

I'm not sure if we have a good reason to not show a validation message for the unfilled rows only.

jreddig:
I think those error messages are important because it's not just that the user left part of the form blank, they filled it in incorrectly. Just the red outline makes sense for an empty required field, but we need to clarify when the values in the field are wrong. I think that in both cases, the field should turn red, but a message is not necessary for a required error.

zackcl:
@jreddig I agree. 

Hi @Yagnik56, if you don't mind, could you make unfilled fields turn red without showing a validation message in the Participants step? Maybe you can take a look at this PR (https://github.com/CarnegieLearningWeb/UpGrade/pull/537) as a reference.

danoswaltCL:
oh my god, I knew I was missing something. ok here's what it looks like:
image

I'm in favor of not doing this particular validation message and letting angular handle the required field error.

I want to make all aware that currently, if a row is completely empty, it will just get removed. That was functionality I added back when we showed 1 default empty row, because otherwise they would need to delete the unfilled default row in order to proceed. That was pretty annoying since it isn't actually required to have anything, so this was added to remedy that.

However we no longer do the default row, I'm in favor of getting rid of the function `removeEmptyRows` that this change was trying to re-purpose, it's no longer a relevant thing to be doing, let's kill it. It would be better for the user to get a required field error and delete the row themselves if they don't want it.

zackcl:
@danoswaltCL I agree!

danoswaltCL:
@Yagnik56 sorry, that was a lot of discussion we probably should have clarified before having you start the task! your changes totally make sense, but now I think we have said

1) no message needed
2) have the required fields turn red like in the PR @zackcl referenced (potentially via `markAsTouched` or `markAllAsTouched` in order to force angular to validate the required fields and apply that red state automatically)
3) remove the `removeEmptyRows` (that function you changed to `returnEmptyRowsCount`) and probably the `checkForEmptyRows` logic altogether, we don't need to be checking or removing empty rows anymore.

Yagnik56:
@danoswaltCL I didn't got last point, can u clarify as remove removeEmptyRows logic was already there so should I make any changes or let it stay as it is. About 2 required field are turning red itself so I think that is also sorted too.

Yagnik56:
I have removed both functions `returnEmptyRowsCount` & `checkForEmptyRows` and added `markAllAsTouched` function for participants form 1 and form 2 below `case NewExperimentDialogEvents.SAVE_DATA:`.

danoswaltCL (approved):

ppratikcr7 (approved):

PR #544, opened 2022-09-16
Fix metrics rows margin

zackcl:
Just a small fix that makes the row height the same as the other table rows (more compact), and also aligns the text vertically (the display name field and the remove icon were a little off).

Before:
before

After:
after


jreddig (approved):
Consistent with the other stepper tables

Issue #548, opened 2022-09-26
Add edit mode to the tables in the experiment stepper


zackcl:
## Problems with the current table
* It's easy for the user to change something in a table by mistake
* It's harder to validate the table values since empty/unfilled rows can exist

## Solutions
* Add edit mode so that a table can only be edited on edit mode
* A row can be validated/updated as the user clicks the "Save" icon/button
* A row editing can be canceled if the user clicks the "Cancel" icon/button
* While on edit mode, the user won't be able to edit other rows or move to the next/previous step

## Reference
[Material Table Editable Examples](https://material-table.com/#/docs/features/editable) (scroll to the bottom of the page)

## Figma Prototype
[The latest UI Prototype](https://www.figma.com/file/uYxnCT2dXTftSZKFQ00Q0y/UpGrade-UI-Prototype-(latest))

## Screenshots
<img width="500" alt="Experiments - Add Experiment-02-01" src="https://user-images.githubusercontent.com/90279765/192312049-41c8e34d-bd8e-45d5-9c74-c616cebce57d.png">
<img width="500" alt="Experiments - Add Experiment-02-02" src="https://user-images.githubusercontent.com/90279765/192312064-b083a134-316f-407e-b078-7f7c263f4022.png">
<img width="500" alt="Experiments - Add Experiment-02-03" src="https://user-images.githubusercontent.com/90279765/192312076-e8274bb5-2fb4-47d6-8430-57c39326aa87.png">
<img width="500" alt="Experiments - Add Experiment-02-04" src="https://user-images.githubusercontent.com/90279765/192312199-ced55500-7ee5-41a5-9636-30c2bd680514.png">
<img width="500" alt="Experiments - Add Experiment-02-05" src="https://user-images.githubusercontent.com/90279765/192312239-3b84f54e-d533-4bbf-b34a-19dcea28fe66.png">
<img width="500" alt="Experiments - Add Experiment-02-06" src="https://user-images.githubusercontent.com/90279765/192312258-330d19b2-7ff2-4ba3-9814-bf49e1ae89bc.png">
<img width="500" alt="Experiments - Add Experiment-02-07" src="https://user-images.githubusercontent.com/90279765/192312269-61d50b8f-9c7f-4095-8ee6-ece08cf75963.png">

## Todos
- [ ] Add edit mode to the decision points and conditions tables in the Design step
- [ ] Add edit mode to the include and exclude tables in the Participants step
[...]

PR #549, opened 2022-09-26
Add edit mode to the Conditions table

zackcl:
This is just a draft. I tried to add edit mode to the Conditions table (the description field to begin with) based on the code @danoswaltCL wrote for the Aliases table. But it seems more complicated than I thought as it involves more Angular stuff that I don't understand yet (I just started taking the Angular course).

For example, I don't know how to properly create a `conditionsTableData` for row editing on `ngOnInit()` so the `rowData.description` in the `experiment-design.component.html` can be read from the beginning. It seems like the `mat-table` should be updated/refreshed through the `dataSource` object which is not the same type used in the Aliases table.

Anyways, I think it would be great if @danoswaltCL or someone can implement the edit mode for at least the description field in the Conditions table so I can learn from it and apply that pattern to the rest of the fields. Please feel free to create another branch and start over. Thank you!

By the way, I posted an issue about adding edit mode to the tables in the experiment stepper: https://github.com/CarnegieLearningWeb/UpGrade/issues/548

It currently looks like this (styles are fine I think) but doesn't work properly:
Screen Shot 2022-09-26 at 2 40 45 PM


danoswaltCL:
hey @zackcl I will take a look at this tomorrow if I get a chance to. I know you're out for a while starting wednesday, so I will make sure to give some notes and potentially implement the rest of the way if it's already pretty close and share with you some comments on some of the Angular-specific stuff for when you get back.

danoswaltCL:
hey @zackcl finally got around to pulling this down and taking a look. Yes it probably can't work exactly the same as alias-table, although it kind of looks like it's close to working the way you've got it so far. let me remind myself of how this worked and dig into this more closely in the code tomorrow.

danoswaltCL:
hey @zackcl i am actually getting to this now, I've got a free moment to do so. I'll let you know how it goes! I will try to implement the rest of the way and I'll share what I did.

PR #550, opened 2022-09-27
Update names in the experiment details page

zackcl:
Changes:
* Changed the "Exclude If Reached" column header font size
* Capitalized the "Condition" column header
* "Enrollment by Decision Point" instead of "Enrollments Over Time by Sites" 
* "Enrollment by Condition" instead of "Enrollments by Condition"
*  Change "Post Rule" to uppercase

Before:
before

After:
after


zackcl:
Please feel free to merge this.

danoswaltCL (approved):

PR #557, opened 2022-09-28
Feature/regression scripts

danoswaltCL:
Initial draft of potential regression test suite. Just vanilla typescript node scripts.

These tests are meant to be run locally against an environment in a sort of 'black-box' mode, calling APIs in a given scenario and logging results against predefined assertions.

Currently this is focused on testing as many scenarios as we can think of for identifying how "consistency rules", "excludeIfReached", and "assignment unit" are currently working, but it should be able to be extended.

- To execute tests, simply do `npm start`.

- To choose the tests you want to run, comment out/in the tests names in `index.ts`.

- To add a spec scenario, you can just copy/paste an existing template object in `ExcludeIfReachedScenarios.ts`. Give it a unique ID and add it to the list of testName constants. Add that name to the `.run` list in `index.ts`.

- There are params here for writing results to file, and environment settings for running against any environment.

Issues to worked out, if others have ideas:
- Currently synchronously completing each spec scenario one at a time, which is slow, but I was getting db locking up trying to do more than 5+ tests in async manner.

- Needs more complete teardown logic.

- Perhaps better logging of results. I'm thinking less json and more colors....


jreddig (approved):
Super excited about this!

PR #558, opened 2022-09-30
Fix for issue 31: Allowing shared decision points for Competing Experiments (without pools)

ppratikcr7:
This PR also contains client side SDK changes. The changes include renaming `experimentPoint`  to `site`  and `partitionId` to `target` . Also, we internally pass an optional parameter `experimentId` to mark API call (not required in the client SDK call, this is an internal parameter to the API), required only in the case of shared decisionPoint. As discussed in the meeting, this PR merge would require a clean DB.


danoswaltCL (changes requested):
Creating experiment breaks from UI if alias conditions are supplied, due to the change in decisionPoint id from site_target to uuid. I will also take a look as I wrote the code for this, there are a couple spots where this might be affected.

danoswaltCL:
@ppratikcr7 @VivekFitkariwala I was able to get the above issue working doing the following changes:

In `experiment-design.component.ts`, on line `519` we need to create a `decisionPoint.id` with `uuid4()` in the UI for new experiments so that we can reference the id when sending in the `conditionAlias` on line `572`:

image

image

Then we also need a change in `alias-table-component` to reference pre-existing decisionPoints in the edit flow... here apparently the `id` won't be so easily accessible the way things are written, but it works just as well to modify this block on line 107 a little to match the site and target separately instead of matching site_target from the old id format.

Change from this:

image

to this:

image

This looked okay to me, but I have not tested it very thoroughly. I am able to add, edit, and re-edit after reloading the app, so it may work, please see if it works for you and if it's okay to generate the UUID for decisionPoint id on UI side in request.

danoswaltCL:
hey @ppratikcr7 the schedule for getting changes into the client library won't work for us to be able to get an update in in time in Mathia to be able to run XPrize experiment; can we make a small change to this PR also to allow both the new and old field names in the (experimentPoint and partitionId) for now so that we don't need to have client interface changes at this time?

VivekFitkariwala (approved):

danoswaltCL:
closing as it's no longer needed

PR #561, opened 2022-10-06
Fix/issue 31: Allowing shared decision points for Competing Experiments (With Pools)

ppratikcr7:
This PR contains the below changes:

1. We now have support for shared decision points in competing experiments
2. User will be assigned an experiment condition from each pool of experiments.
3. A user will get the same assigned experiment from a pool each time in case of a shared decision point as it would be randomly picked using seed random of the user id.
4. Model changes for DecisionPoint. We now use uuid as primary key.
5.  Frontend: Removing validation checks for allowing same decision points in multiple experiment.
6. Added the type column for experiment. (Added a column `type` to allow simple or factorial experiment type) This would be helpful as we won't need to clean db again for this.


ppratikcr7:
@danoswaltCL As per your [comment](https://github.com/CarnegieLearningWeb/UpGrade/pull/558#issuecomment-1267218431) on the previous PR for competing experiment(without pools), we can't have multiple params. Instead, we feel that we will revert those param names and let it continue with the old naming convention. The competing experiment pool's logic should still work and later when you are ready to update the SDK at your end, we can again provide the new naming conventions. :)

ppratikcr7:
@danoswaltCL We are working on a small fix in mark call.

ppratikcr7:
> @danoswaltCL We are working on a small fix in mark call.

This fix has been pushed. Only need to revert the SDK changes

ppratikcr7:
The test suite was resolved and all test cases pass in my local machine. Still the backend workflow shows a failing test case somehow. The same test case passed for me. Can someone check if the test cases fails for them in their machine?
Screenshot 2022-10-08 at 12 33 31 AM
Screenshot 2022-10-08 at 12 35 25 AM
Screenshot 2022-10-08 at 12 39 53 AM

danoswaltCL:
yes, i just saw this and was about to check, I will look

danoswaltCL:
i'm having issues running the tests locally, I keep getting async timeout errors on the last test suite that always takes a while. I doubt that's related, but I'll try again... I can't tell if it's an issue with the test suite or the actual code.

I re-ran the tests here it github runner, sometimes there are those async timeout kinds of issues, or maybe the test-suite has some inconsistent behavior in the newer tests, but I got same result.

danoswaltCL:
No actually I do get the same error locally now: 
image

danoswaltCL:
@ppratikcr7 how can I tell which scenario this test is erroring in? this function with the failed assertion is used by most of the scenario files. Is it failing when running all tests, but succeeding when testing it in isolation?

danoswaltCL:
ok, now I've seen it pass in my local and fail in my local without any changes. I would guess there may be an inconsistency issue with the test suite itself, with some kind of async issue or tests not completely isolated.

ppratikcr7:
> ok, now I've seen it pass in my local and fail in my local without any changes. I would guess there may be an inconsistency issue with the test suite itself, with some kind of async issue or tests not completely isolated.

Yeah same happened for me as well. I tried restarting my long running system and the 5 sec aysnc timeout issue was gone in my local. The logs generated are also too much I feel, 75k lines which even github workflow is unable to load properly, maybe we cut down on a few long logs. Like we have introduced pool's console logs for testing and demo purpose.

jreddig (changes requested):
The Java clientlibs are expecting the marked object response to have `createdAt`, `updatedAt`, `versionNumber`, `userId`, `experimentId`, `enrollmentCode`, `condition`, `id`, `status`, and `decisionPoint` and with the term changes, `markExperimentPoint` is throwing errors. Can we adjust the object returned from mark to have only these values so we dont have to update MathIA? Remove `experimentId`, change `site` to `experimentId`, and `target` to `decisionPoint`?

I have a hack-y way in my local that is compatible with the client libs, but maybe there's a more elegant way we can do this? At least until we can get coordinate with mathia engineering to make an update

```
o['experimentId'] = o['site']
o['decisionPoint'] = o['target']
delete o['site']
delete o['target']
```

VivekFitkariwala (dismissed):

ppratikcr7:
> The Java clientlibs are expecting the marked object response to have `createdAt`, `updatedAt`, `versionNumber`, `userId`, `experimentId`, `enrollmentCode`, `condition`, `id`, `status`, and `decisionPoint` and with the term changes, `markExperimentPoint` is throwing errors. Can we adjust the object returned from mark to have only these values so we dont have to update MathIA? Remove `experimentId`, change `site` to `experimentId`, and `target` to `decisionPoint`?
> 
> I have a hack-y way in my local that is compatible with the client libs, but maybe there's a more elegant way we can do this? At least until we can get coordinate with mathia engineering to make an update
> 
> ```
> o['experimentId'] = o['site']
> o['decisionPoint'] = o['target']
> delete o['site']
> delete o['target']
> ```

Hey, yeah updated the `mark` and `assign` api call response as per MathiA. Removed `experimentId` from `assign` call response as well.

ppratikcr7:
> ok, now I've seen it pass in my local and fail in my local without any changes. I would guess there may be an inconsistency issue with the test suite itself, with some kind of async issue or tests not completely isolated.

I have resolved the test case issue.

danoswaltCL (approved):

jreddig (approved):

PR #566, opened 2022-10-10
remove stylelint

danoswaltCL:


jreddig (approved):
always for removing unused packages

PR #573, opened 2022-10-19
fix for users not excluded reached prior to experiment state changed to enrolling

ppratikcr7:


zackcl:
As I tested using the testing app (https://upgrade-tester.glitch.me/), everything seems to work as expected. So I think it's good to merge this.

jreddig (approved):
Looks good! Thanks for thoroughly testing, Zack!

Issue #562, opened 2022-10-07
Experiment Design UI bug for Dark mode


RidhamShah:
The white background color is used which creates a visual bug in dark mode
![image](https://user-images.githubusercontent.com/49234788/194487992-c806319d-2e11-4dde-b3cd-853a4b6b8daf.png)

Also overlapping of Alias form button is seen
![image](https://user-images.githubusercontent.com/49234788/194489513-bb46856c-cdd3-43bf-a674-89239763c9c4.png)

PR #578, opened 2022-10-25
Fix UI bug for dark mode

zackcl:
This one fixes the design step's dark mode issue: https://github.com/CarnegieLearningWeb/UpGrade/issues/562.

It seems to work fine as I tested but please check if I did it correctly.


frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/new-experiment/new-experiment.theme.scss
```diff
@@ -7,5 +7,11 @@
     .mat-horizontal-stepper-header-container {
       box-shadow: 0 8px 8px -8px mat.get-color-from-palette($background, shadow-color);
     }
+    
+    .shared-modal--buttons-container {
```
danoswaltCL, line 11: cool, I think this is the right way to do it

danoswaltCL (approved):

PR #581, opened 2022-10-26
remove node_modules from version control

jerith666:
this directory is in .gitignore, and AIUI it should never be included in version control, but instead always produced in a local working copy by running 'npm ci' or similar.

this was added in aef0b8bf32ac7c1312ee7478da9ebc223b7602db; the commit message is pretty vague but my read of it is that it must've been accidental.


danoswaltCL:
ty, I guess it's in there forever in the git history, but no this shouldn't have been in there

danoswaltCL (approved):

PR #584, opened 2022-10-27
Fix/issue-575: Handling no experimentId in mark api call

ppratikcr7:


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -267,7 +267,7 @@ export class ExperimentAssignmentService {
         },
       });
 
-      if (experimentDecisionPoint.length) {
+      if (experimentDecisionPoint.length && experimentId !== '') {
         let selectedExperimentDP = experimentDecisionPoint.filter( dp => dp.experiment.id === experimentId);
```
zackcl, line 271: Just wondering if it's a better solution than `let selectedExperimentDP = experimentDecisionPoint.filter( dp => dp.experiment.id === experimentId || experimentId === '');` which was what I did to not see the error.

zackcl: Got it, thank you!

ppratikcr7:
> If this is really all that's needed, then the tertiary condition to set the empty string really isn't necessary, we can simplify a little more:
> 
> ```ts
> if (!experimentId) {
>    if (filteredExperiments.length > 1) {
>       const random = seedrandom(userId)();
>       experimentId = filteredExperiments[Math.floor(random * experiments.length)].id;
>    } else {
>       experimentId = filteredExperiments[0]?.id; // <--- no need for extra conditional here, let it be undefined, no need for empty string
>    }
> }
> ```
> 
> and then you can do this so that it's not just an empty string it's checking, it's checking all falsey values that would potentially get the same runtime error:
> 
> `if (experimentDecisionPoint.length && experimentId) {`
> 
> which also catches any weird edge-case where `experimentId` is somehow falsey for any other reason (such as if somehow `experimentId = filteredExperiments[Math.floor(random * experiments.length)].id;` returns `undefined` or `NaN`, which could technically happen I think if `experiments.length` is undefined or an unexpected number, or if that `id` is messed up. These aren't important scenarios, but you never know.

Okay, I will update the PR. Thanks :)

danoswaltCL (approved):

PR #592, opened 2022-10-28
speed-bump when click on close while edit or create experiment

Yagnik56:
I have added speed bump if they are change made in any tabs in creating new experiment or edit experiment dialog box.
Here is image,
image

I have taken care tracking changed between tabs, reset on click on update button, etc...


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -32,6 +33,8 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
   @Input() currentContext: string;
   @Input() isContextChanged: boolean;
   @Input() animationCompleteStepperIndex: number;
+  @Input() dataChanged: false;
```
danoswaltCL, line 36: you're setting `false` as the type, I think you want `@Input() dataChanged: boolean = false;`

danoswaltCL: I see this in several files, let's change that for all of them

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -301,6 +306,7 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
       this.previousAssignmentWeightValues.splice(groupIndex, 1);  
       this.applyEqualWeight();
     }
+    this.isRowRemoved = true;
```
danoswaltCL, line 309: what is this new variable for (how does it relate to speedbump?).

If it is necessary to have this, it appears to be initialized as false and basically immediately set to true, perhaps it can just be initialized as true.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -486,58 +500,75 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
...
         }
+        this.saveData(eventType);
+        this.dataChanged=false;
+        this.isRowRemoved = false;
+        this.experimentDesignForm.markAsPristine();
         break;
     }
   }
 
+  saveData(eventType){
```
danoswaltCL, line 522: 💯 for refactoring and breaking this out into it its own function, always happy to see these long code blocks broken down into smaller functions

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
Yagnik56: I have used ng-form.dirty to detect changes in form but it wasn't detecting deletion of rows in tables so I have created this variable and set is as false but if any row gets deleted then it will be true and added that variable in if condition with ng-form.dirty. 
image

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
Yagnik56: resolved this

danoswaltCL:
I've been thinking about this. What you have will work, but managing updates about data changes in this way in Angular is going to be painful in the long-run, because it's a lot of code to keep setting and unsetting flags and many components to pass data directly to each other. We have `ngrx` for this sort of thing.

Since this isn't a critical update, I have a request on this. Have you looked at `ngrx` and how it works in our app? This is a pretty perfect type of feature enhancement to learn how to use `ngrx` if not. Can you potentially try using `ngrx` actions and the `ngrx` data store to manage a boolean that indicates "dirty" or "clean" form state? We should be able to manage this feature without using `@Input` and `@Output`.

danoswaltCL (dismissed):
Please try with `ngrx` data-store patterns.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-schedule/experiment-schedule.component.ts
```diff
@@ -134,75 +140,91 @@ export class ExperimentScheduleComponent implements OnInit {
...
-          });
         }
-        break;
+      }
+      this.emitExperimentDialogEvent.emit({
+        type: eventType,
+        formData: scheduleData,
+        path: NewExperimentPaths.EXPERIMENT_SCHEDULE
+      });
     }
```
danoswaltCL, line 228: i think we're missing a `break` statement

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -501,72 +507,89 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
...
+      }
+    );
+
+    if (this.queryMetricKeyError.length === 0 && this.queryStatisticError.length === 0 && this.queryComparisonStatisticError.length === 0 && this.queryNameError.length === 0) {
+      this.emitExperimentDialogEvent.emit({
+        type: eventType,
+        formData: monitoredMetricsFormData,
+        path: NewExperimentPaths.MONITORED_METRIC
+      });
     }
```
danoswaltCL, line 593: looks like a `break` statement is missing

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/new-experiment/new-experiment.component.ts
```diff
@@ -69,6 +76,10 @@ export class NewExperimentComponent implements OnInit {
           this.experimentService.createNewExperiment(this.newExperimentData);
           this.onNoClick();
         }
+
+        // console.log(this.experimentInfo);
```
danoswaltCL, line 80: please remove comments

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/new-experiment/new-experiment.component.ts
```diff
@@ -84,6 +95,15 @@ export class NewExperimentComponent implements OnInit {
     }
   }
 
+  // isSubset = (superObj, subObj) => {
```
danoswaltCL, line 98: not sure what this is or was, can it be removed?

frontend/projects/upgrade/src/app/shared/components/mat-confirm-dialog/mat-confirm-dialog.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { MatConfirmDialogComponent } from './mat-confirm-dialog.component';
+
+describe('MatConfirmDialogComponent', () => {
```
danoswaltCL, line 5: if we aren't going to write tests, we should either delete this file or mark it `xdescribe` for now

frontend/projects/upgrade/src/app/shared/services/dialog.service.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { DialogService } from './dialog.service';
+
+describe('DialogService', () => {
```
danoswaltCL, line 5: if we aren't going to write tests, we should either delete this file or mark it `xdescribe` for now

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
Yagnik56: saveData() function is called in switch case and there are break statement after function called there after flag reset.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-schedule/experiment-schedule.component.ts
Yagnik56: saveData() function is called in switch case and there are break statement after function called there after flag reset.

danoswaltCL: ok, that could be confusing hidden in a function, the `break` statement should be with the `case` code block to keep that syntax clean and readable

danoswaltCL: > I've been thinking about this. What you have will work, but managing updates about data changes in this way in Angular is going to be painful in the long-run, because it's a lot of code to keep setting and unsetting flags and many components to pass data directly to each other. We have `ngrx` for this sort of thing.
> 
> Since this isn't a critical update, I have a request on this. Have you looked at `ngrx` and how it works in our app? This is a pretty perfect type of feature enhancement to learn how to use `ngrx` if not. Can you potentially try using `ngrx` actions and the `ngrx` data store to manage a boolean that indicates "dirty" or "clean" form state? We should be able to manage this feature without using `@Input` and `@Output`.

Let me know if this request makes sense, I'd be happy to show what I mean by this.

danoswaltCL:
> I've been thinking about this. What you have will work, but managing updates about data changes in this way in Angular is going to be painful in the long-run, because it's a lot of code to keep setting and unsetting flags and many components to pass data directly to each other. We have `ngrx` for this sort of thing.
> 
> Since this isn't a critical update, I have a request on this. Have you looked at `ngrx` and how it works in our app? This is a pretty perfect type of feature enhancement to learn how to use `ngrx` if not. Can you potentially try using `ngrx` actions and the `ngrx` data store to manage a boolean that indicates "dirty" or "clean" form state? We should be able to manage this feature without using `@Input` and `@Output`.

@Yagnik56 Let me know if this request makes sense, I'd be happy to do a call to go over what I mean.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-schedule/experiment-schedule.component.ts
Yagnik56: image
They are in `case` code block, I just put code below `SAVE_DATA` in function and called it for both `SEND_FORM_DATA` and `SAVE_DATA` cases.

danoswaltCL: ok apologies, reading it in github-diff made me see it weird then

Yagnik56:
I understood your request, I have looked into few documentation so I will try it out tomorrow. We can do a call if it won't work out as I haven't used `ngrx` before so...

danoswaltCL:
Cool, I hadn't used `ngrx` before this project either, it's a little tedious but it will make things easier in long run

Yagnik56:
> @Yagnik56 I will resolve these conflicts on what is in github today. If you have work that has not been pushed up, let me know and we can work through those merge conflicts as well if need be.

No I don't have any chances to push, I was trying to resolve that row deletion issue without using the flag but ng.dirty isn't detecting button or icon click I assume so it didn't work out.

Yagnik56:
@danoswaltCL I have made all changes you have suggested, let me know if any changes are required.

danoswaltCL:
@Yagnik56 thanks for incorporating ngrx, I think there are still some things that can make this even smoother but this is pretty close to what I was suggesting, it is mostly organization + a couple tricks I want to show. I want to also add that I appreciate the thoroughness of checking for actual data changes, I probably would have been lazy and just always shown the speedbump no matter if the data had changed or not 😄. So this is all a bonus if we can get this working in the simplest way.

I'll try to write out my suggestions here, this may get a little long but nobody has ever really documented this for our app, so I will turn this into the missing documention. Feel free to reach out.

---

Suggestion 1: Organization:

Your `ngrx` files will work the way they are, but it's not consistent with how the rest of the app is laid out for state change patterns. If you take a look in `app -> core` you will see a bunch of feature "services" with structures usually like this:

```
core
  |_ feature
       |_ feature.service.ts
       |_ feature.data.service.ts <-- FYI data.service files are used for making http calls, not relevant for this
       |_ feature.modules.ts
       |_ store <-- ngrx stuff goes here
           |_ feature.actions.ts
           |_ feature.effects.ts
           |_ feature.reducer.ts
           |_ feature.selector.ts
           |_ feature.model.ts
 ```
 
 The component interacts with a feature.service file that handles the actual `ngrx` business logic. Ideally, we don't want `store$` or any other `ngrx` objects pulled into a component's dependencies, that should be the work of the service. That keeps all implementation details out of the component and in service layer.
 
 ```
 ## trigger a data update event from component or service ##
 feature.component -> feature.service -> service dispatches `ngrx` feature.actions ->  feature.reducer updates `store`
 
 ## get notified about data value changes ##
 `store` -> feature.selectors ->  feature.service -> feature.component
 ```
 
So we need to make sure the `ngrx` stuff you've created is located in right spot. You could either create a new feature service under `core` and scaffold out some of files like this, or you could add your stuff to an existing service.

For simplicity, I recommend using `core -> experiments` to locate your new `ngrx` stuff. It would probably have been smart for us to have an "experimentDesignStepper" feature service, but we don't. Potentially you can create that! (please name something other than `data_changed_flag`, as that could mean anything).

The rest of my suggestions will show adding to `experiments` feature `ngrx` services. If you want a new `experimentDesignStepper` service, it would be similar.

1. `experiments.model`: update the `ExperimentState` object type to include this flag, please with a name that is explicit, not just `data_changed`. If you look at other boolean flags on this object, the pattern is like `hasExperimentDesignStepperDataChanged` or something like that.

2. `experiments.actions`: put your actions here.

3. `experiments.reducer`: put your reducer functions here.

4. `experiments.selectors`: create a selector for this flag.

5. `experiment.service`: the service file is where you want to abstract handling `component` <-> `ngrx` interactions. I'd have a method for handling dispatching setting it to true, a method for handling dispatching resetting it to false, and a method to get the last emitted value from the `hasExperimentDesignStepperDataChanged$` selector observable.

(For an example of simple boolean flag in `ngrx`, take a look at how the `isAliasTableEditMode` boolean value on the Experiment State is used.)

Now, with all of that out of the way, you will call the `experiments.service` methods to handle `hasExperimentDesignStepperDataChanged` flag dispatches and updates, so that the component doesn't have to do any of the implementation or even know that `ngrx` exists.

---

Suggestion 2: Some tricks we can use to stop tracking and checking data changes manually and start using events and observables:

It would be much better if we could replace this logic for showing the speedbump:

`if (this.flag || this.participantsForm.dirty || this.participantsForm2.dirty || this.isRowRemoved)`

with this:

`if (this.experimentService.getHasExperimentDesignStepperDataChanged()) {`

Which will get the boolean from the data store, because if we are always setting that flag to true when form is dirty or a row is removed, it will always be the source of truth.
And `this.flag` is not needed at all, we only need to read this value for one purpose, there's no point in storing it on the component if we can avoid it.

- How to get rid of checking `this.isRowRemoved`

Replace `this.isRowRemoved = true` with the service method to dispatch setting the flag to true.
Then `isRowRemoved` can go away, we don't care if it's false!

- How to get rid of checking for `form.dirty`:

One angular trick we should be able to do here is to subscribe to the `statusChanges` observable on any form object during the `ngOnInit` lifecycle hook:

```typescript
this.form.statusChanges
.pipe(
  distinctUntilChanged() // this may be needed to ensure it only fires once when the value chances
)
.subscribe(
    (status) => {
      console.log(status)
      if (status === 'dirty') {
        // call the method to dispatch data change flag to true
      }
    }
  );
```

This code will get redundant, so a service method can be created:

component.ts
```typescript
this.form.statusChanges
.pipe(
  distinctUntilChanged() // this may be needed to ensure it only fires once when the value chances
)
.subscribe(
    (status) => {
      this.experimentsService.updateFormStatus(status);
    }
  );
```

experiments.service.ts
```typescript
updateFormStatus(status: string) {
  if (status === 'dirty') {
    // call the method to dispatch data change flag to true
  }
}
```

In this manner, if any form is `dirty`, it will set that flag to true automatically without any extra code.

If this works, then we're always dispatching an action to set `hasExperimentDesignStepperDataChanged` to true when the form is dirty or when a row is deleted.
So then we can just check the `hasExperimentDesignStepperDataChanged` via the service getter instead of managing a lot of extra flags.

Yagnik56:
@danoswaltCL I went through your suggestion, actually I have tried to implement ngrx as you mentioned in suggestion 1 but as I didn't had previous experience so I wasn't sure how to implement it in already existing experiment feature. Now I have organized it as you have mention in suggestion 1 with new `experimentDesignStepper` service. 

Regarding suggestion 2, I have few things to add.
- `form.dirty` can't be replaced with suggestion you have mentioned because `this.form.statusChanges` will give status as valid, invalid and disabled not true or false for dirty, pristine, untouched. Also updating flag again and again on every changes in form won't be the best solution in my view, instead we should only update flag if we click on any of buttons ( close, next, update or back ) will be much better option. 
- I have removed flags as you suggested but rather than removing `form.dirty` in if condition it's better to keep 
`if (this.experimentService.getHasExperimentDesignStepperDataChanged() || this.participantsForm.dirty || this.participantsForm2.dirty)` instead of writing like below.
image
- This way we only have to update flag if we move to next or previous tab or reset it if data updated.

danoswaltCL:
hey @Yagnik56 i ran out of time today, I will look at this tomorrow, thank you

danoswaltCL:
I see what you mean with the `statusChanges` observable, I am very surprised that it doesn't emit anything for form `dirty` or `touched`. There are definitely ways to create an observable for this though, but I'm okay without this.

I don't see any new commits, so I'm not sure on other questions without seeing more of the code you have.

If you have the `ngrx` files organized then that's probably the main thing I am asking for, it sounds like you have that change in place.

For the other suggestions, it's probably ok the way you have it. The main idea I was hoping to show is that we can simplify by using observables and reactive programming. It allows us the ability to get "live" data from a single source-of-truth, which is really powerful. Otherwise the code in Angular components gets complicated very quickly because local state has to be managed by hand.

danoswaltCL:
Okay great this looks better, I think it's almost ready to merge, but I can see that linting hasn't been run, it is supposed to run and check your code on local on commit but sometimes it doesn't work perfect because `husky` sometimes isn't installed or needed reinstalled (or if it's been committed with `--no-verify`). It hasn't been updated yet to run linting on github quality control checks.

To run this manually on frontend, you can nav to `/frontend` and run `npm run lint` or `npm run lint:fix` to autofix most things. When I pull down your code and run this, it looks like it should be easy lint fixes.

To make sure it will always run `eslint` and `prettier` on the files you are committing automatically, try reinstalling project root packages fresh: from project root directory, delete `node_modules` and `package-lock.json` and `.husky` (if it is there), and then run `npm install`. After that, you should be seeing some messages in console showing it checking linting.

Yagnik56:
> npm run lint:fix

I have ran it manually and there are no lint fixes to make, I have used --no-verify as last time pc got hanged and stay same for hour while commit so I have done manual tests later.

danoswaltCL:
Ok, yes that integration test suite is really slow. I have pushed up the `npm run lint:fix` changes, I'm not sure why it's not showing there are fixes need on your end. I added a changed in lint-staged setup that may help.

frontend/projects/upgrade/src/app/core/experiments/experiment-design-stepper.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  getHasExperimentDesignStepperDataChanged() {
+    return this.expStepperDataChangedflag;
+  }
+
+  experimentStepperDataChanged() {
+    this.store$.dispatch(experimentAction.experimentStepperDataChanged());
+  }
+
+  experimentStepperDataUpdated() {
```
danoswaltCL, line 30: last request, can you change this name to `experimentStepperDataReset` so it matches the service call and is clear this is for reseting?

Yagnik56: I have made changes but I am facing this error in back-end testing while commit, I am facing same since I pulled code from dev today. Are you facing similar error? I have tried to pull from this branch but nothing helps so not sure what to do. 
image

danoswaltCL: yes this looks like the Postgres db is not running, as the integration test suite requires a real db connection.

danoswaltCL: If this continues to be an issue, I'm 100% these backend tests pass for this and I can see that the frontend tests passed, so in this case I'd say `--no-verify` is safe to commit and push this small change and I will merge the pr

Yagnik56: sorry I fall asleep later, I will ask someone from Playpower if they can merge this PR

danoswaltCL (approved):

Issue #569, opened 2022-10-14
The api/assign request fails after the DB has been cleared


zackcl:
**Describe the bug**
The demo app uses `api/clearDB` to clear the DB whenever a new user logs in and starts the guided tour. But when using the latest dev branch, I'm not able to get a condition with either `getExperimentCondition()` or `api/assign` after the DB has been cleared with `api/clearDB`.

I get the following response:
```
{
    "details": "Error in assignment",
    "type": "Error in the assignment algorithm"
}
```
If this happens, I need to restart the UpGrade backend app to make it work again. It seems to work fine until I call `api/clearDB` though. 

**To Reproduce**
1. Open the `.env` file and set the `APP_DEMO` to `true`
2. Start UpGrade backend locally
3. Create an enrolling experiment (either manually or with `POST api/experiments`)
4. Make a `POST api/init` request
5. Make a `POST api/assign` request
6. You will see the correct response
7. Now make a `DELETE api/clearDB` request (all experiments and DB will be cleared)
8. Create an enrolling experiment again
9. Make a `POST api/init` request
10. Make a `POST api/assign` request
11. You will see the following response: `{"details": "Error in assignment", "type": "Error in the assignment algorithm"}`
12. Now you'll need to restart the UpGrade backend app to make it work again.

**Expected behavior**
The `api/assign` request should not fail even after the DB has been cleared (otherwise it will be a problem for the demo app).

**Video going through the above steps**
https://user-images.githubusercontent.com/90279765/195953946-c460b781-72ed-4c20-822b-c015b1f17d76.mov

PR #593, opened 2022-10-29
fix for assign call fail after clearDB

ppratikcr7:
Fix for issue #569 
clearDB removed the default Global Exclude Segment that was returning 500 internal server error if the backend is not restarted.


ppratikcr7:
@jreddig There is a unit test case that is failing as I have added the code to insert global exclude segment after clearing DB. Can you look into this, as I tried but couldn't resolve the unit test case. It would be easier for you. :)
Screenshot 2022-10-29 at 6 55 50 PM

ppratikcr7:
@jreddig Thanks a lot for the test case resolution :)

danoswaltCL (approved):

Issue #590, opened 2022-10-27
Migrate frontend to ESLint + Prettier, Remove TSLint


danoswaltCL:
I plan to do this in one commit, but may break it into pieces if I need to

danoswaltCL:
@mswartzCL is "Fall Release" what we should tag current work like this?

mswartzCL:
Yes, let's tag with Fall Release. Thanks!

PR #600, opened 2022-11-03
Feature/linting root level

danoswaltCL:
This will (hopefully) make it so `eslint`, `prettier`, `husky`, and `lint-staged` will only need to be installed and run from the project root.

#601 #590


danoswaltCL:
hm clientlibs fails but this builds locally with exact same commands as job runner. I'm a little confused, but maybe needs fresh eyes if anyone has a moment.

jreddig (approved):

PR #615, opened 2022-11-08
updated swagger to match current objects

jreddig:
Renamed expId/expPoint to site/target. 
Added object definitions and examples. 
Updated experiment format to reflect new changes with segments/aliases.


danoswaltCL (approved):
💯 💯 💯 💯 💯

PR #621, opened 2022-11-11
Fix/issue 588 - Change Client-side API request and response

RidhamShah:
**JAVA Changes are remaining**

### **Changes Done in JS**

- Change every `experimentPoint` to `site`
- Change every `experimetId / partitionId` to `target`
- Change `SWAGGER` according to following

### In API calls: init / group / workingGroup

-  Changed data-type of group from `Map<string,string[]>` to `Record<string,string[]>`
-  Changed data-type of workingGroup from `Map<string,string>` to `Record<string,string>`

### Init Call
- Response structure:
```
{
	userId: string,
	group: {},
	workingGroup: {},
}
```

### Group Call
- Request Type changed to `PATCH`
- Response structure:
```
{
	userId: string,
	group: {},
}
```

### WorkingGroup Call
- Request Type changed to `PATCH`
- Response structure:
```
{
	userId: string,
	workingGroup: {},
}
```

### UserAliases Call
- Request Type changed to `PATCH`
- Response structure:
```
{
	userId: string,
	aliases: [],
}
```

### Mark Call
- Removed createdAt, updatedAt, versionNumber from Response

### Assign Call
- Response structure:
```
{
	site: string,
        target: string,
	assignedCondition: {
            condition: string
	}
}
```
`Note` - Other properties of assignedCondition like experimentId, factors, alias, etc will be added in Factorial PR

### Log Call
- Removed createdAt, updatedAt, versionNumber from Response

### ClearDB Call
- Will return no content


backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -159,15 +161,18 @@ export class ExperimentUserService {
...
       let aliasesUsers = await this.userRepository.save(userAliasesDocs);
       aliasesUsers = aliasesUsers.map((user) => {
         const { originalUser, ...rest } = user;
         return { ...rest, originalUser: originalUser.id };
       });
-      return [...aliasesUsers, ...alreadyLinkedAliases];
+      aliasesToReturn = [...aliasesToReturn, ...aliasesUsers.map((alias) => alias.id)];
     }
-    return alreadyLinkedAliases;
+    return { ...userToReturn, aliases: aliasesToReturn };
```
VivekFitkariwala, line 175: Instead of `...userToReturn` just use `userId: ...`

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -271,11 +276,10 @@ export class ExperimentUserService {
     }
   }
 
-  public async clearDB(logger: UpgradeLogger): Promise<string> {
+  public async clearDB(logger: UpgradeLogger): Promise<void> {
```
VivekFitkariwala, line 279: Have you checked if status code 204 is getting returned over here?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -409,7 +404,7 @@ export class ExperimentClientController {
     @Req()
     request: AppRequest,
     experiment: MarkExperimentValidator
-  ): Promise<MonitoredDecisionPoint> {
+  ): Promise<Omit<MonitoredDecisionPoint, 'createdAt' | 'updatedAt' | 'versionNumber'>> {
```
VivekFitkariwala, line 407: Creating an explicit type will be safe just in case if we modify `MonitoredDecisionPoint`

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -623,7 +581,7 @@ export class ExperimentClientController {
     @Req()
     request: AppRequest,
     logData: LogValidator
-  ): Promise<Log[]> {
+  ): Promise<Omit<Log, 'createdAt' | 'updatedAt' | 'versionNumber'>[]> {
```
VivekFitkariwala, line 584: Create explicit type here also

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -939,11 +883,11 @@ export class ExperimentClientController {
    *            description: DEMO mode is disabled
    */
   @Delete('clearDB')
-  public clearDB(@Req() request: AppRequest): Promise<string> {
+  public clearDB(@Req() request: AppRequest) {
```
VivekFitkariwala, line 886: Is status code here returned is 204?

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
RidhamShah: code updated

VivekFitkariwala (approved):

VivekFitkariwala:
We also need the Java client changes along with this to make it compatible before deploying this.

jreddig:
I believe I've updated the Java lib to match the new request/response structure! These are breaking changes that aren't compatible with the old libraries, right? We'll need to coordinate changes in MathIA before pushing this to prod.

mswartzCL:
Dependent on MATHia and Portal updating to the latest client library.
MATHia - https://carnegielearning.atlassian.net/browse/STU-2269
Portal - https://carnegielearning.atlassian.net/browse/PORTAL-942

VivekFitkariwala:
Added code for backward compatibility.
New client lib will call `*/api/v1/*` http endpoint
Old client lib endpoint will still work `*/api/*`

jreddig:
Tested this change with api v1 and the Java libs and it looks good!

VivekFitkariwala (approved):

jreddig:
Are we expecting other changes in this branch? This needs to go in before other teams can make changes to MathIA and Portal

VivekFitkariwala:
This branch is good for merging. 
Are we integrating this version in Mathia or should we wait till factorial specific API changes are added in this version?

danoswaltCL:
> Added code for backward compatibility. New client lib will call `*/api/v1/*` http endpoint Old client lib endpoint will still work `*/api/*`
> 
> Update environment variable `CONTROLLERS=src/api/controllers/**/*Controller.ts,src/api/controllers/**/*Controller.v*.ts`

Just to be clear, "v1" refers to this new version, and some later breaking change will then be "v2"? Will docs refer to old 'default' version as version 0? Version nothing?

PR #626, opened 2022-11-15
[DEVOPS-2522] increase default node_max_old_space_size to 6G

telfekycl:


VivekFitkariwala:
Just curious are we getting heap related error?

danoswaltCL:
I believe that this was an issue when we updated node to 16, the build would run into heapspace errors here and there. If there are smarter solutions, happy to try, I believe @ppratikcr7 and I tried a few things to resolve that and found that this common solution was the simplest fix.

danoswaltCL:
should we delete the line `ENV NODE_OPTIONS=--max_old_space_size=4096` from the Dockerfile so there is not a conflict?

Also, after this goes in, what will we need to do in our AWS environments to process these terraform change? Is this functionally equivalent to an elastic beanstalk configuration getting updated by hand?

telfekycl:
@danoswaltCL  , no need to change it , its acting as default value and overwritten by the beanstalk env vars

I did the change manually , so this PR to catch the drift only

danoswaltCL:
ah ok i see that update made in prod config, ty @TamerElfeky

danoswaltCL (approved):

PR #632, opened 2022-11-21
Solved participants getting removed error

RidhamShah:


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -643,11 +643,34 @@ export class ExperimentService {
 
         experimentDoc.experimentSegmentInclusion = oldExperiment.experimentSegmentInclusion;
         experimentDoc.experimentSegmentExclusion = oldExperiment.experimentSegmentExclusion;
+
+        const currentIncludeData = experimentSegmentInclusion.segment
```
danoswaltCL, line 647: can we refactor into a reusable function so we are not duplicating code?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -643,11 +643,34 @@ export class ExperimentService {
...
+          ? {
+              ...experimentSegmentInclusion.segment,
+              userIds: experimentSegmentInclusion.segment.individualForSegment.map((x) => x.userId),
```
danoswaltCL, line 650: can we please name things what they are instead of using `x`? thank you

backend/packages/Upgrade/src/api/services/ExperimentService.ts
RidhamShah: Created a new function for both include & exclude

backend/packages/Upgrade/src/api/services/ExperimentService.ts
RidhamShah: Code updated

danoswaltCL:
@ppratikcr7 @RidhamShah what does it mean when PRs are put up in draft mode for a long time? Should I wait until they are fully "opened" to leave comments / approve? I don't know if this is still being worked on.

danoswaltCL (approved):

PR #639, opened 2022-12-02
Enhancement of participants stepper tab as per new factorial designs

ppratikcr7:
Now, the end user should not select Include Specific/Include All Except... from the dropdown in the Participants Stepper Tab.

User needs to select type as All in Include list's first row, so that they can enable the Except list.
In case the user later changes any other Type to All, then it would automatically delete all the other Include list rows as they are not relevant any more.

Screenshot 2022-12-05 at 2 20 33 PM


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.scss
```diff
@@ -91,6 +91,14 @@
       margin-left: 10%;
       width: 25%;
     }
+
```
danoswaltCL, line 94: .no-include-message-container,
.no-exclude-message-container {
      color: #8F9BB3;
}

Let's combine these rules, it's same thing 😄 . Also, is this a standard color we are using elsewhere, particularly for this kind of message text? We should probably have a color variable defined globally for just about almost everything that's re-used (such as `variables.css`). Otherwise there are going to be inconsistent little color rules sprinkled everywhere.

zackcl:
According to the original [Figma design](https://www.figma.com/file/uYxnCT2dXTftSZKFQ00Q0y/UpGrade-UI-Prototype-(latest)), the ID/Name field should be auto-filled to "All" which shouldn't be editable.

Experiments - Add Experiment-03-04

And the disabled color I used for the `+Add Member` button is `#7194FF` (`rgb(113, 148, 255)`). I'm not sure if we already have a disabled blue color used in UpGrade though.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.scss
ppratikcr7: Yeah, I will combine them. And the color code is taken from the figma design. Okay, I will create a variable for. consistent color code for such messages.

ppratikcr7:
> According to the original [Figma design](https://www.figma.com/file/uYxnCT2dXTftSZKFQ00Q0y/UpGrade-UI-Prototype-(latest)), the ID/Name field should be auto-filled to "All" which shouldn't be editable.
> 
> Experiments - Add Experiment-03-04
> 
> And the disabled color I used for the `+Add Member` button is `#7194FF` (`rgb(113, 148, 255)`). I'm not sure if we already have a disabled blue color used in UpGrade though.

Okay Zack, I will add that color code for disabled button.

ppratikcr7:
> This looks good, got a couple of comments that are primarily design-related. When first reaching this tab, we'll see both the Include and Exclude. I think we will probably not want to show the exclude until there is something selected in the Include, right?
> 
> image
> 
> More importantly, I do also see that when an Individual, Segment, or Group is selected, the Exclude section is not displayed. I think it should always get displayed when valid selection is made in the Include list, as you can definitely have Groups and Segments included, and sets of Exclusions within those Inclusions.
> 
> Even in case where the Include list is entirely Individuals (and logically we could hide the exclusion section without harm), for simplicity's sake I'd say we should just always show Exclude when the Include list has anything in it.
> 
> image
> 
> When ALL is selected, I like that we are clearing out anything in there and preventing adding anything else. We should probably have the '+ Add Member' button be visually greyed out though so that it's clear that it you can't add any more Inclusions if ALL is selected.
> 
> @zackcl @mswartzCL @amurphy-cl

@danoswaltCL So, we do have two cases: `Include All Except, Include Specific`. `Include All Except` with be the case when we select "All" in Include list, which will enable the except list to exclude a few. `Include Specific` is we just select some specific entries to include. We dont require the exclude list then. This is how we were functioning earlier too, do we really need to have the exclude list on `Include Specific`?

@nirmalpatel @VivekFitkariwala What do you think?

danoswaltCL:
I'm not sure what is listed in spec for this, maybe there is a mistake, because with 'Include specific', we have plenty of use-cases where we need the exclude list still. For example, if we want to limit a study to just one school group, we would want "Include Specific" logic to only allow that school, but individual users within that school may opt out of and need excluded specifically.

The XPrize study that we are currently running in production is set up in this way, we have schools that are included specifically, and a list of individual students that have opted out of all studies, so this logic does currently exist to limit the set of participants from both lists.

I think the only time we really ought to hide the exclude list section is when the "Include" section list is empty. If the Include section has >= 1 member, show "Exclude" section, otherwise clear it out and hide it.

danoswaltCL:
@zackcl for your comment "According to the original [Figma design](https://www.figma.com/file/uYxnCT2dXTftSZKFQ00Q0y/UpGrade-UI-Prototype-(latest)), the ID/Name field should be auto-filled to "All" which shouldn't be editable.", if we want this to not be editable (except to make a different selection in the dropdown), then the figma design should probably not show the "action buttons" for edit/done and clear.

zackcl:
> Include Specific is we just select some specific entries to include. We don't require the exclude list then. This is how we were functioning earlier too, do we really need to have the exclude list on Include Specific?

@ppratikcr7 I think we've been requiring the exclude list even when `Include Specific` is used. The below screenshots are the previous Figma design.

Stepper - Participants 2 (1)
Stepper - Participants 1 (1)

zackcl:
@ppratikcr7 @danoswaltCL Because the current Figma design I shared works with the edit mode which is not implemented in our tables yet, I thought about how things should behave without the edit mode.

1. Only the Include section should be shown in the beginning by default.
Experiments - Add Experiment-03-01
2. When the user clicks the "Add Member" button, a row will be added to the Include table, and the Except section should appear at the bottom. And if the user deletes the row (empties the Include table), it will go back to 1 (the Except table should be cleared and hidden).
Experiments - Add Experiment-03-02 (3)
3. When the "All" type is selected, all other rows in the Include table should be removed (if exists), and the ID/Name field should be auto-filled to "All" which shouldn't be editable. And the "Add Member" button should be disabled and shown in a lighter color (#7194FF). The button should be enabled again if the type changes to something else.
Experiments - Add Experiment-03-04 (4)

I hope this makes sense.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -56,6 +56,7 @@ export class ExperimentParticipantsComponent implements OnInit {
   subSegmentIds = [];
   segmentNameId = new Map();
   subSegmentTypes: any[];
+  allSubSegmentTypes: any[];
```
VivekFitkariwala, line 59: Use proper types here

VivekFitkariwala (dismissed):

ppratikcr7:
@zackcl @danoswaltCL  The light-blue color css is not working for disabled 'Add Member' button. The original dark blue color gets override. It should work as other button's are also having the similar css styling. @Yagnik56 will look into this later. Rest all things as per your peer review are covered in my last [commit](https://github.com/CarnegieLearningWeb/UpGrade/pull/639/commits/f70dc2d94b4676de404b0e668ab250995f6ee072).

ppratikcr7:
> @zackcl @danoswaltCL The light-blue color css is not working for disabled 'Add Member' button. The original dark blue color gets override. It should work as other button's are also having the similar css styling. @Yagnik56 will look into this later. Rest all things as per your peer review are covered in my last [commit](https://github.com/CarnegieLearningWeb/UpGrade/pull/639/commits/f70dc2d94b4676de404b0e668ab250995f6ee072).

@danoswaltCL @zackcl @jreddig You can review this PR now. The issue is resolved.

zackcl:
@ppratikcr7 I reviewed this and I changed "ALL" option to "All" just like in the original design. I also fixed one bug so that the rest of the rows are correctly deleted when the first row changes to "All" (Previously, only the second row was deleted).

And one thing I would like to suggest updating is that currently a dropdown is used for completing the Segment Name, but I think a combo box should be used for completing the Segment Name considering there can be hundreds of Segments in the future, and it's easier for the user to type and select.

Here's the original design in Figma:
Experiments - Add Experiment-03-19

zackcl:
And I found out that the ID/Name field is not being cleared when the type is changed in the Except table (The Include table works properly). Please see the video recording:

https://user-images.githubusercontent.com/90279765/213243126-d6526952-520d-40f7-bea9-2aebdf08843c.mov

I think this should be fixed.

danoswaltCL:
@zackcl @ppratikcr7 where we at on this one? Can this one be merged in after the conflicts are resolved?

ppratikcr7:
> @zackcl @ppratikcr7 where we at on this one? Can this one be merged in after the conflicts are resolved?

@Yagnik56 is working on the last review comment from zack and the conflicts. Should be done soon.

zackcl:
I will test this branch

zackcl:
@Yagnik56 @ppratikcr7 I just tested this branch and the issues I mentioned earlier (combo box should be used for the Segment Name, ID/Name field is not being cleared when the type is changed in the Except table) seem to be fixed, but I'm getting an error when selecting a Segment Type in the Except table. Please check this video:

https://user-images.githubusercontent.com/90279765/219807911-7ed2afea-c6c2-4108-a8da-b0c5a5248461.mov

Yagnik56:
> @Yagnik56 @ppratikcr7 I just tested this branch and the issues I mentioned earlier (combo box should be used for the Segment Name, ID/Name field is not being cleared when the type is changed in the Except table) seem to be fixed, but I'm getting an error when selecting a Segment Type in the Except table. Please check this video:

@zackcl Can you check now? I have resolved that issue from my end

zackcl:
@Yagnik56 Now the error is gone. But can you make the "Add Segment Id/Name from Suggested Values only" message appear when the user types something into the `Id/Name` field instead of selecting one from the menu? Currently, the message shows up right after you select the "Segment" type as if something's wrong with the type, but the message is about the `Id/Name` field so I think it should appear when the user types something into the `Id/Name` field (instead of selecting one from the menu).

https://user-images.githubusercontent.com/90279765/220438407-33540f89-1b9e-4f2b-9464-b351464e9196.mov

Yagnik56:
> @Yagnik56 Now the error is gone. But can you make the "Add Segment Id/Name from Suggested Values only" message appear when the user types something into the `Id/Name` field instead of selecting one from the menu? Currently, the message shows up right after you select the "Segment" type as if something's wrong with the type, but the message is about the `Id/Name` field so I think it should appear when the user types something into the `Id/Name` field (instead of selecting one from the menu).

@zackcl I made change so that if we type something then only warning will appear.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -248,6 +306,22 @@ export class ExperimentParticipantsComponent implements OnInit {
     });
   }
 
+  checkSegmentvalidity(members : any, table: number){
```
danoswaltCL, line 309: please provide type

danoswaltCL (dismissed):
Please make sure to have all input and return types annotated

zackcl:
> @zackcl I made change so that if we type something then only warning will appear.

@Yagnik56 Thanks, I could confirm this issue was fixed.

I think the combo box for the Segment field is not working as expected in the Include/Except tables.
The list (selection menu) should only appear if there's any option that matches (starts with) what you typed.
It should work like how it works in the Decision Points and Conditions table. Please check this video:

https://user-images.githubusercontent.com/90279765/220692740-bcf0e645-fbe4-4646-94c8-4c932d4cedfd.mov

zackcl:
> But I think the combo box (when the Segment type is selected) is not working as expected in the Include/Except tables.
> The list (selection menu) should only appear if there's any option that matches (starts with) what you typed.
> It should work like how it works in the Decision Points and Conditions table.

@Yagnik56 Please let me know if this is not so simple to fix, I think we can move on and merge this for now as this is a minor issue.

Yagnik56:
> > But I think the combo box (when the Segment type is selected) is not working as expected in the Include/Except tables.
> > The list (selection menu) should only appear if there's any option that matches (starts with) what you typed.
> > It should work like how it works in the Decision Points and Conditions table.
> 
> @Yagnik56 Please let me know if this is not so simple to fix, I think we can move on and merge this for now as this is a minor issue.

@zackcl I tried to implement your suggestion but it had some issue, it will be good if we can take this issue separately. I kept code as it is in ts file, but moved back to using array in html so we can use same code later. If we don't have any other issues we can move forward with this PR.

ppratikcr7:
> > But I think the combo box (when the Segment type is selected) is not working as expected in the Include/Except tables.
> > The list (selection menu) should only appear if there's any option that matches (starts with) what you typed.
> > It should work like how it works in the Decision Points and Conditions table.
> 
> @Yagnik56 Please let me know if this is not so simple to fix, I think we can move on and merge this for now as this is a minor issue.

Yes, please review and let this get merged for now, Yagnik will work on the filtering of values in separate PR. :)

zackcl (approved):

Issue #644, opened 2022-12-05
An imported experiment should always be inactive


zackcl:
**Describe the bug**
~Currently, if an enrolling experiment is exported, the exported JSON file contains `"state":"enrolling"`. So if the file is imported, the experiment status is set to "enrolling" making it not possible to edit or schedule the experiment. Therefore, I believe an exported experiment's status should always be "inactive".~

An imported experiment should always be set to "inactive". (updated due to Steve's comment below)

**To Reproduce**
1. Create an experiment in UpGrade
2. Set the experiment status to "enrolling"
3. Export the experiment file in JSON
4. Delete the experiment and try importing the exported file
5. You will see the imported experiment is set to "enrolling"

**Expected behavior**
An imported experiment's status should always be "inactive".

**Video**
https://user-images.githubusercontent.com/90279765/205711101-85a999a2-91e0-47e7-a395-3ef7c95ea80b.mov

SritterCL:
I would allow the export (you might want to document the experiment design or duplicate and modify it for some purpose), but I'd ignore "state" when importing and always set the imported experiment to "inactive".

PR #646, opened 2022-12-05
Set the exported experiment's status to inactive

zackcl:
This PR fixes the https://github.com/CarnegieLearningWeb/UpGrade/issues/644

It seems to work fine but I'm not sure if this is the proper way to do it.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/experiments/store/experiments.effects.ts
+++ b/frontend/projects/upgrade/src/app/core/experiments/store/experiments.effects.ts
@@ -45,6 +45,7 @@ import { interval } from 'rxjs';
 import { selectCurrentUser } from '../../auth/store/auth.selectors';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { ENV, Environment } from '../../../../environments/environment-types';
+import { EXPERIMENT_STATE } from 'upgrade_types';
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/core/experiments/store/experiments.effects.ts
```diff
@@ -486,6 +487,7 @@ export class ExperimentEffects {
                 this.download('Experiments.zip', content, true);
               });
             } else {
+              data[0].state = EXPERIMENT_STATE.INACTIVE;
```
danoswaltCL, line 490: i think this approach works fine, you'll also want to do it in the 'bulk export' in the if clause.

```
data.forEach((experiment, index) => {
                experiment.state = EXPERIMENT_STATE.INACTIVE;
                zip.file(experiment.name + ' (File ' + (index + 1) + ').json', JSON.stringify(experiment));
              });
              zip.generateAsync({ type: 'base64' }).then((content) => {
                this.download('Experiments.zip', content, true);
              });
```

zackcl:
Closing this due to Steve's comment: https://github.com/CarnegieLearningWeb/UpGrade/issues/644#issuecomment-1337973390

PR #661, opened 2022-12-08
Fix/issue 620: Enhancement for Importing Experiment feature

ppratikcr7:
In this PR, we are enhancing the importing of experiment JSON's feature.

It will now display the missing fields(if any) in the experiment JSON per file, in the error's section. This will help us find which file out of the bulk upload's is missing what. 

It will also let you know in case the uploaded JSON file version is either newer or an older version. 

**Note**: I have also updated the backend version to `3.0.10` in the frontend `package.json` file, which would reflect on our UI header logo section as well and will help in exporting the experiment with correct backend version(as per the release version). Let me know in case we still want to keep this version at `1.0.0` due to some specific reason if any.

Screenshot 2022-12-08 at 11 26 32 PM


frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
```diff
@@ -76,14 +78,32 @@ export class ImportExperimentComponent implements OnInit {
...
     this.openSnackBar();
   }
-  async validateExperimentJSONVersion(experimentInfo: any): Promise<any> {
-    const version = await this.versionService.getVersion();
-    if (experimentInfo.backendVersion !== version) {
-      return false;
+
+  compareVersion(currentBackendVersion, uploadedExperimentBackendVersion) {
+    currentBackendVersion = currentBackendVersion
+      .split('.')
```
danoswaltCL, line 84: why add this padding?

zackcl:
This looks great but does it actually validate if the experiment is importable? I tried importing a corrupted file which I renamed `conditionCode` to `conditionCode2` and it doesn't seem to be detecting this on the "Import Experiment" popup window. 
Screen Shot 2022-12-09 at 5 08 33 PM

And when I click "Import", it just shows the "Network call failed" error and this is kind of hidden by the "Import Segments and add them in the imported experiment if required" message box. 
Screen Shot 2022-12-09 at 5 07 56 PM

I'm not sure if we really need this message box and in the future, I think we can add a warning icon if something is missing like in the Figma design:
Experiment Detail - Participants (Inactive) 3

Anyways, I think the "Import Experiment" popup should actually validate if the file is importable or not. We should also check if there's already the same experiment using the same experiment ID.

And my suggestion is to have 3 columns that are `File`, `Warning`, and `Error` to distinguish between the two. Then it makes more sense to show the "Non-error experiments will be imported" because currently, it looks like some error experiments can still be imported.

ppratikcr7:
> This looks great but does it actually validate if the experiment is importable? I tried importing a corrupted file which I renamed `conditionCode` to `conditionCode2` and it doesn't seem to be detecting this on the "Import Experiment" popup window. Screen Shot 2022-12-09 at 5 08 33 PM
> 
> And when I click "Import", it just shows the "Network call failed" error and this is kind of hidden by the "Import Segments and add them in the imported experiment if required" message box. Screen Shot 2022-12-09 at 5 07 56 PM
> 
> I'm not sure if we really need this message box and in the future, I think we can add a warning icon if something is missing like in the Figma design. Maybe we can open another issue about this. Experiment Detail - Participants (Inactive) 3
> 
> Anyways, I think the "Import Experiment" popup should actually validate if the file is importable or not. We should also check if there's already an experiment using the same experiment ID.
> 
> And my suggestion is to have 3 columns that are `File`, `Warning`, and `Error` to distinguish between warning (still importable) and error (not importable). And if there's any file that has an error, then I think it's better to disable the "Import" button, and expect them to close and re-import the files, or add a trash icon on each row so the user can delete the files with error and click the import button - I think this is an ideal way.

Okay. Currently we had a check for missing fields. I will add validation for field names as well. We are thinking that we should have only 1 column for Error/Warning with a header "Issues/Problems" and have the material UI red and yellow fonts for error and warnings so that we dont have several empty cells in between for two separate columns for Error and warnings.

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
```diff
@@ -30,14 +30,15 @@ interface ImportExperimentJSON {
...
-  isExperimentJSONVersionValid = true;
+  experimentJSONVersionStatus = 0;
   missingAllProperties: string;
   allPartitions = [];
   allPartitionsSub: Subscription;
   allExperiments: Experiment[] = [];
   importFileErrorsDataSource = new MatTableDataSource<{ filename: string; error: string }>();
   importFileErrors: { filename: string; error: string }[] = [];
   displayedColumns: string[] = ['File Name', 'Error'];
+  uploadedFileCount: any;
```
VivekFitkariwala, line 41: Should we use number here?

VivekFitkariwala (approved):

PR #667, opened 2022-12-21
Key Value Bubbles (dev tool)

danoswaltCL:
This tool give you the ability to create moveable "bubbles" on the screen while working that display key/values, basically a console.log in the UI, except it will live update and live on the screen.

This is a debugging tool that had sort of half-developed sitting in a branch on my end that I decided to finally polish up and share. I think could be really useful when troubleshooting UI issues, like implementing the factorial UI @Yagnik56 is working on, for example, to see what form data is looking like as you click and type through a complicated form. I find this pretty convenient for quick answers compared to digging through debugger.

You can drag the bubbles wherever too; I am still trying to get it to save the last position of the bubble if you dragged it so that it will return to that position after reloading, but this was trickier than I expected. 

Also: it fully contained in its own module and won't be a part of any prod build. Only works in dev builds.

image

https://user-images.githubusercontent.com/97542869/208991813-688c68df-ca09-4c4b-a915-c85ec02e3c66.mov


jreddig (approved):
much needed!

PR #672, opened 2023-01-04
Add Edit Mode to the Decision Points table

zackcl:
https://user-images.githubusercontent.com/90279765/210600102-c7394b7b-9bb1-4e4a-a345-9cb3c720cc1c.mov

I've mostly just copied the code written for the Conditions table by @danoswaltCL so I feel most changes are done correctly, but please review them, especially the `removeConditionOrPartition()` and `isDecisionPointTableRowValid()` functions in the `experiment-design.component.ts` file. And I will update the styles in a separate PR once this is merged.

@danoswaltCL I have a question, the Decision Points table currently validates the duplicated names while editing (as you can see from the video) so I think this should be fixed. To do this, can I simply delete the following code from the `experiment-design.component.ts` file?
```
this.experimentDesignForm.get('partitions').valueChanges.subscribe((newValues) => {
      this.validatePartitionNames(newValues);
});
```


danoswaltCL (approved):
perfecto, I think this nails it. looks great!

PR #686, opened 2023-01-13
use default button disable style for alias conditions button

danoswaltCL:
Just making the alias-button disabled state same as the other buttons on the page (lighter blue, not grey).


jreddig (approved):
Checked it out and it looks good locally!

PR #687, opened 2023-01-17
Creating and Editing Factorial Experiment through stepper

RidhamShah:


ppratikcr7:
> i can hopefully devote some time to give some review tomorrow (Wednesday) if it is ready for that

Yes, it's ready with functionality wise. But we are facing jest timeout issues in test cases. Still you can do the code review.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -127,6 +172,172 @@ export class ExperimentDesignStepperService {
...
+
+    if (tableData.length > 0) {
+      tableData.forEach((factorialConditionTableRow) => {
+        sumOfAssignmentWeights += parseFloat(Number(factorialConditionTableRow.weight).toFixed(1));
+        parseFloat(Number(factorialConditionTableRow.weight).toFixed(1)) < 0 ? (negativeweightError = true) : null;
+      });
+
+      // checking if sum is not equal to 100
+      if (Math.round(sumOfAssignmentWeights) !== 100.0) {
+        return true || negativeweightError;
```
danoswaltCL, line 300: this is will always return true, ` || negativeweightError` will never evaluate.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -127,6 +172,172 @@ export class ExperimentDesignStepperService {
...
+      } else {
+        return false || negativeweightError;
```
danoswaltCL, line 302: this will always return `negativeweightError`, so the `false ||` left side isn't needed.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -127,6 +172,172 @@ export class ExperimentDesignStepperService {
...
+
+    return factorialConditionsRequestObject;
+  }
+
+  checkConditionTableValidity() {
+    const tableData = this.getFactorialConditionTableData();
+    let sumOfAssignmentWeights = 0.0;
+    let negativeweightError = false;
+
+    if (tableData.length > 0) {
```
danoswaltCL, line 292: there has to be a simpler way to do this.

1. negative number check can probably be eliminated before we get to this point by using native html number validation (min = 0)

2. this can probably be accomplished without any type-casting between numbers and strings with a simple `reduce` function.

3. why-oh-why use a ternary for this:
`parseFloat(Number(factorialConditionTableRow.weight).toFixed(1)) < 0 ? (negativeweightError = true) : null;`

frontend/projects/upgrade/src/app/core/experiment-design-stepper/store/experiment-design-stepper.model.ts
```diff
@@ -32,18 +40,101 @@ export interface ConditionsTableRowData {
...
+  factorialDesignData: ExperimentFactorialDesignData;
+  factorialConditionsTableData: FactorialConditionTableRowData[];
+  hasExperimentStepperDataChanged: boolean;
 }
 
 export interface State extends AppState {
   experimentDesignStepper: ExperimentDesignStepperState;
 }
+
+export const DUMMY_CONDITION_TABLE_DATA = {
```
danoswaltCL, line 96: please remove

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -142,10 +142,10 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
 
     if (this.isContextChanged) {
       this.isContextChanged = false;
-      this.partition.clear();
-      this.condition.clear();
-      this.partitionDataSource.next(this.partition.controls);
-      this.conditionDataSource.next(this.condition.controls);
+      this.partition?.clear();
```
danoswaltCL, line 145: why are changes made to simple experiment form?

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/new-experiment/new-experiment.component.html
```diff
@@ -26,15 +26,37 @@
...
+    <home-factorial-experiment-design *ngIf="newExperimentData?.type=='Factorial'"
+      [experimentInfo]="experimentInfo"
+      [currentContext]="currentContext"
+      [isContextChanged]="isContextChanged"
+      [animationCompleteStepperIndex]="animationCompletedIndex"
+      (emitExperimentDialogEvent)="getExperimentData($event)"
+    ></home-factorial-experiment-design>
   </mat-step>
 
+  <!-- <mat-step>
```
danoswaltCL, line 45: please remove old commented code

frontend/projects/upgrade/src/assets/js/demo-app.js
```diff
@@ -110,7 +110,7 @@ if (window.self !== window.top) {
       activeWindowEvent.target.removeEventListener(activeWindowEvent.type, activeWindowEvent.callback, true);
```
danoswaltCL, line 110: why are demo app changes in this PR?

frontend/projects/upgrade/src/app/features/dashboard/home/components/table-row/table-row.component.html
```diff
@@ -32,7 +32,6 @@
```
danoswaltCL, line 32: why are linting changes committed for files that aren't relevant to this PR?

danoswaltCL (changes requested):
I'm sure this probably works for most requirements, but it bothers me that we seem to be okay with code that works but is not cleaned up from the development process.

Can we please use a tool like "CodeMetrics" extension in VsCode to review code complexity of functions in the typescript files of this PR? It will do a static scan and score functions based on a cyclomatic complexity as we develop. There are too many places to go through and offer specific code review comments. The validation methods in the component.ts for example can definitely be simplified.

This function for example is rated a 10, which is okay for a complicated algorithm maybe, but this should be a very simple function to sum the weights and return a boolean:

image

It doesn't have to be a hard rule to keep it under a certain number, but it's usually a pretty good hint that when the score is high, it's going to be hard to understand the code.

There are a lot of places where we have high code complexity for no reason. This is another giant PR, if we're going to have giant PRs, we have to have cleaner code please or else it's impossible to review.

Please also review for dead code and naming conventions (camel case, name things carefully so they are clear about what they are or what they do). Brand new code should try not to use old terminology like "partitions" / "exp points" / "exp ids" where we can avoid it.

I will go in and clean up some of the code that I've written also, I was not aware this would be going up for PR already.

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/conditions-table/conditions-table.component.ts
```diff
@@ -0,0 +1,270 @@
...
+    const newTableData = this.getFactorialConditions().controls.map((control, index) => {
+      const thisRowWeight: number = control.get('include').value ? equalWeight : 0;
+      control.get('weight').setValue(thisRowWeight.toFixed(1));
+      return { ...tableData[index], weight: this.experimentDesignStepperService.formatDisplayWeight(thisRowWeight) };
+    });
+
+    return newTableData;
+  }
+
+  validateWeightSumEqualTo100(factorialConditions: any) {
```
danoswaltCL, line 179: please replace `any` with proper types

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
Yagnik56: components were loading before we choose which experiment in overview tab so we added this.

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/conditions-table/conditions-table.component.ts
Yagnik56: it is getting direct new form values, which don't have specific type so it's kept as any.

frontend/projects/upgrade/src/app/features/dashboard/home/components/table-row/table-row.component.html
Yagnik56: I am not sure but these changes are due to merge conflicts we were getting right? @RidhamShah

frontend/projects/upgrade/src/assets/js/demo-app.js
Yagnik56: they are mostly related to formatting I guess as I format document with prettier in last so...

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/conditions-table/conditions-table.component.ts
danoswaltCL: it can definitely get a specific type, if there isn't one created we should create that type. we shouldn't be using `any` anywhere in new code without a really good reason.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
Yagnik56: 1. it's not number so min=0 won't work here and it will give that side button if we make it number so let's just keep it that way.
image

2. reduce func will do same thing this is doing, plus we have to convert string to number in that so this seems more optimal in my view.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
Yagnik56: done

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
Yagnik56: done

frontend/projects/upgrade/src/app/core/experiment-design-stepper/store/experiment-design-stepper.model.ts
Yagnik56: done

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/new-experiment/new-experiment.component.html
Yagnik56: done

frontend/projects/upgrade/src/assets/js/demo-app.js
danoswaltCL: we should remove linting changes that aren't relevant to the files in factorial feature work

danoswaltCL: however if it's the lint-staged prettier hook that is somehow producing this when you commit your work, let me know, it shouldn't be doing that

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
danoswaltCL: right ok

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
```diff
@@ -0,0 +1,603 @@
...
+      duration: 500,
+      easing: 'easeOutCubic',
+    });
+  }
+
+  get factor(): FormArray {
+    return this.factorialExperimentDesignForm?.get('factors') as FormArray;
+  }
+
+  get level(): FormArray {
```
danoswaltCL, line 588: this should be dead code and can be removed

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
```diff
@@ -0,0 +1,603 @@
...
+        },
+        path: NewExperimentPaths.EXPERIMENT_DESIGN,
+      });
+
+      // scroll back to the factors table
+      this.scrollToFactorsTable();
+    }
+  }
+
+  convertToPartitionData(factorialExperimentDesignFormData) {
```
danoswaltCL, line 528: @Yagnik56 @RidhamShah @ppratikcr7 @VivekFitkariwala This is the type of function I'm looking at when I say we need to clean things up, it's not really ready for PR. I'd like to prevent us from being okay with continuing to merge in code that works functionally but is a not cleaned up for the benefit of other developers on the team, it's really hard to work this way. 

- no types are provided
- outdated terms are used (partition)
- code complexity is too high for routine operations, this can easily be broken down into small readable functions

This kind of logic could probably be moved to stepper service also, that's what it's there for, as there's no dependency on the component class or template bindings.

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
```diff
@@ -0,0 +1,603 @@
...
+
+  expandedId: number = null;
+  levelIds: string[] = [];
+
+  factorDisplayedColumns = ['expandIcon', 'factor', 'site', 'target', 'removeFactor'];
+  levelDisplayedColumns = ['level', 'alias', 'removeLevel'];
+
+  // Used for condition code, experiment point and ids auto complete dropdownfilteredExpFactors$: Observable<string[]>[] = [];
+  filteredExpFactors$: Observable<string[]>[] = [];
+  filteredExpPoints$: Observable<string[]>[] = [];
```
danoswaltCL, line 81: `ExpPoint` should be changed to `Site` or `DecisionPointSite`
`ExpId` should be changed to `Target` or `DecisionPointTarget`
`Partition` should be changed to `DecisionPoint`

There are some places where these old terms are deeply embedded and are hard to change, but when we're just creating our own local var names, we should stick to the correct terms.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
danoswaltCL: css can remove the number input controls, we have code in `experiment-design.scss` to do exactly this. 

```
 /* Remove arrows/spinners from input type number */
  input::-webkit-outer-spin-button,
  input::-webkit-inner-spin-button {
    -webkit-appearance: none;
    margin: 0;
  }

  input[type='number'] {
    -moz-appearance: textfield;
  }
}

if we can hide those controls, I'm not sure what benefit there is to writing out a routine native validation manually.

I see that there's already code in the component that is checking for valid weights also, so why is it reimplemented here?

**Comment by danoswaltCL** [comment]
I pushed a change to remove strict validation of context metadata in this branch.

https://carnegielearning.slack.com/archives/CTD5UG8R2/p1674241667763139

File: frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts [hunk 13]
```diff
@@ -127,6 +172,174 @@ export class ExperimentDesignStepperService {
...
+    return tableData;
+  }
+
+  mergeExistingConditionsTableData(experimentInfo: ExperimentVM): FactorialConditionTableRowData[] {
+    const existingConditions = experimentInfo.conditions;
+    const existingConditionAliases = experimentInfo.conditionAliases;
+    const existingPartitions = experimentInfo.partitions;
+
+    const levelOrder = {};
+    existingPartitions.forEach((partition) => {
```
danoswaltCL, line 221: I'm not sure this sort strategy will always work? Levels will always "tie" with the other levels in a factor, right? so what guarantees the order of levels within a factor?

We're sending the index as `order` to the backend for `factors`, "`partitions`", and `conditions`, so can we just add an `order` property in for `levelCombinationsElements` and `conditionAliases` so we don't have to do a workaround like this?

@RidhamShah @ppratikcr7 @VivekFitkariwala ?

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
Yagnik56: min=0 isn't working, it will just limit side controls to 0 but u can type negative values.
In component we are checking for error message in condition table but if person ignore warning or don't go to condition table and move next then this validation will stop it from moving forward as we can't stop them from condition table func so this is added.

frontend/projects/upgrade/src/assets/js/demo-app.js
Yagnik56: my prettier isn't working so I don't, I think it has only formatting related changes

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
Yagnik56: no level are called and it will give error if removed

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
Yagnik56: type was missed, it is added now

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
Yagnik56: done

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
danoswaltCL: `Validators.min(0)` would be used in conjunction with this, but we have abandoned Angular's built-in validators for some reason and decided to code everything manually, so I guess we'll code it manually. (see `experiment-form-validators.ts` for an example of the normal way for handling validation in Angular)

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
danoswaltCL: ok i see, I was thinking we had replaced this with `getLevels()`, but it's different.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
danoswaltCL: ok, maybe I've got it a little confused then what was showing up out of order exactly, it seems like there still should be a good way to save the order when submitting to backend instead of doing this work in UI. same for `conditionAliases` table in simple experiment.

RidhamShah: We will look into it.

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
danoswaltCL: ok, sorry to pick on this particular function, it's not at all the most complex one in here, just happened to be where I got frustrated trying to review because the PR has a lot of changes and it's not always easy to read.

I guess we have different philosophies on coding style, so maybe I'll just give this up if the rest of the team disagrees and does not find this PR hard to read, but I want to explain why I made this comment and we can move on.

The thing is I see lots of long functions with nested conditionals and subtasks in this PR (and in the codebase in general) that could easily be simplified, purely for the readability of future developers, and I think that is a hugely important thing.

When you're working with existing code, would you rather see small functions that can be quickly understood by a even a new developer, or dig through long mazes of nested conditionals and anonymous functions that you have to go find the original developer to explain?

Or look at it this way, there's another huge benefit: If I asked you to write a jest unit test for this function, how many cases would it take to get 100% coverage? How many ways would you have to mock up data? How about the `ngOnInit` function, which has all kinds of things going on? That's what I mean by code complexity. If it's hard for a machine to process all the possibilities, a human is going to make mistakes. A set of small pure functions is exponentially easier to mock and test than one umbrella function that technically accomplishes exactly the same tasks.

The hard work of getting the functionality is done, polishing the code for readability is free and easy. Once it gets merged in, we can't really say we'll clean it up later, because that rarely actually happens. Better spend a little extra time now than spend weeks debugging complex code later.

danoswaltCL: and obviously, please let me know where my contributions to this PR can be improved also. I almost never hear from anyone on Playpower in PRs so I never know what others are thinking (until I've broken something in dev branch).

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
Yagnik56: done for existingPartition

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/conditions-table/conditions-table.component.ts
```diff
@@ -78,7 +78,7 @@ export class ConditionsTableComponent implements OnInit, OnDestroy {
       const formControls = this._formBuilder.group({
         levels: [tableDataRow.levels],
         alias: [tableDataRow.alias],
-        weight: [this.experimentDesignStepperService.formatDisplayWeight(tableDataRow.weight)],
+        weight: [this.experimentDesignStepperService.formatDisplayWeight(tableDataRow.weight),Validators.min(0)],
```
danoswaltCL, line 81: If we're not hooking into Angular form validation techniques for showing the message, this is isn't adding anything to what is already there, it can be taken out.

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
ppratikcr7: @danoswaltCL You are absolutely right. We should do as much simplification as possible to make code readable friendly for all. We will make sure to get it done in this PR and future PR's as well. There is no significant improvements that I could see in your contributions, else would have definitely pointed it out. Just a few console logs that you have kept in the conditions table component. :)

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -127,6 +174,210 @@ export class ExperimentDesignStepperService {
...
+        };
+
+        tableData.push(tableRow);
+      });
+    });
+
+    return tableData;
+  }
+
+  convertToPartitionData(factorialExperimentDesignFormData: ExperimentFactorialDesignData) {
```
danoswaltCL, line 217: please change this to something that does not use the word partition.

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -127,6 +174,210 @@ export class ExperimentDesignStepperService {
...
+        order: factorOrder++,
+        levels: currentLevels,
+      };
+      if (
+        !decisionPoints.find(
+          (existingDecisionPoint) =>
+            existingDecisionPoint.site === decisionPoint.site && existingDecisionPoint.target === decisionPoint.target
+        )?.factors.push(currentFactors)
+      ) {
+        const partitionData = {
```
danoswaltCL, line 237: please change this, it still says partition

danoswaltCL (approved):

PR #689, opened 2023-01-18
Service unit tests

jreddig:
added new unit test files, updated coverage on existing files, and linted


jreddig:
forgot i switched to node 12 to get more helpful messages and never switched back to 16. good to go now!

danoswaltCL (approved):

PR #699, opened 2023-01-25
Feature/clean up simple experiment terms

danoswaltCL:
This change is primarily focused on cleaning up `partition`, `exp id`, `exp point` and some other things the `experiment-design` component. Other files outside of this component should only have simple name changes where shared identifiers like Type names or `en.json` keys were shared with symbols used in the `experiment-design` component.

- Eliminate most use of `partition`, `exp ids`, `exp points`
- Refactor to use a new `constants.ts` file for most strings.
- Add in some missing types in a few places
- Minor other refactoring to reduce redundancies and simplify long-winded code

I plan on going through each component in UI eventually, it should be simpler elsewhere, to update all outdated terms.


jreddig:
I fully support this initiative! Nothing seemed out of the ordinary when I pulled this branch and ran it locally, but someone who has worked more recently on the frontend code should confirm that there's no red flags

PR #718, opened 2023-02-22
Solved errors and increased performance of paginated API

RidhamShah:
Removed unnecessary left joins in paginated API to increase the performance speed

solved an error where paginated API doesn't send all experiments due to duplicated experiment ids


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -154,7 +142,8 @@ export class ExperimentService {
 
     queryBuilder = queryBuilder.skip(skip).take(take);
```
VivekFitkariwala, line 143: This is no longer needed as we have added limit and offset

RidhamShah: updated

danoswaltCL (approved):
looks good to me 👍

PR #721, opened 2023-02-25
V2 Version of Client side SDK

RidhamShah:
Added context to the header while initiating the user.
Corresponding changes are done in client APIs


danoswaltCL:
per this discussion, https://carnegielearning.slack.com/archives/CTD5UG8R2/p1678818605387439, @VivekFitkariwala @RidhamShah @ppratikcr7 @nirmalpatel sry to suggest another name change for this, but it may actually make more sense to match the new API version to the major version of the app. The `Ardent Angle` release should get bumped to `4.x`, so can we amend this to make say `v4`?

That way, all of our documentation can refer to same `4` version. Any reason not to?

VivekFitkariwala:
@danoswaltCL we will update the version to v4 in the new clientSDK. And we will create a new PR with changes of frontend, backend and clientSDK.
@jreddig please don't start now as according to the discussion from the last meeting we need to update the clientSDK.

RidhamShah:
The PR is not needed now as the changes are applied and pushed in `Factor & Client side changes` [PR](https://github.com/CarnegieLearningWeb/UpGrade/pull/748)

PR #722, opened 2023-03-01
Feature/google auth update

danoswaltCL:
This PR migrates google auth from the "gapi" libraries to the "gsi" library https://developers.google.com/identity/gsi/web/guides/migration

- Removes GAPI library code in `assets/js`
- Adds GSI script to index.html `<script src="https://accounts.google.com/gsi/client"></script>`
- Uses localStorage to manage in-browser automatic "returning user" login (previously GAPI managed this for us)
- Reorganized and renamed some things (stuff in auth.effects should really have been in auth.service)
- Removed a few frontend specs that would need redone from scratch... for now... this is a little time-sensitive
- No backend changes (though it should be noted we're not technically following the recommendations for decoding the Google Auth JWT and managing user session flow from backend, it's all getting handled in frontend. This is in line with keeping the changes for this minimal and not interrupting working code, but it could be something we may want to revisit in the future).

The UX experience should mostly be unchanged, though you will now notice that the google sign-in button mysteriously knows who you are before you click it.

image


zackcl:
This is a minor thing but how about renaming all `gapiClientId` to `googleClientId` since we no longer use `gapi`?

danoswaltCL:
> This is a minor thing but how about renaming all `gapiClientId` to `googleClientId` since we no longer use `gapi`?

yes, that's probably a good idea for completeness, we'll need to make that change in the `environment.json` files as well, and any local dev copies of environment files will also need updated.

danoswaltCL:
@ppratikcr7 @VivekFitkariwala FYI for naming consistency any "environment.json" files sitting on servers should update these terms once they are using this version: https://github.com/CarnegieLearningWeb/UpGrade/pull/722/files#diff-c9d23ad3e1826d294dae37b7b68273675c7a2ff2aab13035bc48b1b5bcb116d5R28

gapiClientId -> googleClientId
endpointApi -> apiBaseUrl

I made it optional so it won't break though.

jreddig (approved):
Thanks for updating!

frontend/projects/upgrade/src/app/app.module.ts
```diff
@@ -11,24 +11,24 @@ import { FormsModule } from '@angular/forms';
...
   return () =>
     http
       .get('/environment.json')
       .toPromise()
-      .then((data) => {
-        env.apiBaseUrl = (data as any).endpointApi;
-        env.gapiClientId = (data as any).gapiClientId;
+      .then((config: RuntimeEnvironmentConfig) => {
+        env.apiBaseUrl = config.endpointApi || config.apiBaseUrl;
+        env.googleClientId = config.googleClientId || config.googleClientId;
```
ppratikcr7, line 31: @danoswaltCL I feel we should have optional clientId fields to handle previous naming convention. You have probably had same names. It should be:
`env.googleClientId = config. gapiClientId || config.googleClientId;`

PR #747, opened 2023-03-15
major version bump for all of the package.jsons

danoswaltCL:
We should make a practice of bumping the major (or minor) dev branch version immediately after creating a release. Otherwise the dev branch "version" remains version looks like the current release, when it actually represents the next release!

Next release will be 4.0, so I'm aligning all of the package.json versions to match, no reason to maintain those separate versions individually if we're not operating that way.

When we create the next release, we should bump `dev` to 4.1. We can always then bump it to 5.0 if we find we have major changes. As long as it out-pips the last release.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "ab_testing_backend",
-    "version": "1.0.0",
+    "version": "4.0",
     "description": "Backend for A/B Testing Project",
     "scripts": {
         "install:all": "npm ci && cd packages/Scheduler && npm ci && cd ../Upgrade && npm ci",
--- a/backend/packages/Scheduler/package.json
+++ b/backend/packages/Scheduler/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ppl-upgrade-serverless",
-  "version": "1.0.0",
+  "version": "4.0",
   "description": "Serverless webpack example using Typescript",
   "main": "handler.js",
   "scripts": {
--- a/backend/packages/Upgrade/package.json
+++ b/backend/packages/Upgrade/package.json
@@ -1,6 +1,6 @@
 {
     "name": "ab_testing_backend",
-    "version": "3.0.10",
+    "version": "4.0",
     "description": "Backend for A/B Testing Project",
     "main": "index.js",
     "scripts": {
--- a/clientlibs/js/package.json
+++ b/clientlibs/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "upgrade_client_lib",
-  "version": "1.1.7",
+  "version": "4.0",
   "description": "Client library to communicate with the Upgrade server",
   "main": "dist/bundle.js",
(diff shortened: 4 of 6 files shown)


jreddig:
Do we see a need to increment by patch versions? major.minor.patch?

danoswaltCL:
oh right right i should tack the patch version on there also to keep it clear. Yeah, it'll still be good to have the patch version for bugfixes and hotfixes. But during development the patch version would remain 0... I've seen projects increment patch version for literally any merged PR, which is interesting but I don't know the benefit of that.

Currently the Create-Release action is auto-pipping the patch version in order to create a new release tag. If we're making sure to update the version on dev branch as a part of creating a release, I guess it doesn't need to do that, although it wouldn't hurt anything.

I have to write it out, here's an imagined scenario where we're tracking supported releases + dev version

- Current:
latest release: 3.0.17
dev: 3.0.0 (development for Ardent Angle. See, it's in the past, according to versioning! Kinda confusing!)

- Current (after this is merged):
latest release: 3.0.17
dev: 4.0.0 (development for Ardent Angle)

- After create Ardent Angle release with github action:
latest release: 4.0.1 (Ardent Angle)
dev: 4.1.0 (development for Brilliant Base)

- After create bugfix for Ardent Angle release:
latest release: 4.0.2 (Ardent Angle)
dev: 4.1.0 (development for Brilliant Base)

- After like 20 more bugfixes/hotfixes
latest release: 4.0.22 (Ardent Angle)
dev: 4.1.0 (development for Brilliant Base)

- After create Brilliant Base release
latest release: 4.1.1 (Brilliant Base)
dev: 4.2.0 (development for Curious Cube)

- After breaking API changes for (Curious Cube)
Ardent Angle: 4.0.22
latest release: 4.1.1 (Brilliant Base)
dev: 5.0.0 (development for Curious Cube)

- After 5 bugfixes for Brilliant Base and 1 hotfix for Ardent Angle, supposing we are still supporting Ardent Angle for at least 1  "customer"
Ardent Angle: 4.0.23
latest release: 4.1.6
dev: 5.0.0 (Curious Cube)

- After C Release
Ardent Angle: 4.0.23
Brilliant Base: 4.1.6
Curious Cube: 5.0.1
dev: 5.1.0 (D release)

So the major.minor will indicate a published release
Legacy = < 3.x.x
Ardent Angle = 4.0.x
Brilliant Base = 4.1.x
Curious Cube = 5.0.x
Planned D Release = 5.1.x, but could become 6.0.x before publishing release if breaking API changes

Which I hope will make it easier to track and talk about things... in this schema, I think this would indicate to anyone using the client library that they should expect it to work with any major release. So you have client lib 4.0.15 Ardent Angle, you're still good with 4.1.1234 Brilliant Base, because in theory, breaking API changes happen on the majors.

danoswaltCL:
actually, pipping the patch version for any merge to dev might be pretty helpful if you're wanting to know exactly what is currently deployed to dev region.

image

jreddig (approved):

PR #748, opened 2023-03-21
Independent Factors and Decision Points, Client SDK Changes

RidhamShah:
- Factors are now generated at Experiment Level along with Conditions & Decision Points
- Factors are not related to Decision Points
- Frontend UI updated to match the same
- V2 versioning of Client Side SDK is added


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1902,6 +1896,14 @@ export class ExperimentAssignmentService {
...
-    return [factorialCondition, aliases];
+
+    const objectToReturn = {};
+    objectToReturn['factorialCondition'] = factorialCondition;
+    objectToReturn['aliases'] = aliases;
+    objectToReturn['assignedFactor'] = assignedFactor;
+    objectToReturn['conditionAlias'] = factorialConditionAlias;
+
+    return objectToReturn;
+    return [factorialCondition, aliases, assignedFactor];
```
danoswaltCL, line 1907: two return lines here, what is supposed to be returned?

clientlibs/js/src/functions/addMetrics.ts
```diff
@@ -8,25 +8,23 @@ export default async function addMetrics(
...
-    if (response.status) {
-      response.data = response.data.map((metric) => {
+  const response = await fetchDataService(
+    url,
+    token,
+    clientSessionId,
+    { metricUnit: metrics },
+    Types.REQUEST_TYPES.POST
+  );
+  if (response.status) {
```
VivekFitkariwala, line 18: Can you check the status to be in range of 200 – 299. And every where else. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

clientlibs/js/src/functions/getAllExperimentConditions.ts
```diff
@@ -9,24 +9,32 @@ export default async function getAllExperimentConditions(
...
+  const experimentConditionResponse = await fetchDataService(
+    url,
+    token,
+    clientSessionId,
+    params,
+    Types.REQUEST_TYPES.POST
+  );
+  if (experimentConditionResponse.status) {
+    experimentConditionResponse.data = experimentConditionResponse.data.map(
+      (data: { assignedCondition: { conditionCode: any; conditionAlias: any; experimentId: any } }) => {
```
VivekFitkariwala, line 25: You can use IExperimentAssignment here?

clientlibs/js/src/functions/getAllfeatureFlags.ts
```diff
@@ -7,30 +7,30 @@ export default async function getAllFeatureFlags(
   token: string,
   clientSessionId: string
 ): Promise<IFeatureFlag[]> {
-  try {
-    const featureFlagResponse = await fetchDataService(url, token, clientSessionId, {}, Types.REQUEST_TYPES.GET);
-    if (featureFlagResponse.status) {
-      return featureFlagResponse.data.map((flag) => {
+  const featureFlagResponse = await fetchDataService(url, token, clientSessionId, {}, Types.REQUEST_TYPES.GET);
```
VivekFitkariwala, line 10: avoid using any

clientlibs/js/src/functions/getExperimentCondition.ts
```diff
@@ -1,21 +1,80 @@
...
-  try {
-    if (experimentConditionData) {
-      const result = experimentConditionData.filter((data) =>
-        target ? data.target === target && data.site === site : data.site === site && !data.target
-      );
+  experimentPoint: string,
+  partitionId?: string
+): Assignment {
+  if (experimentConditionData) {
+    const result = experimentConditionData.filter((data) =>
```
VivekFitkariwala, line 61: Use find instead of filter

clientlibs/js/src/functions/getExperimentCondition.ts
```diff
@@ -1,21 +1,80 @@
...
+    this._experimentType = assignedFactor ? EXPERIMENT_TYPE.FACTORIAL : EXPERIMENT_TYPE.SIMPLE;
+    this._assignedFactor = assignedFactor;
+  }
+
+  public getCondition(): string {
+    return this._conditionCode;
+  }
+
+  public getPayload(): IPayload {
+    return this._conditionAlias ? { type: PAYLOAD_TYPE.STRING, value: this._conditionAlias } : null;
```
VivekFitkariwala, line 25: Replace alias with payload in the code

clientlibs/js/src/functions/getFeatureFlag.ts
```diff
@@ -1,23 +1,19 @@
...
-        const activeVariation = getActiveVariation(result[0]) as any;
-        return {
-          ...result[0],
-          variations: activeVariation,
-        };
-      } else {
-        throw new Error('Feature flag with given key not found');
-      }
+  if (featureFlagsData) {
+    const result = featureFlagsData.filter((data) => data.key === key);
```
VivekFitkariwala, line 5: Use find instead of filter

clientlibs/js/src/functions/init.ts
```diff
@@ -9,32 +10,36 @@ export default async function init(
...
+    id: userId,
+  };
 
-    if (group) {
-      data = {
-        ...data,
-        group,
-      };
+  if (group) {
+    if (!(group instanceof Map)) {
```
VivekFitkariwala, line 18: Now we are taking group and working group as an object instead of Map so this part of the code is incorrect. Ideally it should throw an error if you are passing object for group and working group. Remove the checking of instance of Map

clientlibs/js/src/functions/init.ts
```diff
@@ -9,32 +10,36 @@ export default async function init(
...
+      throw new Error('Group type should be Map<string, Array<string>>');
     }
+    const groupObj = convertMapToObj(group);
```
VivekFitkariwala, line 21: We no longer need this conversion

clientlibs/js/src/functions/init.ts
```diff
@@ -9,32 +10,36 @@ export default async function init(
...
+    };
+  }
 
-    if (workingGroup) {
-      data = {
-        ...data,
-        workingGroup,
-      };
+  if (workingGroup) {
+    if (!(workingGroup instanceof Map)) {
```
VivekFitkariwala, line 29: Same as above

clientlibs/js/src/functions/init.ts
```diff
@@ -9,32 +10,36 @@ export default async function init(
...
+      throw new Error('Working group type should be Map<string, string>');
     }
+    const workingGroupObj = convertMapToObj(workingGroup);
```
VivekFitkariwala, line 32: Same as above

clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -12,26 +12,28 @@ export default async function markExperimentPoint(
   status: MARKED_DECISION_POINT_STATUS,
   target?: string
 ): Promise<Interfaces.IMarkExperimentPoint> {
-  try {
-    let data: any = {
-      site,
-      condition,
-      userId,
-      status,
+  let data: any = {
```
VivekFitkariwala, line 15: Avoid using any here.

clientlibs/js/src/functions/addMetrics.ts
VivekFitkariwala: Sorry I haven't seen the fetchDataService. It is already checking that ignore the above comment.

clientlibs/js/src/functions/setGroupMembership.ts
```diff
@@ -9,20 +9,20 @@ export default async function setGroupMembership(
...
-      Types.REQUEST_TYPES.PATCH
-    );
-    if (response.status) {
-      return response.data;
-    } else {
-      throw new Error(JSON.stringify(response.message));
-    }
-  } catch (error) {
-    throw new Error(error.message);
+  if (!(group instanceof Map)) {
```
VivekFitkariwala, line 12: Remove this code we are not sending group as Map

clientlibs/js/src/functions/setGroupMembership.ts
```diff
@@ -9,20 +9,20 @@ export default async function setGroupMembership(
...
+    throw new Error('Group type should be Map<string, Array<string>>');
+  }
+  const groupObj = convertMapToObj(group);
```
VivekFitkariwala, line 15: Also delete convertMapToObject as this should not be used anymore.

clientlibs/js/src/functions/setWorkingGroup.ts
```diff
@@ -9,20 +9,20 @@ export default async function setWorkingGroup(
...
-      Types.REQUEST_TYPES.PATCH
-    );
-    if (response.status) {
-      return response.data;
-    } else {
-      throw new Error(JSON.stringify(response.message));
-    }
-  } catch (error) {
-    throw new Error(error.message);
+  if (!(workingGroup instanceof Map)) {
```
VivekFitkariwala, line 12: Update here also

VivekFitkariwala:
@RidhamShah Can you check the PR again the changes which you already did for v1 client side SDK are not getting reflected here.

clientlibs/js/src/functions/getAllExperimentConditions.ts
RidhamShah: code updated

clientlibs/js/src/functions/getExperimentCondition.ts
RidhamShah: code updated

clientlibs/js/src/functions/getExperimentCondition.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1458,29 +1466,33 @@ export class ExperimentService {
...
     const allLevels: Level[] = [];
     const allLevelCombinationElements: LevelCombinationElement[] = [];
 
-    partitions.forEach((partition) => {
-      partition.factors.forEach((factor) => {
+    // creating factors docs
+    const allFactors =
+      factors &&
+      factors.length > 0 &&
+      factors.map((factor) => {
```
danoswaltCL, line 1482: this `map` returns the factor to nowhere, but I think `forEach` is what was meant?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1458,29 +1466,33 @@ export class ExperimentService {
   }
 
   private async addFactorialDataInDB(
-    partitions: DecisionPoint[],
+    factors: any[],
```
danoswaltCL, line 1469: please avoid `any`

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1505,31 +1517,26 @@ export class ExperimentService {
...
+      error.message = `Error in creating factors & levels`;
       logger.error(error);
       throw error;
     }
 
     return [factorDoc, levelDoc, levelCombinationElementsDoc];
   }
 
-  private formatingFactorAndLevels(partitions: any[], factors: any[], levels: any[]): any[] {
+  private formatingFactorAndLevels(factors: any[], levels: any[]): any[] {
```
danoswaltCL, line 1528: please avoid `any`

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -23,44 +23,51 @@ import * as uuid from 'uuid';
...
-    getAllExperimentConditions: null,
-    markExperimentPoint: null,
-    setGroupMemberShip: null,
-    setWorkingGroup: null,
-    failedExperimentPoint: null,
-    getAllFeatureFlag: null,
-    log: null,
-    altUserIds: null,
-    addMetrics: null,
+    init: null as string,
```
danoswaltCL, line 26: `null as string` is a bug waiting to happen, we can't be doing that, it makes me nervous that it's a band-aid for something somewhere else, but I think these can simply be empty strings instead of `null` right?

clientlibs/js/src/functions/addMetrics.ts
```diff
@@ -8,25 +8,23 @@ export default async function addMetrics(
...
+    response.data = response.data.map(
+      (metric: { [x: string]: any; createdAt: any; updatedAt: any; versionNumber: any }) => {
```
danoswaltCL, line 20: please avoid `any` and variable names like `x`.

clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -12,26 +20,28 @@ export default async function markExperimentPoint(
...
-    } else {
-      throw new Error(JSON.stringify(response.message));
-    }
-  } catch (error) {
-    throw new Error(error.message);
+  }
+  const response = await fetchDataService(url, token, clientSessionId, data, Types.REQUEST_TYPES.POST);
+  if (response.status) {
+    return response.data;
+    // return {
```
danoswaltCL, line 38: can we just delete commented out "old" code everywhere? If we want to revisit old code, it's always there in git history

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-query-result/experiment-query-result.component.ts
```diff
@@ -46,14 +46,13 @@ export class ExperimentQueryResultComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     const queryIds = [];
-
+    // sort the factors:
+    this.experiment.factors = this.experiment.factors.slice().sort((a, b) => (a.order > b.order ? 1 : b.order > a.order ? -1 : 0));
```
danoswaltCL, line 50: can this sort method please be pulled out into a class helper method? It's implemented twice in the same function. Something like

```typescript
sortFactorsByOrderAscending(factorA: ExperimentFactor, factorB: ExperimentFactor): number {
  factorA.order > factorB.order ? 1 : factorB.order > factorA.order ? -1 : 0)
}

so
```

danoswaltCL:
hey, I know it's a work in progress, wanted to note that in local when I pulled this it seems to error in the design step if I don't click on the conditions button before trying to proceed to next step.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-query-result/experiment-query-result.component.ts
danoswaltCL: if `this.experiment.factors` is not an array, this would break; will `this.experiment.factors` still exist as an array if this component is loaded as a simple experiment?

ppratikcr7:
> hey, I know it's a work in progress, wanted to note that in local when I pulled this it seems to error in the design step if I don't click on the conditions button before trying to proceed to next step.

Yes, we are working on row-wise edit feature which will save the updated condition table on unlocking rows. For now, I can hit the update conditions table function on next button.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-query-result/experiment-query-result.component.ts
ppratikcr7: > can this sort method please be pulled out into a class helper method? It's implemented more than once. Something like
> 
> ```ts
> sortFactorsByOrderAscending(factorA: ExperimentFactor, factorB: ExperimentFactor): number {
>   factorA.order > factorB.order ? 1 : factorB.order > factorA.order ? -1 : 0)
> }
> ```

Sure, I will do that.

ppratikcr7: > if `this.experiment.factors` is not an array, this would break; will `this.experiment.factors` still exist as an array if this component is loaded as a simple experiment?

It should not break, I will move it inside the check when the experiment is factorial.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-query-result/experiment-query-result.component.ts
```diff
@@ -261,13 +250,11 @@ export class ExperimentQueryResultComponent implements OnInit, OnDestroy {
...
     let levelName;
-    this.experiment.partitions.forEach((decisionPoint) => {
-      decisionPoint.factors.forEach((factor) => {
-        factor.levels.forEach((level) => {
-          if (level.id === levelId) {
-            levelName = level.name;
-          }
-        });
+    this.experiment.factors.forEach((factor) => {
+      factor.levels.forEach((level) => {
```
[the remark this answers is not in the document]
ppratikcr7: > this is a minor thing but I think a similar searching pattern like this comes up a lot and should be a hint that we're doing too much work: One thing is that `forEach` isn't the right tool for the job, as it will continue iterating after it finds a match, so`find` might be best.
> 
> But it would be better to take a step back altogether, because this would do all of these iterations every time `getLevelName()` is invoked also. Ultimately it's not a huge deal as there won't be 1000 factors and levels to sift through, but if you cache a flat reference map of levels at startup (or in ngrx store as soon as the experiment is upserted...) then it's just a straight lookup each time, which would be ideal:
> 
> for example (this is untested code for illustration purposes, use at your own risk!)
> 
> ```ts
> /**
>  * What we want is a flat map of "id: Level":
>  * { 
>  *   'abc1': Level,
>  *   'd6f3': Level,
>  *   'a2b5': Level
>  * }
>  */
> type LevelsMap = { [key: string] : ExperimentLevel };
> 
> this.levelsMap: LevelsMap = {};
> 
> ngOnInit() {
>   // ...do things
>   this.levels = this.createLevelsMap(this.experiment.factors); // make a flat lookup map one time
>   // ...do more things
> }
> 
> createLevelsMap(factors: ExperimentFactor[]): LevelsMap {
>   return factors.reduce((levelsMap, factor) => {
>       factor.levels.forEach((level) => {
>          levelsMap[level.id] = level
>       })
>       
>       return levelsMap;
>   }, {})
> }
> 
> // then this accessor is simple:
> getLevelName(levelId: string): string {
>   const level = this.levelsMap[levelId];
>   
>   return level?.name || "";
> }
> ```

Okay, I will look into this. Thanks :)

clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -12,26 +20,22 @@ export default async function markExperimentPoint(
   status: MARKED_DECISION_POINT_STATUS,
```
VivekFitkariwala, line 20: condition is no longer required as we are sending status.
condition is used on the server to check if client has applied the assignment correctly but now we are also sending status so we should remove condition. @danoswaltCL @jerith666

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -3,14 +3,14 @@ import { Interfaces } from './identifiers';
 import setWorkingGroup from './functions/setWorkingGroup';
 import getAllExperimentConditions from './functions/getAllExperimentConditions';
 import {
-  IExperimentAssignment,
   IFeatureFlag,
   ISingleMetric,
   IGroupMetric,
   ILogInput,
   MARKED_DECISION_POINT_STATUS,
+  IExperimentAssignment2,
```
VivekFitkariwala, line 11: Can we have proper version number in the type name ie v4

backend/packages/Upgrade/src/api/services/ExperimentService.ts
ppratikcr7: Okay, you are right. Thanks

clientlibs/js/src/functions/markExperimentPoint.ts
RidhamShah: code removed

backend/packages/Upgrade/src/api/services/ExperimentService.ts
RidhamShah: We need a map so that the value from different factors combined to allFactors and also using forEach sometimes throw some null error in further code.

clientlibs/js/src/UpgradeClient.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.v4.ts
```diff
@@ -0,0 +1,27 @@
...
+    assignedFactor: object | undefined;
+  };
+
+  @IsNotEmpty()
+  @IsDefined()
+  public site: string;
+
+  @IsNotEmpty()
+  @IsDefined()
+  public assignedCondition: { conditionCode: string; experimentId: string };
```
jreddig, line 22: Similarly here, I believe we only use `data.assignedCondition` and `data.assignedFactor` so can we remove these two?

jreddig:
Updated the Java libs:
- experiment type included in assign response
- created mark data object - site/target/condition/factors contained in one object (mark now takes status, id, and data object)
- condition object includes payload object
- response from mark no longer includes status/code
- assign response has condition object and map of factors
- getFactorLevel and getFactorPayload

What else am I missing?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v4.ts
```diff
@@ -0,0 +1,937 @@
...
+   *          '500':
+   *            description: User not defined
+   */
+  @Post('mark')
+  public async markExperimentPoint(
+    @Body({ validate: { validationError: { target: false, value: false } } })
+    @Req()
+    request: AppRequest,
+    experiment: MarkExperimentValidator2
+  ): Promise<IMonitoredDeciosionPoint> {
```
ppratikcr7, line 428: Change spelling error `IMonitoredDeciosionPoint` to `IMonitoredDecisionPoint`

RidhamShah: Code updated

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.v4.ts
RidhamShah: Same as above

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.v4.ts
```diff
@@ -0,0 +1,27 @@
...
+  public data: {
+    site: string;
+    target: string | undefined;
+    assignedCondition: { conditionCode: string; experimentId: string };
+    assignedFactor: object | undefined;
+  };
+
+  @IsNotEmpty()
+  @IsDefined()
+  public site: string;
```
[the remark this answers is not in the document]
RidhamShah: Ya, I forget to remove this. The code is updated now.

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -35,13 +35,7 @@ import { env } from '../../env';
...
   expId: string;
   expPoint: string;
-  twoCharacterId: string;
-  description: string;
-  assignedCondition: {
-    conditionCode: string;
-    twoCharacterId: string;
-    description: string;
-  };
+  assignedCondition: object;
```
danoswaltCL, line 38: we should really avoid `object`, it is almost as meaningless as `any` or `unknown`, it will give unexpected warnings that aren't helpful.

If we really have places that we could get an object that could be anything, I see `Record<string, unknown>` as generally recommended best alternative (I have also used `{ [key: string]: unknown }` in places.

But for this, we should be able to give `assignedCondition` an explicit type.

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v4.ts
```diff
@@ -0,0 +1,938 @@
...
+import { AppRequest } from '../../types';
+import { env } from '../../env';
+import { MonitoredDecisionPointLog } from '../models/MonitoredDecisionPointLog';
+import { MarkExperimentValidatorv4 } from './validators/MarkExperimentValidator.v4';
+
+interface ILog {
+  id: string;
+  uniquifier: string;
+  timeStamp: Date;
+  data: any;
```
danoswaltCL, line 50: If this literally could be any unknown object, we should use `Record<string, unknown>` so that we still get typescript warnings about unknowable properties on this object potentially being undefined if we try to access it in code.

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v4.ts
```diff
@@ -0,0 +1,938 @@
...
+import { MetricService } from '../services/MetricService';
+import { ExperimentUserAliasesValidator } from './validators/ExperimentUserAliasesValidator';
+import { Metric } from '../models/Metric';
+import * as express from 'express';
+import { AppRequest } from '../../types';
+import { env } from '../../env';
+import { MonitoredDecisionPointLog } from '../models/MonitoredDecisionPointLog';
+import { MarkExperimentValidatorv4 } from './validators/MarkExperimentValidator.v4';
+
+interface ILog {
```
danoswaltCL, line 46: Is there any reason why the 3 versioned ExperimentClientController files have interfaces defined in the controller file instead of in `Types`? I can understand if they were private and local just to this file, but these look like they are effectively the public types for response DTOs of these endpoints. Unless I'm missing something, that's important type information that is getting buried because it's being treated as a private, local type when they could be a public, global type.

Can these be moved to types files so that they organized and exported like the other response types so that it is documentable and also not duplicated across files?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v4.ts
```diff
@@ -0,0 +1,938 @@
...
+   *         - Client Side SDK
+   *       produces:
+   *         - application/json
+   *       responses:
+   *          '200':
+   *            description: Client side reported error
+   *          '500':
+   *            description: null value in column "id\" of relation \"experiment_user\" violates not-null constraint
+   */
+  @Post('failed')
```
danoswaltCL, line 744: do we need this still in v4? I thought this was removed

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.v4.ts
```diff
@@ -0,0 +1,19 @@
...
+export class MarkExperimentValidatorv4 {
+  @IsNotEmpty()
+  @IsDefined()
+  public userId: string;
+
+  public data: {
+    site: string;
+    target: string | undefined;
+    assignedCondition: { conditionCode: string; experimentId: string };
+    assignedFactor: object | undefined;
```
danoswaltCL, line 13: as noted in another comment, `object` is not a great type, can we give this a real type for `assignedFactor`?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
danoswaltCL: my bad, I misread the code 👍

RidhamShah:
> What's the expected behavior when the client includes an error string? If the client responds with CONDITION_FAILED_TO_APPLY, should we still save that information? I'm not sure that returning 500 Server Error when the client reports an error on their side is appropriate. Maybe we could return 204 to say that their error was successfully received and logged?

@jreddig There is a small change that we missed, Now clientError won't throw an error but only logs the error in logger. Further code will remain the same and so mark will return normal 200 response

zackcl:
I just tested this branch and identified three issues:
1. The confirm icon, while in edit mode, does not function unless changes are made to the row.
2. The condition payload, as opposed to the level payload, is not displayed on the details page.
3. The Conditions table reverts to its default settings (no payloads and equal weights) when the "Save" button is clicked twice within the experiment edit modal.

https://user-images.githubusercontent.com/90279765/230178567-59e503ad-e41a-4e42-aa30-611d414d04db.mov

Although issues 1 and 3 may be tolerable for now, as they do not hinder the execution of experiments, issue 2 should be addressed to ensure that payloads can be inspected on the details page.

jreddig:
Thanks @RidhamShah! The Java client lib changes are complete now

zackcl:
@ppratikcr7 Thank you! I tested the branch once more.

- Issue 1: I can confirm that it has been definitively resolved.
- Issue 2: I encountered a situation where the condition payloads were not displayed on the details page once, but I have been unable to replicate this issue. It appears to be working most of the time, so it's possible I made a mistake.
- Issue 3: While I can confirm that this issue is fixed, it seems that updating weights is not functioning as intended. Please refer to the attached video:

https://user-images.githubusercontent.com/90279765/230662897-6abf6219-7313-4217-b055-79e2412ccf13.mov

VivekFitkariwala (approved):

ppratikcr7:
> @ppratikcr7 Thank you! I tested the branch once more.
> 
> * Issue 1: I can confirm that it has been definitively resolved.
> * Issue 2: I encountered a situation where the condition payloads were not displayed on the details page once, but I have been unable to replicate this issue. It appears to be working most of the time, so it's possible I made a mistake.
> * Issue 3: While I can confirm that this issue is fixed, it seems that updating weights is not functioning as intended. Please refer to the attached video:

Hi Zack. Thanks for confirming. The issue you are referring in updating weights, seems to be correct functioning. If we exclude a condition, then its weight is set to 0 and rest conditions are assigned weights equally. Is there anything else you want to refer to?

zackcl:
@ppratikcr7 What I meant was that updating the weights in the modal is not applied to the details page, which means it's not possible to change the weights. Please watch the video from `1:32`.

ppratikcr7:
> @ppratikcr7 What I meant was that updating the weights in the modal is not applied to the details page, which means it's not possible to change the weights. Please watch the video from `1:32`. Please let me know if I did something wrong.

Okay. Got it. I will check on this. All condition table bugs would be pre-existing as we haven't updated the component in this PR.

ppratikcr7:
Sorry. Somehow, while typing the previous reply, my trackpad clicked the close PR button. Reopening the PR.

Issue #750, opened 2023-03-23
Generate Client Library Docs for Gitbook with TSDocs/JavaDocs


danoswaltCL:
**Is your feature request related to a problem? Please describe.**
Let's just write documentation once in the codebase and import it into Gitbook for client libraries!

**Describe the solution you'd like**
TSDocs looks like a good solution for JS library, maybe JavaDocs can be made to do a similar thing.

PR #751, opened 2023-03-23
[No merge! Just showing example typedocs integration!]

danoswaltCL:
#750

Here is a very bare-bones example of what we can do with TSDocs integration.

This would install a CLI package called TypeDoc (not to be confused with TSDoc, which is just a syntax standard!) that will read the code and output docs in different flavors.

If you want to try out this branch, pull and do `npm install` because it installs a couple things.

if you run `npm run docs:markdown`, in `clientlibs/js` directory, you will get docs generated in a new output folder called `docs`.

image

Then the resulting markdown file can be imported into gitbook. Here is test page of what this looks like without doing any TSDoc annotations at all, it just knows quite a bit right out of the box because it can parse the TypeScript classes.

https://upgrade-platform.gitbook.io/docs/v/4.0/developer-guide/reference/client-libraries-testing/javascript-sdk

We can configure it to output what we like, but pretty cooool


VivekFitkariwala:
Thanks @danoswaltCL it looks good. I will pull the branch and try it.

PR #759, opened 2023-03-31
support for caliper graded profile

jreddig:
Added an endpoint & client lib function that accepts graded events in caliper format and converts to our metric format

Envelope:
```
{
  "sensor": "https://example.edu/sensors/1",
  "sendTime": "2023-02-08T11:05:01.000Z",
  "dataVersion": "http://purl.imsglobal.org/ctx/caliper/v1p2",
  "data": [ {event1}, {event2}, {eventN} ]
}
```

Event Data:
```
{
  "@context": "http://purl.imsglobal.org/ctx/caliper/v1p2",
  "id": "urn:uuid:a50ca17f-5971-47bb-8fca-4e6e6879001d",
  "type": "GradeEvent",
  "profile": "GradingProfile",
  "actor": {
    "id": "https://example.edu/autograder",
    "type": "SoftwareApplication",
    "version": "v2"
  },
  "action": "Graded",
  "object": {
    "id": "https://example.edu/terms/201601/courses/7/sections/1/assess/1/users/554433/attempts/1",
    "type": "Attempt",
    "assignee": {
      "id": "https://example.edu/users/554433",
      "type": "Person"
    },
    "assignable": {
      "id": "https://example.edu/terms/201601/courses/7/sections/1/assess/1",
      "type": "Assessment"
    },
    "count": 1,
    "dateCreated": "2016-11-15T10:05:00.000Z",
    "startedAtTime": "2016-11-15T10:05:00.000Z",
    "endedAtTime": "2016-11-15T10:55:12.000Z",
    "duration": "PT50M12S"
  },
  "eventTime": "2016-11-15T10:57:06.000Z",
  "edApp": "https://example.edu",
  "generated": {
    "id": "https://example.edu/terms/201601/courses/7/sections/1/assess/1/users/554433/attempts/1/scores/1",
    "type": "Score",
    "attempt": "https://example.edu/terms/201601/courses/7/sections/1/assess/1/users/554433/attempts/1",
    "maxScore": 15.0,
    "scoreGiven": 10.0,
    "scoredBy": "https://example.edu/autograder",
    "comment": "auto-graded exam",
    "dateCreated": "2016-11-15T10:56:00.000Z"
  },
  "group": {
    "id": "https://example.edu/terms/201601/courses/7/sections/1",
    "type": "CourseSection",
    "courseNumber": "CPS 435-01",
    "academicSession": "Fall 2016"
  }
}
```

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/validators/CaliperLogData.ts
+++ b/backend/packages/Upgrade/src/api/controllers/validators/CaliperLogData.ts
@@ -1,0 +1,39 @@
+import { IsNotEmpty, IsDefined, IsString, IsJSON, IsObject } from 'class-validator';
+import { Attempt, CaliperActor, ScoreObject } from '../../../../../../../types/src/Experiment/interfaces';
+
+export class CaliperLogData {
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsString()
+  public profile: string;
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsJSON()
+  public actor: CaliperActor;
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsString()
+  public action: string;
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsString()
+  public eventTime: string;
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsJSON()
+  public object: Attempt;
+
+  @IsObject()
+  public extensions: object;
+
+  @IsObject()
+  @IsNotEmpty()
+  @IsJSON()
+  public generated: ScoreObject;
+
+}
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -17,7 +17,8 @@ import { ExperimentUser } from '../models/ExperimentUser';
 import { ExperimentUserService } from '../services/ExperimentUserService';
 import { UpdateWorkingGroupValidator } from './validators/UpdateWorkingGroupValidator';
 import { MonitoredDecisionPoint } from '../models/MonitoredDecisionPoint';
-import { ISingleMetric, IGroupMetric, SERVER_ERROR } from 'upgrade_types';
+import { ISingleMetric, IGroupMetric, SERVER_ERROR, ILogInput } from 'upgrade_types';
+import { parse, toSeconds } from 'iso8601-duration';
 import { FailedParamsValidator } from './validators/FailedParamsValidator';
 import { ExperimentError } from '../models/ExperimentError';
 import { FeatureFlag } from '../models/FeatureFlag';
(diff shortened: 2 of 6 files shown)


backend/packages/Upgrade/src/api/controllers/validators/CaliperLogData.ts
```diff
@@ -0,0 +1,39 @@
...
+  @IsString()
+  public eventTime: string;
+
+  @IsDefined()
+  @IsNotEmpty()
+  @IsJSON()
+  public object: Attempt;
+
+  @IsObject()
+  public extensions: object;
```
danoswaltCL, line 32: `object` and `Object` Typescript types are less useful than they appear, as it will allow anything that is an object at all (arrays are objects!) and if you were to do write `extensions.something` it will complain that `something` is an unknown property of `extensions`.

There are a couple of alternatives that I see floated around to avoid `object` type where all you really know is that you could get an object with random stuff, which is either:

`{ [key: string] : any }` or `Record<string, any>`

I also see `Record<string, unknown>`, but I'm not 100% the benefit of unknown vs any. Not sure how you'd get around any or unknown though

danoswaltCL: I think `Record<string, unknown>` is what we want if we are just throwing out this data and/or not doing any processing of those vars to try to access data on them or transform before writing to db.

Based on my thorough stack overflow skimming, `any` is like turning off type-checking entirely, where `unknown` means allow anything, but yell at me if I actually try to use it for anything.

danoswaltCL: Or maybe the type can be `ILogInput | Record<string, unknown>`?

danoswaltCL: Or maybe not; I think you'd just want `Record<string, unknown>` if we don't want to force the format... I think the use case would be to flexibly allow any unknown object, but yell at anyone who tries to access properties on `extensions` to remind them that nothing is guaranteed about these objects.

jreddig: I like `Record<string, unknown>` to just pass it on and save the metrics. If I do the same on the client lib side, does unknown make it hard to create the object?

danoswaltCL: good question... I guess it could. I think it's probably ok to say `Record<string, any>` if we're going to allow any objects, as they'd run into annoying issues if they were creating the object in steps:

```typescript
const extensions: Record<string, unknown> = {};

extensions.prop.nestedProp = 4; // it would complain about nested prop being unknown
```

jerith666:
Note that we've started using https://github.com/ImagineLearning/caliper-ts in some internal projects; worth a look here?

jreddig:
Thanks @jerith666 ! Maybe I can remove support for the endpoint in the client libs since this exists? And caliper producers can use the upgrade endpoint as a sensor client?

jerith666:
That seems ideal if it can be made to work, yeah!  I'm not sure how you'd authenticate those requests though.

jreddig:
Talked with Kyle Lauffer (RSM) about the Caliper library he's building with caliper-ts for CL. Will need to file a ticket to add support for the GradingProfile, then I can test that this UpGrade API will work with our applications

jreddig:
Resolved conflicts from factorial branch. Adding the GradingProfile to the CL caliper library will take some time, so I'm keeping the client library functions in this branch. This is ready to go!

PR #768, opened 2023-04-12
Edit feature factorial table - new

ppratikcr7:
This Draft PR is to add edit feature in factor and level nested tables. This PR also aims at renaming and refactoring frontend code for stepper design component for edit feature of various tables as they were introduced by various developers with various naming conventions and had placed code snippets at scattered locations in the code file. That made it difficult to read and understand. I have organized them in particular uniform fashion and naming conventions to better readability and understanding of the code. 

There was a bug in decision point count validation, the stepper allowed to move forward without any decision point. I have handled that as well. 

I have removed some unwanted commented code snippets as a part of code cleanup.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "ab_testing_backend",
-    "version": "4.0",
+    "version": "4.1.0",
     "description": "Backend for A/B Testing Project",
     "scripts": {
         "install:all": "npm ci && cd packages/Scheduler && npm ci && cd ../Upgrade && npm ci",
--- a/backend/packages/Scheduler/package.json
+++ b/backend/packages/Scheduler/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ppl-upgrade-serverless",
-  "version": "4.0",
+  "version": "4.1.0",
   "description": "Serverless webpack example using Typescript",
   "main": "handler.js",
   "scripts": {
--- a/backend/packages/Upgrade/package-lock.json
+++ b/backend/packages/Upgrade/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "ab_testing_backend",
-    "version": "4.0.0",
+    "version": "4.0.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "ab_testing_backend",
-            "version": "4.0.0",
+            "version": "4.0.1",
             "license": "ISC",
             "dependencies": {
                 "@golevelup/ts-jest": "^0.3.2",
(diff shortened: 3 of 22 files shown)


danoswaltCL:
hey I see this is draft status, so maybe this has already been addressed.

1) if I create a Factor and click edit button, I can click on the arrow and open up Level and start editing the level ... this allows the edit mode to get into a funky state where it stops working correctly in other ways.

in edit mode pattern, everything except that row should be locked until the row is valid or the changes are discarded. I can see that this is a little tricky for the Factors row because of the nested Level rows. I'd expect that the left dropdown carat is locked (probably also opacity lightened? @zackcl?) while Factor row itself is in edit mode and the levels are collapsed. If the levels are expanded, I would expect that the dropdown carat and level rows are locked (and in opacity lightened). If Factor name is invalid, one should not be allowed to exit edit mode until it is valid / discarded / or the factor is deleted.

2) when I hit "close" while in edit mode and return to the stepper, the design form is locked up and uneditable, so maybe this way of exiting the stepper is not firing "reset" action.

Thanks for cleaning up / organizing @ppratikcr7, just scanning the code it looked good to me. I'll give code a closer look when it's "official".

ppratikcr7:
@danoswaltCL @zackcl @jreddig You ca review the edit feature changes here and once approved, let's not merge this PR as its already been included in the alias to payload [PR](https://github.com/CarnegieLearningWeb/UpGrade/pull/774).

zackcl:
@ppratikcr7 Thanks, I just tested this branch and found the 2 issues:

1. When I first create a factor row, the row becomes disabled which is confusing. I think this happens because a level row is added automatically after a factor row is created (which I think is nice). But in this case, the factor row should be expanded automatically so that the user can know the level row is added. Also, it would be nice if the `Level` input field is focused automatically as well.
2. When editing an existing row in the Factors table, clicking on the "X" icon should simply cancel row editing instead of deleting the row. But it currently deletes the row. It should work just like how Decision Points table works.

https://user-images.githubusercontent.com/90279765/232155960-0ec9a909-7d18-484d-bc8c-a8fa63539ee8.mov

ppratikcr7:
> @ppratikcr7 Thanks, I just tested this branch and found the 2 issues:
> 
> 1. When I first create a factor row, the row becomes disabled which is confusing. I think this happens because a level row is added automatically after a factor row is created (which I think is nice). But in this case, the factor row should be expanded automatically so that the user can know the level row is added. Also, it would be nice if the `Level` input field is focused automatically as well.
> 2. When editing an existing row in the Factors table, clicking on the "X" icon should simply cancel row editing instead of deleting the row. But it currently deletes the row. It should work just like how Decision Points table works.

Hi @zackcl Okay. Yes, level row is added automatically so factor row becomes disabled. I will expand the factor row on locking the factor row. And focus the level input. For the second point, you are correct, I will replace that part to revert changes and not delete row on cancel button click.

ppratikcr7:
> @ppratikcr7 Thanks, I just tested this branch and found the 2 issues:
> 
> 1. When I first create a factor row, the row becomes disabled which is confusing. I think this happens because a level row is added automatically after a factor row is created (which I think is nice). But in this case, the factor row should be expanded automatically so that the user can know the level row is added. Also, it would be nice if the `Level` input field is focused automatically as well.
> 2. When editing an existing row in the Factors table, clicking on the "X" icon should simply cancel row editing instead of deleting the row. But it currently deletes the row. It should work just like how Decision Points table works.
> 
>  Factors.Table.Bugs.mov

@zackcl I have handled the enhancements and bug you have reported in this comment, can you please verify once.

zackcl:
@ppratikcr7 Thank you! The level input field is not focused, but I think this is fine. The "X" icon/button now works as expected. I think it is OK to be merged. 

@danoswaltCL Do you want to quickly review the code?

ppratikcr7:
> @ppratikcr7 Thank you! The level input field is not focused, but I think this is fine. The "X" icon/button now works as expected. I think it is OK to be merged.

@zackcl Focus as in? You want level name input to be active in typing mode? I checked, but we dont use any such focus tying ready input anywhere in design page. Also, we don't need to merge this now. Its already been pulled into the "Alias to Payload" PR that just got merged. We can close this PR now. It was only for review purpose.

PR #776, opened 2023-04-18
bump to Brilliant Base release version (4.2)

danoswaltCL:
Bumping dev to "next release" version.


danoswaltCL:
ok yeah, I guess we do need to wait on those if we're planning merging those into dev and re-releasing Ardent Angle. That is a bit of why I was thinking about release branches. If we had a `releases/v4.1` branch available on remote, those PRs could be reopened against that branch so `dev` PRs can continue without waiting.

I think we'd then be able to merge bugfixes into the release branch, and re-run "create release" button from that branch (we'd have to make sure it doesn't auto-delete the release branch though!).

Something to think about; if we can get Mike's attention via the publishing ticket I opened we can make sure that all jibes with the workflows he designed. Something feels slightly backwards about introducing long-running release branches, but if we're essentially doing "git-flow" for everything else, we should probably go all the way. We're nowhere close to "continuous deployment" for this project.

danoswaltCL:
@jreddig @zackcl we're good to go with this now, can I get an approve?

jreddig (approved):
Lets go!

PR #777, opened 2023-04-18
Enhancement/creating condition table dynamically

Yagnik56:
This PR is created to make condition table dynamically with no limitation over number of factors in factor table.


frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -294,45 +302,76 @@ export class ExperimentDesignStepperService {
...
+    conditionData.forEach((level) => {
+      levels.push({ id: level.id, name: level.name });
+    });
+    return levels;
+  }
+
+  createConditionString(conditionData: FactorLevelData[]) {
+    let alias: string;
+    let count = 0;
+    conditionData.forEach((level) => {
```
danoswaltCL, line 364: Couple nitpicky things, ignore me if you like. fyi `Array.forEach` comes with an index argument: `conditionData.forEach((level, index)`, you don't have to track it manually with `count`.

Also I don't know if there's a possibility if this, but I notice that if `conditionData` is empty somehow, this function will return `undefined` for alias, which I doubt is a happy result. Maybe just initialize as empty string?

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -294,45 +302,76 @@ export class ExperimentDesignStepperService {
...
-      });
+      const tableRow: FactorialConditionTableRowData = {
+        id: uuidv4(), // TODO: maybe not the right place?
+        levels: conditionLevelsData,
+        condition: conditions,
+        alias: '',
+        weight: '0.0',
+        include: true,
+      };
+      console.log("tableRow.condition.split(';').length");
```
danoswaltCL, line 319: oops console logs...

Yagnik56: Oops sorry missed that, it's removed now

frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
Yagnik56: resolved both

jreddig (approved):

PR #782, opened 2023-04-18
Update Factors table validator

zackcl:
This fixes the issues https://github.com/CarnegieLearningWeb/UpGrade/issues/779 and https://github.com/CarnegieLearningWeb/UpGrade/issues/780.

https://user-images.githubusercontent.com/90279765/232914347-4de13fab-2a70-4c05-a3bf-891fc7d792d3.mov

I tried to make the validation messages consistent with the Conditions table in simple experiments, but perhaps they could all be improved in the future.

Screen Shot 2023-04-18 at 6 01 03 PM

PR #787, opened 2023-04-19
Update Factors table validator for edit mode

### LINKED ISSUE #779: Disallow factors with identical names in the Factors table
**Describe the bug**
Currently, it is possible to create factors with identical names in the Factors table. This could cause issues when using the client method `getFactorLevel(factor)`. As a result, we should not allow the creation of factors with the same names in the Factors table UI.

**To Reproduce**
1. Go to Design Step (Factorial Experiment)
2. Create 2 factors with the same name
3. You will see that the experiment can be created

**Expected behavior**
We should display an error message below the table when two factors with the same name are created and prevent further progress.

**Screenshots**
Screen Shot 2023-04-18 at 2 29 44 PM

### LINKED ISSUE #780: Update validation error message for Factors table when duplicate levels exist
**Describe the bug**
Currently, the validation error message for Factors table when duplicate levels exist is not properly written. It's currently shown as `{Factor} factor's {Level}undefined` which doesn't make sense. I think it should be updated to `Level should be unique` (This is how we show the message in the Conditions table in a simple experiment).

**To Reproduce**
1. Go to Design step in a factorial experiment
2. Create a factor and add 2 levels with the same name
3. You will see the weird error message

**Expected behavior**
The validation message should be `Level should be unique`

**Screenshots**
Screen Shot 2023-04-18 at 3 29 31 PM

zackcl:
Closes #779, closes #780. 
Reworked the previous/closed PR (https://github.com/CarnegieLearningWeb/UpGrade/pull/782) for the validator to work with edit mode added to the Factors table.

https://user-images.githubusercontent.com/90279765/233191256-397a2db1-7cd8-42e8-a478-8c5efb35d025.mov

I tried to copy how the Decision Points table handles the row validation on edit mode. Please review before approval.


frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
```diff
@@ -473,22 +471,30 @@ export class FactorialExperimentDesignComponent implements OnInit, OnChanges, On
...
+    return !this.factorPointErrors.length;
+  }
 
-    // Factor Points error messages
-    if (duplicateFactors.length === 1) {
-      this.factorPointErrors.push(duplicateFactors[0] + this.factorErrorMessages[0]);
-    } else if (duplicateFactors.length > 1) {
-      this.factorPointErrors.push(duplicateFactors.join(', ') + this.factorErrorMessages[1]);
+  isLevelTableRowValid(levelRowIndex: number, factorIndex: number): boolean {
+    const levels = this.factorialExperimentDesignForm.value.factors[factorIndex].levels;
```
danoswaltCL, line 488: this is fine of course, just wanted to point out a convenience function on this component, since this is :

`this.getLevels(factorIndex)` can be used instead of `this.factorialExperimentDesignForm.value.factors[factorIndex].levels` if you so please

frontend/projects/upgrade/src/app/features/dashboard/home/components/factorial-experiment-design/factorial-experiment-design.component.ts
```diff
@@ -473,22 +471,30 @@ export class FactorialExperimentDesignComponent implements OnInit, OnChanges, On
     }
   }
 
-  validateFactorNames(factors: ExperimentFactorFormData[]) {
+  isFactorTableRowValid(factorRowIndex: number): boolean {
+    const factors = this.factorialExperimentDesignForm.value.factors;
```
danoswaltCL, line 475: there's a convenience getter function that already holds a reference to the factors that you could use. i.e. `this.factors` should be equivalent to `const factors = this.factorialExperimentDesignForm.value.factors;`

zackcl:
@danoswaltCL Thanks, I've updated them using `this.factor.value` as you suggested (I couldn't find `this.getLevels(factorIndex)`).

I'm not sure why the testing fails. It seems to be not related to this PR.

ppratikcr7:
> @danoswaltCL Thanks, I've updated them using `this.factor.value` as you suggested (I couldn't find `this.getLevels(factorIndex)`).
> 
> I'm not sure why the testing fails though. It seems to be not related to this PR.

@zackcl `this.getLevels(factorIndex)` has been removed as it was redundant. We now have `this.getFactorialLevelsAt(factorIndex)`. You can use that. Also, the test cases are failing for a missing dependency in java clientlib. We didn't saw that in any of the previous PR. Maybe @jreddig can help us out here. :)

zackcl:
@ppratikcr7 Thanks, I've updated it as you suggested.

danoswaltCL (approved):

Issue #620, opened 2022-11-10
Warning message for import during version mismatch could be more helpful


amurphy-cl:
**Is your feature request related to a problem? Please describe.**
When importing an experiment that was created in a different version of UpGrade, a warning message appears about the mismatch and that "some parts may not function as intended". This is a good start but it would be more helpful to eventually have a more informative warning message here (at minimum if the imported file is older or newer than the version of UpGrade it's being imported to). 

<img width="502" alt="image" src="https://user-images.githubusercontent.com/65311369/201176602-19a4ab8e-03c2-427c-959b-ded0a337a4fa.png">

PR #813, opened 2023-04-28
Bugfix/706: Handling old experiment json formats for importing experiments and import feature enhancement

ppratikcr7:
In this PR:

1. We have managed to import mostly all formats of old experiment json.
2. I have also disabled the import button until a file is selected.
3. I have added vertical scroll to show all errors/warning with fixed height import modal.
4. Added warnings in validation message and table header.

Note: In case someone wants to verify various old experiment JSON formats mentioned in the issue and few extra one's, I have attached a zip of experiments.
[experiments.zip](https://github.com/CarnegieLearningWeb/UpGrade/files/11351544/experiments.zip)


danoswaltCL:
FYI ^ This automerge job is "failing" because currently only the `dev` branch technically requires approvals, so it's confused. I have opened a devops tickets to protect "release" branches, which would probably fix this. However I think it will not prevent us from merging.

jreddig:
I was able to import a bunch of older experiments in my local! However, for the warning message, I didn't get a list of missing properties, just the old version error message. How do we want to handle the messaging for older experiments @amurphy-cl ?

ppratikcr7:
> I was able to import a bunch of older experiments in my local! However, for the warning message, I didn't get a list of missing properties, just the old version error message. How do we want to handle the messaging for older experiments @amurphy-cl ?

For missing properties, most of them would have been modified and picked from the necessary old format fields or either filled with default values. Can you any specific missing property which you are looking for in the error messages?

ppratikcr7:
> FYI ^ This automerge job is "failing" because currently only the `dev` branch technically requires approvals, so it's confused. I have opened a devops tickets to protect "release" branches, which would probably fix this. However I think it will not prevent us from merging.

Correct, it shouldn't block us from merging. Thanks :)

jreddig:
> > I was able to import a bunch of older experiments in my local! However, for the warning message, I didn't get a list of missing properties, just the old version error message. How do we want to handle the messaging for older experiments @amurphy-cl ?
> 
> For missing properties, most of them would have been modified and picked from the necessary old format fields or either filled with default values. Can you any specific missing property which you are looking for in the error messages?
The current release states that "factors" is missing from the format:
Screenshot 2023-04-28 at 1 58 12 PM
while the same experiment imported here gives a generic message:
Screenshot 2023-04-28 at 1 58 38 PM
Issue #620 was to make the messaging more helpful, and I just want to make sure that we're not undoing completed work. Which message is more helpful about why the version is mismatched?

amurphy-cl:
If we're able to identify missing properties, that could be helpful. A generic incompatibility message is also fine, but I'd rephrase it to something like "Warning: this experiment file is incompatible with the current version of UpGrade. Some features may not import or function as intended. Continue with import?"

danoswaltCL:
Once we're set that this is working right functionally, I'm for merging it so we are able to import things for this release. But I think we need to follow-up with a priority tech-debt task to consider revising/refactoring the code in the import feature files, such as `import-experiment.component.ts`. Code complexity / readability keeps drifting in wrong direction, and there are lots of missing types for runtime bugs to hide in.

ppratikcr7:
> > > I was able to import a bunch of older experiments in my local! However, for the warning message, I didn't get a list of missing properties, just the old version error message. How do we want to handle the messaging for older experiments @amurphy-cl ?
> > 
> > 
> > For missing properties, most of them would have been modified and picked from the necessary old format fields or either filled with default values. Can you any specific missing property which you are looking for in the error messages?
> 
> The current release states that "factors" is missing from the format: Screenshot 2023-04-28 at 1 58 12 PM while the same experiment imported here gives a generic message: Screenshot 2023-04-28 at 1 58 38 PM Issue #620 was to make the messaging more helpful, and I just want to make sure that we're not undoing completed work. Which message is more helpful about why the version is mismatched?

@jreddig The XPrize Indian River is a Simple Experiment and not a factorial experiment. So, during previous codebase we might not be exporting an empty array of `factors` in the JSON object. But, now we do export `factors` field. It is not needed for a Simple experiment, but for consistency we have added it and while on import it will append any default fields that needs to be there. So the earlier message was not actually correct for missing property. In case for factorial experiment if we miss `factors` field at root level, we add it and take `factors` object from the `partitions` object as we had it in here. Most incompatibilities of old versions have been now internally taken care of for smooth import. In case a critical property is missing whose value cannot be deduced from the uploaded old JSON, we would see that as a missing property warning in the modal. Hope this clarifies your doubt. :)

ppratikcr7:
> If we're able to identify missing properties, that could be helpful. A generic incompatibility message is also fine, but I'd rephrase it to something like "Warning: this experiment file is incompatible with the current version of UpGrade. Some features may not import or function as intended. Continue with import?"

But, I have already handled most of the missing properties if their information can be deduced from the uploaded old JSON object or with default values. This will enable a smooth import, and on the next export we get a new format JSON object.

I will update the warning message as per above. Thanks. :)

ppratikcr7:
> Once we're set that this is working right functionally, I'm for merging it so we are able to import things for this release. But I think we need to follow-up with a priority tech-debt task to consider revising/refactoring the code in the import feature files, such as `import-experiment.component.ts`. Code complexity / readability keeps drifting in wrong direction, and there are lots of missing types for runtime bugs to hide in.

Okay, I will check for the code complexity using our code metrics tool and update the file. Thanks :)

ppratikcr7:
> 

Currently we compare the version and based on the comparison, let the user know that the uploaded JSON version is either an older/newer version. Should we have that in the warning message?

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
```diff
@@ -140,186 +250,156 @@ export class ImportExperimentComponent implements OnInit {
...
-    const isFactorialExperiment = experiment.type === EXPERIMENT_TYPE.FACTORIAL;
-    let missingPropertiesFlag = true;
+  async validateExperimentJSON(experiment: Experiment) {
+    let missingProperties = this.checkForMissingProperties({ schema: this.experimentSchema, data: experiment });
+    const missingPropertiesList = missingProperties !== '' ? missingProperties.split(', ') : [];
+
+    if (missingPropertiesList.includes('backendVersion')) {
+      const currentBackendVersion = await this.versionService.getVersion();
+      experiment = { ...experiment, backendVersion: currentBackendVersion.toString() };
+      const index = missingPropertiesList.indexOf('filtebackendVersionrMode');
```
danoswaltCL, line 260: `'filtebackendVersionrMode'` looks like a typo

ppratikcr7: Oh yes, thanks for pointing this out. :)

danoswaltCL:
thx for the refactor! I will plan on testing this with experiment jsons provided and some others that I can find this afternoon. 👍

danoswaltCL (changes requested):
In my local, batch imports over 3+ files cause /imports to hang and then fail. Afterwards, the /paginated and /segments calls no longer work. @ppratikcr7 please take a look.

ppratikcr7:
> In my local, batch imports over 3+ files cause /imports to hang and then fail. Afterwards, the /paginated and /segments calls no longer work. I investigated other branches but this release has too many changes to truly compare existing behavior to know if this is related to this PR or other pre-existing release code issues in the import service. @ppratikcr7 please take a look.

Okay, let me check on this.

ppratikcr7:
> In my local, batch imports over 3+ files cause /imports to hang and then fail. Afterwards, the /paginated and /segments calls no longer work. I investigated other branches but this release has too many changes to truly compare existing behavior to know if this is related to this PR or other pre-existing release code issues in the import service. @ppratikcr7 please take a look.

@danoswaltCL I checked in dev branch, this is an existing issue for bulk experiments imports over 3-4 files which hangs Upgrade UI. We should create a new issue for this and I can work on that separately. And let this PR go in release, if all other things are fine. :)

danoswaltCL:
Ok, I was thinking the same thing. I'll bring this up in our morning mtg and see if we're okay to merge with eyes on the existing issue.

ppratikcr7:
> Ok, I was thinking the same thing. I'll bring this up in our morning mtg and see if we're okay to merge with eyes on the existing issue.

Okay, I have created separate [issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/835) for the bulk import issue and assigned @Yagnik56.

PR #816, opened 2023-05-01
Bugfix/664+metrics stats validation

RidhamShah:
Added errors for not choosing the metric name from drop-down and in the participant stepper.
The Statics menu in the metric stepper will be able to edit after choosing a metric.


jreddig (approved):

jreddig:
Double checking that this should be merged into dev for the next release? Or should it be merged into the release branch since it's a bugfix?

danoswaltCL:
i think we should be okay going into next release myself, doesn't seem like a terrible blocker or anything

PR #821, opened 2023-05-02
change markdata object

jreddig:


clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -23,14 +25,18 @@ export default async function markExperimentPoint(
   clientError?: string
 ): Promise<Interfaces.IMarkExperimentPoint> {
   let data: markData = {
```
danoswaltCL, line 27: can we call this something like `request` or `requestBody` or something so we don't have that awkward `data.data` below?

danoswaltCL (approved):

PR #829, opened 2023-05-03
Moving payload table to design stepper scrolling

Yagnik56:
This PR is created to update design stepper view for simple experiment by removing buttons which switch between condition table and condition-payload table and moving condition-payload table below condition table accessible through scrolling. This changes are done to keep design stepper view consistent between factorial and simple experiment.

Here are some screenshots,
conditonPayload table will be hidden until both decision points table and condition table have atleast one row.
image

Once both table have atleast row value in tables, conditionPayload table will be visible.
image
image


danoswaltCL:
💯 looks good to me, my only question is to @zackcl and @amurphy-cl: would it be preferable to have the payload table show as empty instead of appearing? With a message that says something like "payload options will appear once valid decision point and conditions have been entered"?

danoswaltCL:
@zackcl @amurphy-cl any thoughts on this?

zackcl:
@danoswaltCL I'm fine with either way, but I personally prefer hiding the payload table until at least one decision point and one condition are added so that the user can focus on one thing at a time.

amurphy-cl:
agree with Zack

PR #832, opened 2023-05-04
Bugfix/830 - Solved no-payload in client side

RidhamShah:
Solved `experimentCondition` throws error on not given payload
Modified `markExperiment` data structure in client side.


clientlibs/js/src/functions/getExperimentCondition.ts
```diff
@@ -5,16 +5,16 @@ export class Assignment {
   private _payloadType: PAYLOAD_TYPE;
```
VivekFitkariwala, line 5: You have missed updating the return type of `getPayload` and `getFactorPayload` to `IPayload | null`

clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -1,15 +1,11 @@
...
 
 interface markData {
   userId: string;
   status: MARKED_DECISION_POINT_STATUS;
-  data: {
-    site: string;
-    assignedCondition: { conditionCode: string; experimentId?: string },
-    target?: string;
-  }
+  data: IExperimentAssignmentv4;
```
VivekFitkariwala, line 8: @danoswaltCL this is important change because the structure of `getAllExperimentCondition` for simple experiment and factorial experiment is different and the change need to be reflected on Java client.

clientlibs/js/src/functions/markExperimentPoint.ts
```diff
@@ -1,15 +1,11 @@
 import { Interfaces, Types } from '../identifiers';
```
VivekFitkariwala, line 1: Also update `IMarkExperimentPoint` which has `target` as optional parameter

clientlibs/js/src/functions/getExperimentCondition.ts
VivekFitkariwala, line 5: `getExperimentCondition` has optional `target` parameter remove `?` that.

VivekFitkariwala (changes requested):
@RidhamShah check there are no optional `target` parameter in the client side SDK

VivekFitkariwala (dismissed):

jreddig (approved):
I believe this solves the issues Zack reported, and I was able to do the client flow using this library

danoswaltCL:
@RidhamShah is this ready to go? It's got 2 approvals but it's still in draft

ppratikcr7 (approved):

PR #858, opened 2023-05-18
replace ng-datetime-picker package dependency

Yagnik56:
This PR replace ng-datetime-picker package dependency with @danielmoncada/angular-datetime-picker.


Yagnik56:
Hey @danoswaltCL, before you approve this PR it will be good if you check this changes locally to confirm it is working fine. I checked it from my side but it will be good if u can do it as well.

danoswaltCL (approved):

PR #861, opened 2023-05-25
Add frontend changes

zackcl:
Changes in the frontend:

1. Updating upgrade types
  a. Adding within subject in Unit of assignment enum
  b. Disable consistency rule
2. Adding new unit of assignment Within Subject with options (lets remove round robin fixed as it can be achieved through round robin ordered)
  a. Random
  b. Round robin random
  c. Round robin ordered
3. Update metrics
  a. hide non-repeatable/simple metrics
  b. For repeatable/group metrics hide Earliest and Most Recent option in Repeated Measure
  
Discussion needed:
  - No Repeated Measure option will appear for categorical metric in within-subjects experiment (after hiding the Earliest and Most Recent options). Should we not support categorical metrics in within-subjects experiments?
  - What should happen if the user creates a non-within-subjects experiment using simple metrics, then later change the Unit of Assignment to "Individual", should we remove only the simple metrics, or clear the whole metrics table?

### DIFF
--- a/backend/packages/Upgrade/src/api/models/Experiment.ts
+++ b/backend/packages/Upgrade/src/api/models/Experiment.ts
@@ -64,6 +64,7 @@ export class Experiment extends BaseModel {
 
   @IsNotEmpty()
   @Column({
+    nullable: true,
     type: 'enum',
     enum: CONSISTENCY_RULE,
   })
--- a/backend/packages/Upgrade/src/api/models/Experiment.ts
+++ b/backend/packages/Upgrade/src/api/models/Experiment.ts
@@ -16,7 +16,7 @@ import {
   IExperimentSearchParams,
   IExperimentSortParams,
   EXPERIMENT_TYPE,
-  WITHING_SUBJECT_ALGORITHM,
+  CONDITION_ORDER,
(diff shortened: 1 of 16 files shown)


zackcl:
@RidhamShah When I ran your branch, I had this error:
```
QueryFailedError: column experiment.algorithm does not exist
    at QueryFailedError.TypeORMError [as constructor] 
```
I was able to fix this by setting the `TYPEORM_SYNCHRONIZE` to `true`.
Maybe this is the reason why it says some checks were not successful? 
Could you please see if you can fix this? Thank you!

RidhamShah:
image

@zackcl can you check on this, `Mean` should be the only option so I don't think there must be any drop-down, and also clicking on the drop-down does nothing.

zackcl:
@RidhamShah But I think "Mean" is also not supported for the categorical metrics, so I'm not sure if we even need to support categorical metrics for within-subjects experiments. Let's discuss this in the meeting.

danoswaltCL (commented):
This all looks pretty straight-forward to me, I'd say this can get merged if it's done? @zackcl

RidhamShah (approved):

PR #865, opened 2023-05-26
Within subjects experiment feature

Yagnik56:
This PR contains below verticals as per the scope of work [document](https://docs.google.com/document/d/1f7F1qyWMI6tVH_eLBRuU_XRvR6c00d5CvTJSETLUk_c/edit?usp=sharing):

1. CRUD
2. Assign & Mark Operation
3. Experiment Detail View - Enrollments
4. Experiment Detail View - Metrics


Yagnik56:
GetAllConditions changes are being implemented in this branch so avoid merging this

types/src/index.ts
```diff
@@ -50,5 +51,7 @@ export {
   CaliperGradingProfile,
   CaliperEnvelope,
   IPayload,
+  AssignedCondition,
```
VivekFitkariwala, line 54: Don't export `AssignedCondition` use it like this
`IExperimentAssignmentv5.AssignedCondition[number]`

Because each Experiment Assignment can have it's own type of `AssignedCondition`

backend/packages/Upgrade/src/api/Algorithms.ts
```diff
@@ -0,0 +1,200 @@
...
+  experiment: Experiment,
+  conditionPayloads: ConditionPayloadDTO[],
+  site: string,
+  target: string,
+  factors: FactorDTO[],
+  userID: string,
+  monitoredDecisionPointLogsLength: number
+): IExperimentAssignmentv5 {
+  const assignedData = convertToAssignedCondition(experiment, conditionPayloads, site, target, factors);
+  let assignedConditionsArray = assignedData;
```
VivekFitkariwala, line 18: Not required. Just convert `const assignedData`to `let assignedData`

VivekFitkariwala: Why not make this into `Switch` statement? It will be more readable

backend/packages/Upgrade/src/api/Algorithms.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  return {
+    site: site,
+    target: target,
+    assignedCondition: assignedConditionArray,
+    assignedFactor: experiment.type === EXPERIMENT_TYPE.FACTORIAL ? assignedFactorsArray : null,
+  };
+}
+
+function getAssignedFactor(conditionAssigned: ExperimentCondition, factors: FactorDTO[]): object {
```
VivekFitkariwala, line 177: Change the return type to proper typing `object` is too generic

backend/packages/Upgrade/src/api/controllers/validators/MarkExperimentValidator.v4.ts
```diff
@@ -16,6 +16,6 @@ export class MarkExperimentValidatorv4 {
   };
 
   public status?: MARKED_DECISION_POINT_STATUS;
-
+  public uniquifier?: string;
```
VivekFitkariwala, line 19: Why do we need uniquifier in v4?

RidhamShah: code updated

backend/packages/Upgrade/src/api/Algorithms.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/Algorithms.ts
RidhamShah: code updated

types/src/index.ts
RidhamShah: code updated

zackcl:
I've made a commit regarding the Statistic and Repeated Measure fields in the Metrics step.
* The Repeated Measure field is located above the Statistic field
* Categorial metric has "Earliest", "Most Recent", "Count", and "Percentage" options for Repeated Measure field
* The available options for Statistic depends on Repeated Measure selection for Categorial metric
* There's no difference in the UI between between-subjects and within-subjects except that the global metrics are not supported in a within-subjects experiment.

I still need to work on invalidating the global (simple) metric rows whenever the unit of assignment updates to "Within-subjects" (in case the experiment had global metrics), but if I try to save global metrics in the stepper, I get the following error:

* Console:
Screen Shot 2023-06-21 at 8 00 08 PM

* Terminal:
Screen Shot 2023-06-21 at 8 09 23 PM

Could this issue be fixed? Thanks!

ppratikcr7:
> I've made a commit regarding the Statistic and Repeated Measure fields in the Metrics step. Here are the changes:
> 
> * The Repeated Measure field is located above the Statistic field
> * Categorial metric has "Earliest", "Most Recent", "Count", and "Percentage" options for Repeated Measure field
> * The available options for Statistic depends on Repeated Measure selection for Categorial metric
> * There's no difference in the UI between between-subjects and within-subjects except that the global metrics are not supported in a within-subjects experiment.
> 
> I still need to work on invalidating the global (simple) metric rows whenever the unit of assignment updates to "Within-subjects" (in case the experiment had global metrics), but if I try to save global metrics in the stepper, I get the following error:
> 
> * Browser:
> 
> Screen Shot 2023-06-21 at 8 00 08 PM
> * Terminal:
> 
> Screen Shot 2023-06-21 at 8 09 23 PM
> Could this issue be fixed? Thanks!

Okay, let me check once.

zackcl:
For this sprint, I've removed both `Count` and `Percentage` from the `Repeated Measure` options for categorical metric. Additionally, I've renamed `Repeated Measure` to `Individual Statistic`, and `Statistic` to `Aggregate Statistic` in the Metrics table. Note that these changes only affect the displayed strings, not the variable names yet.

ppratikcr7:
> I've made a commit regarding the Statistic and Repeated Measure fields in the Metrics step. Here are the changes:
> 
> * The Repeated Measure field is located above the Statistic field
> * Categorial metric has "Earliest", "Most Recent", "Count", and "Percentage" options for Repeated Measure field
> * The available options for Statistic depends on Repeated Measure selection for Categorial metric
> * There's no difference in the UI between between-subjects and within-subjects except that the global metrics are not supported in a within-subjects experiment.
> 
> I still need to work on invalidating the global (simple) metric rows whenever the unit of assignment updates to "Within-subjects" (in case the experiment had global metrics), but if I try to save global metrics in the stepper, I get the following error:
> 
> * Browser:
> 
> Screen Shot 2023-06-21 at 8 00 08 PM
> * Terminal:
> 
> Screen Shot 2023-06-21 at 8 09 23 PM
> Could this issue be fixed? Thanks!

Hey Zack, can you please elaborate the exact scenario in which the global metrics is not updated. I tried to create a non within-subject experiment and I was able to add and update global metrics. Are you updating the experiment to within subject and have global metrics in it and you get error on saving? I was able to save the experiment after updating to within subjects experiment without any error.

jreddig:
I do get this same error when saving a global metric, the previous error is 
`driverError: error: null value in column "repeatedMeasure" of relation "query" violates not-null constraint`
The experiment being saved is:
`{"name":"fewaf","description":"","consistencyRule":"individual","assignmentUnit":"individual","type":"Simple","context":["mathstream"],"tags":[],"logging":false,"conditions":[{"id":"4140c580-efc8-4475-a3f6-ec6271684d54","conditionCode":"question-hint-default","assignmentWeight":"50","description":null,"order":1,"name":""},{"id":"72f123ef-6b63-4ce3-8a1f-001513d2bea1","conditionCode":"question-hint-tutorbot","assignmentWeight":"50","description":null,"order":2,"name":""}],"conditionPayloads":[],"partitions":[{"id":"21a813a5-fa70-4124-8bd6-aa4264001e17","site":"lesson-stream","target":"question-hint","description":"","order":1,"excludeIfReached":false}],"experimentSegmentInclusion":{"userIds":[],"groups":[],"subSegmentIds":[],"type":"private"},"experimentSegmentExclusion":{"userIds":[],"groups":[],"subSegmentIds":[],"type":"private"},"filterMode":"excludeAll","queries":[{"name":"fdsaf","query":{"operationType":"sum"},"metric":{"key":"totalTimeSeconds"},"repeatedMeasure":null}],"endOn":null,"enrollmentCompleteCondition":null,"startOn":null,"state":"inactive","postExperimentRule":"continue","revertTo":null}`

Repeated measure is null, which I think makes sense for global statistics?

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -98,7 +112,7 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
           operationType: [null, Validators.required],
           compareFn: [null, Validators.required],
           compareValue: [null, Validators.required],
-          repeatedMeasure: [REPEATED_MEASURE.mostRecent, Validators.required],
+          repeatedMeasure: [null, Validators.required],
```
jreddig, line 115: It looks like this line is behind @zackcl 's problem. Should repeated measure have a default value if it's not allowed to be null?

jreddig: Actually no, it's line 289

frontend/projects/upgrade/src/app/features/dashboard/home/components/metrics/metrics.component.ts
```diff
@@ -286,7 +299,7 @@ export class MonitoredMetricsComponent implements OnInit, OnChanges, OnDestroy {
     operationType = null,
     compareFn = null,
     compareValue = null,
-    repeatedMeasure = REPEATED_MEASURE.mostRecent
+    repeatedMeasure = null
```
jreddig, line 302: Actually, this is the line that is causing @zackcl 's global metric issue. Should repeatedMeasure have a default value if null is unacceptable?

ppratikcr7: Thanks @jreddig. Yeah, I just checked, Zack has updated this in his [commit](https://github.com/CarnegieLearningWeb/UpGrade/pull/865/commits/490351ff173f5114f21b7a87a5d1773d0b75db0b). Actually `repeatedMeasure` has `MOST RECENT` as default value and is not nullable:true in the model. We need to have `MOST RECENT` by default in the global metrics as well. I will revert this change.

clientlibs/js/src/functions/getExperimentCondition.ts
```diff
@@ -52,25 +76,51 @@ export class Assignment {
...
+        result.site,
+        result.target,
+        markObject,
+        result.assignedCondition[0].conditionCode,
+        result.assignedCondition[0].payload,
+        result.assignedFactor ? result.assignedFactor[0] : null
       );
+
+      // rotate the condition queue
+      result.assignedCondition.push(result.assignedCondition.shift());
```
danoswaltCL, line 120: isn't this rotation supposed to happen after doing `markExperimentPoint`, not `getAssignment`? The Java library is doing this on mark.

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -108,15 +109,15 @@ export default class UpgradeClient {
...
-      logCaliper: `${hostUrl}/api/v4/log/caliper`,
+      init: `${hostUrl}/api/v5/init`,
+      getAllExperimentConditions: `${hostUrl}/api/v5/assign`,
+      markExperimentPoint: `${hostUrl}/api/v5/mark`,
+      setGroupMemberShip: `${hostUrl}/api/v5/groupmembership`,
+      setWorkingGroup: `${hostUrl}/api/v5/workinggroup`,
+      failedExperimentPoint: `${hostUrl}/api/v5/failed`,
+      getAllFeatureFlag: `${hostUrl}/api/v5/featureflag`,
+      log: `${hostUrl}/api/v5/log`,
+      logCaliper: `${hostUrl}/api/v4/logCaliper`,
```
danoswaltCL, line 120: Should these be:

logCaliper: `${hostUrl}/api/v5/log/caliper`,
altUserIds: `${hostUrl}/api/v5/useraliases`,
addMetrics: `${hostUrl}/api/v5/metric`,

clientlibs/js/src/UpgradeClient.ts
```diff
@@ -225,16 +226,7 @@ export default class UpgradeClient {
...
- * Will return all the experiment conditions for the user.
- * Internally this uses the `context` and `userId` to query conditions for all eligible decision points at enrolling experiments for this user.
- *  
- * @example
- * ```typescript
- * const allExperimentConditionsResponse: IExperimentAssignmentv4[] = await upgradeClient.getAllExperimentConditions(workingGroup);
- * ```
- */
-  async getAllExperimentConditions(): Promise<IExperimentAssignmentv4[]> {
+  async getAllExperimentConditions(): Promise<IExperimentAssignmentv5[]> {
```
danoswaltCL, line 229: Can someone who made the TS Client Lib updates for v5 also update the TSDoc comments/examples so that they accurately reflect to library user (and the testers 😉 ) how the v5 functions are designed to work?

clientlibs/js/src/functions/getExperimentCondition.ts
```diff
@@ -52,25 +76,51 @@ export class Assignment {
       return null;
     }
   }
+
+  public markDecisionPoint(status: MARKED_DECISION_POINT_STATUS, uniquifier?: string, clientError?: string) {
```
danoswaltCL, line 80: Is this new public method on Assignment necessary? It looks like is just proxying the `Client.markExperimentPoint`, but I don't know what the benefit is to pass all of this `markObject` data this way?

clientlibs/js/src/functions/getExperimentCondition.ts
RidhamShah: Solved this part, now conditions will rotate on getAssignement call

clientlibs/js/src/UpgradeClient.ts
RidhamShah: Updated this code

clientlibs/js/src/UpgradeClient.ts
RidhamShah: Added TSDoc for `getAllExperimentConditions()`

clientlibs/js/src/functions/getExperimentCondition.ts
danoswaltCL: "we have removed the client.markExperimentPoint call" This is interesting, but unexpected! maybe we missed this in discussions @mswartzCL has set up some time to on Tues at 9 to meet, might be easier to chat in person to review changes in the lib API to know what the Java client library should be doing.

jerith666: I'm only noticing this in passing at a very high level, but: relying on "business logic" in the client libraries is, I think, a code smell.

We ought to strive for very simple client libraries, so as to encourage use of UpGrade from clients written in languages for which we don't provide a first-party library.

If you have to get a bunch of fiddly logic right in a client lib for an esoteric (or just unsupported) language, that's a deterrent to adoption.

Again, that may not be what's going on here at all, but sounded like it might be ... :)

VivekFitkariwala: @danoswaltCL We have decided in the meeting that we will use the assignment object to mark the experiment and that is the reason we have removed the mark function directly from the upgrade client.
```
const client = new UpgradeClient();
client.markExperimentPoint(site, target, ....)           // this method is remove from the v5

const assingment = client.getDecisionPoint(site, target);
assignment.markExperimentPoint(....).                  // this method is added in v5
```

RidhamShah: Client mark call is back. Reverted those changes and stack is also rotated from `client.markDecisionPoint()` call

RidhamShah: @danoswaltCL I would add a commit to rename every `markExperimentPoint` to `markDecoisionPoint`

danoswaltCL (approved):

PR #868, opened 2023-06-07
:sparkles: Updated client lib bundler

VivekFitkariwala:
Added separate bundler for browser and node environment
`UpgradeClient` is exported by default not a named export anymore.

# For browser
## Load as script
```
const lib = window["upgrade-client-lib"];
const Upgrade = lib.default;
const client = new Upgrade("user1", "http://localhost:8000", "portal");
```
## Load in typescript/es6 project
```
import UpgradeClient from 'upgrade_client_lib/dist/browser';
const client = new UpgradeClient("vivek", "http://localhost:3030", "math");
```
# For Node
```
import UpgradeClient from 'upgrade_client_lib/dist/node';
const client = new UpgradeClient("vivek", "http://localhost:3030", "math");
```

### DIFF
--- a/clientlibs/js/package-lock.json
+++ b/clientlibs/js/package-lock.json
@@ -18,6 +18,7 @@
         "@types/jest": "^29.2.1",
         "@types/node": "^18.11.9",
         "@types/uuid": "^9.0.1",
+        "dts-bundle-generator": "^8.0.1",
         "jest": "^29.2.2",
         "ts-jest": "^29.1.0",
         "ts-loader": "^9.4.1",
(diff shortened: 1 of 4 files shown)


VivekFitkariwala:
> What does `const Upgrade = lib.default;` do? Would it be possible to not require this line?

I have updated the webpack config to address this issue

zackcl:
I tried copying the `clientlibs/js/dist/browser/index.js` file but an error "Cannot read properties of null (reading 'type')" occurs when the experiment is running. You can also try it here: https://jsfiddle.net/m0w938zb/17/

Screen Shot 2023-06-08 at 8 50 36 AM

danoswaltCL:
I tested in the client-lib-tester and this seems to work (will need to make some changes to the tester code after this gets merged since it changes the import syntax). This works for me I think, we can merge into dev, but we'll need to discuss cherry-picking this and publishing a patched version for Mathstream on monday, I think they'll need a version 3 ts lib. @mswartzCL @zackcl

danoswaltCL (approved):

PR #874, opened 2023-06-09
update docker local .env

danoswaltCL:
updating local .env.local.docker to versioned controllers and adding mathstream / assign-prog context-metadatas cuz why not


jreddig (approved):
ty, i always forget to do this locally

PR #879, opened 2023-06-15
revert apache connector

jreddig:
fix for demo links


jerith666 (commented):
The changes in `clientlibs/java/` look right, but the changes to the two files under `backend/` look unrelated ... ?

jreddig:
Those were changes missed when pulling in /v1 :/ , the test file updates the controller endpoints to /v1 and the .env file adds the /v1 controller

danoswaltCL (approved):

PR #883, opened 2023-06-20
java libs for within subjects

jreddig:
updated libs to work with an array of conditions

Having some trouble with the async/caching for getting one condition after another. At what point do we rotate the array to pass on the next condition? markExperimentPoint? or getExperimentCondition? I'm having a hard time guaranteeing the global experiment list is defined by the time a condition is received or marked

Is rotating the conditions something the libs should do or can we rely on UpGrade's APIs to rotate the conditions?


jreddig:
Got something that works if you follow the assign -> mark process, but not if you get multiple assignments in a row (same condition each time). Not a fan of how I did it but its a start

jreddig:
moved popping the condition to markExperimentPoint, however in the attached Main.java, the first two conditions received are the same while the third condition is the second in the list

jreddig:
Okay, this should work, as long as the client isn't trying to do rapid-fire assign/mark calls

jreddig:
This is my proposal on how we move mark to Assignment. We call the one ExperimentClient mark method from Assignment so we dont have to recreate the apiservices or pass any extra objects/data around. assignment.markExperimentPoint will need the ExperimentClient as a parameter in order to access this method. 

We could also setup the api services in the Assignment object, just let me know!

clientlibs/java/src/main/java/org/upgradeplatform/responsebeans/Assignment.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class Assignment {
+
+    private String target;
+    private String site;
+    private ExperimentType experimentType;
+    private Condition assignedCondition;
+    private Map<String, Factor> assignedFactor;
+
+    public Assignment(String target, String site, ExperimentType experimentType, Condition assignedCondition) {
```
danoswaltCL, line 19: what if the constructor could take client as an argument? So you don't have to pass it into `markExperimentPoint` at the point of calling?

jreddig: i like that idea!

jreddig:
Per Dan's suggestion, moved the ExperimentClient parameter to the Assignment constructor so we can use the ExperimentClient's api services and the client can rotate the condition array

Also renamed ExperimentPoint to DecisionPoint

jreddig:
@jerith666 adding you to the pr to keep you appraised of upcoming changes and to collect your feedback (if you have any) on the updated library design

jreddig:
Included the previous usage of markDecisionPoint

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -256,48 +308,74 @@ private ExperimentsResponse findExperimentResponse(String site, String target,
...
+					(isStringNull(target) ? isStringNull(t.getTarget().toString())
+					: t.getTarget().toString().equalsIgnoreCase(target))).findFirst().orElse(null);
+			if (result != null) {
+				Condition[] rotatedCondition = Arrays.copyOf(result.getAssignedCondition(),
+						result.getAssignedCondition().length);
+				List<Condition> rotatedList = Arrays.asList(rotatedCondition);
+				Collections.rotate(rotatedList, -1);
+				rotatedCondition = rotatedList.toArray(rotatedCondition);
+				result.setAssignedCondition(rotatedCondition);
+				result.setAssignedFactor(result.getAssignedFactor());
```
jerith666, line 323: don't have time for a full review, and I mentioned this on another PR for the JS client, but: this kind of business logic in the client library feels like a code smell to me.  Why can't this condition rotation happen on the server?

PR #895, opened 2023-06-22
static mark status enums

danoswaltCL:
allow MARKED_DECISION_POINT_STATUS to be exposed on the client a la `UpgradeClient.MARKED_DECISION_POINT_STATUS.CONDITION_APPLIED` <--- static public property on the class constructor itself

this will allow js-only users or those who are not using the upgrade types package to use this enum for markExperimentPoint()


jreddig (dismissed):
good to merge? I think it's a good idea to expose the status enum

jreddig (approved):

PR #918, opened 2023-07-13
mark back on client plus renaming and organization

danoswaltCL:
This change:

- adds `markDecisionPoint` back to Client object
- changes `markExperimentPoint` to `markDecisionPoint`
- refactor and clean-up
- adds documentation
- improve the exported types so that everything mentioned in documentation is actually accessible as a type when 
- added better tests for within-subjects in client-lib tester


jreddig (approved):
Nice renaming!

PR #920, opened 2023-07-18
Feature/export csv within subject

Yagnik56:
This PR contain changes about exporting CSV file for withinSubject experiment.

### DIFF
--- a/backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
+++ b/backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
@@ -445,6 +445,69 @@ export class AnalyticsRepository {
       .execute();
   }
 
+  public async getCSVDataForWithInSubExport(
+    experimentId: string,
+    skip: number,
+    take: number
+  ): Promise<
+    Array<{
+      experimentId: string;
+      experimentName: string;
+      userId: string;
+      groupId: string;
+      partition: string;
+      conditionName: string;
+      firstDecisionPointReachedOn: string;
+      decisionPointReachedCount: number;
+      context: string[];
+      assignmentUnit: string;
+      group: string;
+      site: string;
+      target: string;
+    }>
+  > {
+    const individualEnrollmentRepository = getCustomRepository(IndividualEnrollmentRepository, 'export');
+    return individualEnrollmentRepository
+      .createQueryBuilder('individualEnrollment')
+      .select([
+        'experiment.id as "experimentId"',
+        'experiment.name as "experimentName"',
+        'experiment.context as "context"',
+        'experiment.assignmentUnit as "assignmentUnit"',
+        'experiment.group as "group"',
+        'monitored.site as "site"',
+        'monitored.target as "target"',
+        '"individualEnrollment"."userId" as "userId"',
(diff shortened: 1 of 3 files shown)


backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -445,6 +445,69 @@ export class AnalyticsRepository {
       .execute();
   }
 
+  public async getCSVDataForWithInSubExport(
+    experimentId: string,
+    skip: number,
+    take: number
+  ): Promise<
+    Array<{
```
danoswaltCL, line 453: Please declare a type instead of passing in a generic object definition. That way we can use that type where we are calling `any` in the other file.

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -183,7 +185,12 @@ export class AnalyticsService {
       let skip = 0;
       const take = 50;
       do {
-        const data = await this.analyticsRepository.getCSVDataForSimpleExport(experimentId, skip, take);
+        let data: any[];
```
danoswaltCL, line 188: I have a request, data needs a type for sure but we're not declaring types for the return values of `getCSVDataForSimpleExport` and `getCSVDataForWithInSubExport` right now, so:

Can you update the return types for those two functions with declared types, like `Promise<CSVDataForSimpleExport[]>` and `Promise<CSVDataForWithinSubExport[]>`.

Also, please call `data` something else, because everything is data... `csvExportData` at least?

Then we can say `let csvExportData: CSVDataForSimpleExport[] | CSVDataForWithinSubExport[];`, or even create a CSVExportData parent interface that both of these extend so you can have `let csvExportData: CSVExportData`.

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -320,6 +327,7 @@ export class AnalyticsService {
         // write in the file
         const csv = new ObjectsToCsv(csvRows);
         try {
+          experiment.type === EXPERIMENT_TYPE.FACTORIAL? csv.delimiter = ',' : false;
```
danoswaltCL, line 330: This ternary is pretty strange, that `false` is meaningless. We're just saying this, right?
```
if (experiment.type === EXPERIMENT_TYPE.FACTORIAL) {
  csv.delimiter = ',';
}
```

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
Yagnik56: resolved

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
Yagnik56: resolved

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
Yagnik56: resolved

danoswaltCL (approved):
great thank you @Yagnik56

PR #927, opened 2023-07-19
Enhancement: within subject metric logs refactor and complexity reduction

ppratikcr7:


danoswaltCL (approved):
Thank you @ppratikcr7 for taking the time to refactor to reduce complexity, looks like `analysis` went down from 63 to 50 so looking better so step in the right direction.

ppratikcr7:
> Thank you @ppratikcr7 for taking the time to refactor to reduce complexity, looks like `analysis` went down from 63 to 50 so looking better so step in the right direction.

@danoswaltCL Yeah. I want to still reduce it further.

PR #932, opened 2023-07-21
resolved issue of zero weight replaced by undefined

Yagnik56:
This PR resolve issue of zero condition weight converted to undefined, so every time 0 had to get added.  
issue: if weight is 0 then we were getting value from previous weight value but it is only assigned for equal weight in conditions, in other case when equal weight checkout is false it stays as undefined.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
@@ -758,7 +758,7 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
         this.conditions.controls.forEach((control, index) => {
           const assignmentWeightFormControl = control.get(SIMPLE_EXP_CONSTANTS.FORM_CONTROL_NAMES.ASSIGNMENT_WEIGHT);
           assignmentWeightFormControl.setValue(
-            control.value.assignmentWeight ? control.value.assignmentWeight : this.previousAssignmentWeightValues[index]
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-design/experiment-design.component.ts
```diff
@@ -758,7 +758,7 @@ export class ExperimentDesignComponent implements OnInit, OnChanges, OnDestroy {
         this.conditions.controls.forEach((control, index) => {
           const assignmentWeightFormControl = control.get(SIMPLE_EXP_CONSTANTS.FORM_CONTROL_NAMES.ASSIGNMENT_WEIGHT);
           assignmentWeightFormControl.setValue(
-            control.value.assignmentWeight ? control.value.assignmentWeight : this.previousAssignmentWeightValues[index]
+            control.value.assignmentWeight ? control.value.assignmentWeight : 0
```
danoswaltCL, line 761: are there are no scenarios here where we would want the previous assignment weight instead of 0 when assignment weight is undefined?

Yagnik56: the previous assignment weight assigned value only if equal weight is true and I didn't change it for equal weight case, but if equal weight isn't true then it is just empty array replacing our 0 value to undefined when we try to read any index value. I think even if we remove that line it will be fine but if in any case we get null or undefined weight this will also replace it with 0 and not throw error so it's win win in my view.

danoswaltCL (approved):

PR #934, opened 2023-07-25
Update all Client-Controller to validate API (New)

RidhamShah:
Created clientSide API's body validation for all versions.


danoswaltCL:
This will be great to have in. @RidhamShah In general, what should we now see if we put in a bad request? Just a 400 or are there specific error messages to expect when testing this?

Can the documentation for expected errors be updated in the swagger comments?

danoswaltCL:
@RidhamShah @VivekFitkariwala @ppratikcr7 is there still working going into these? they are still in draft but they haven't had updates. if Vivek's good I'm good, y'all can merge these in whenever

VivekFitkariwala:
Looks good to me.

VivekFitkariwala (approved):

danoswaltCL (approved):

PR #944, opened 2023-07-31
Analytics and other remaining class validations

RidhamShah:
I have added, Analytics, Setting, Scheduladr APIs Class-validations

### DIFF
--- a/backend/packages/Upgrade/test/unit/controllers/AnalyticsController.test.ts
+++ b/backend/packages/Upgrade/test/unit/controllers/AnalyticsController.test.ts
@@ -49,7 +49,7 @@ describe('Analytics Controller Testing', () => {
       .post('/api/stats/enrollment/date')
       .send({
         experimentId: uuid(),
-        dateEnum: 'xyz',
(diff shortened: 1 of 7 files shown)


backend/packages/Upgrade/test/unit/controllers/AnalyticsController.test.ts
```diff
@@ -49,7 +49,7 @@ describe('Analytics Controller Testing', () => {
       .post('/api/stats/enrollment/date')
       .send({
         experimentId: uuid(),
-        dateEnum: 'xyz',
+        dateEnum: 'last_seven_days',
```
VivekFitkariwala, line 52: Use the enum here

VivekFitkariwala:
Looks good to me.

backend/packages/Upgrade/test/unit/controllers/AnalyticsController.test.ts
RidhamShah: Code updated

PR #952, opened 2023-08-03
Export All segments features added and export segment api modified

Yagnik56:
This PR contain export all segment feature, while we had button for export all segments but it's callback function wasn't there so it was throwing error. This issue is resolved now.


jreddig:
Looks good to me! Checked it out locally and I can export all segments or each segment individually

jreddig (approved):

Issue #931, opened 2023-07-20
UpGrade TS lib changes for cli-services integration


danoswaltCL:
**Is your feature request related to a problem? Please describe.**
We would like to integrate the typescript UpGrade library into cli-services, but currently the library needs some structural enhancements in order to play according to cli-services requirements.

**Describe the solution you'd like**
1. `cli-services` will prefer to use their own http client provider. We can restructure the TS lib to offer the core-functionality `UpGradeClient` as well as an `UpGradeHTTPClient` default client, something like:

```ts
import UpGradeClient, { ICustomHTTPClient } from 'upgrade-client-lib/dist/browser';
import UpGradeHTTPClient from 'upgrade-client-lib/dist/http-client';

const upgradeClient = new UpGradeClient(userId, host, context, { httpClient: UpGradeHTTPClient });

// OR

const customHTTPClient: ICustomHTTPClient = {
  // provide an interface for wrapping custom http client calls.
}

const upgradeClient = new UpGradeClient(userId, host, context, { httpClient: customHTTPClient });
```

This will allow users who want more control over their HTTP requests to use our library with their own provider (and keep the library modular so they can only import the parts they need, which will be less of an overhead in terms of size.) It also gives allows users who don't care to be able to still use UpGrade out of the box with all that it needs.

This would also necessitate the creation of a custom HTTP client interface that can wrap whatever client is provided.

2. `cli-services` likely has some constraints on what peer dependencies get pulled in via our library. We will need to make sure we audit our dependencies for issues that could block integration.

3. We need to make sure we are exporting our lib in a javascript module format that works for cli-service imports.

4. These changes will necessitate some minor refactoring of the library's code structure around the `fetchDataService` to straighten out some things that are creating unnecessary challenges.

danoswaltCL:
@kschaefe let us know if this makes sense as a starting point for integrating into cli-services

PR #955, opened 2023-08-10
TS Lib reorg / bundling with and without httpClient module

danoswaltCL:
This change is a pretty substantial re-arrange of the TS Client Library in order support https://github.com/CarnegieLearningWeb/UpGrade/issues/931. Apologies for a large PR, not sure I could have broken this down into smaller bits.

The end-result to know is that we will now have 4 types of bundles; currently there is `browser` and `node`, now we will also have `browser-lite` and `node-lite`. The `lite` bundles will not bundle an http module. Instead, a lib consumer can opt to only import the functionality without a bundled http-provider. They will provide their own http-client by binding to an`IHttpClientWrapper` interface. See ([this](https://github.com/CarnegieLearningWeb/UpGrade/blob/feature/POC-custom-http-client-ts-lib/clientlibs/libTesters/client-lib-tester-frontend/src/app/mockFrontendClientAppComponents/general-test-for-version5.service.ts#L248) and [this](https://github.com/CarnegieLearningWeb/UpGrade/blob/feature/POC-custom-http-client-ts-lib/clientlibs/js/src/DefaultHttpClient/DefaultHttpClient.ts#L15) for examples).

This will be passed in constructor via `new UpGradeClient('userId', 'url', 'context', { httpClient: <myHttpClient> })`

The bundles are created using two new global boolean variables injected from webpack, `USE_CUSTOM_HTTP_CLIENT` and `IS_BROWSER`. The default bundle will automatically use the http client same as usual.

The public API remains unchanged and the logic remains the same, but organized into class modules in order to allow for bundling with OR without an http-client module. Hopefully this organization is also a little easier to follow (and is in part made to mimic the structure of the Java client more closely where it makes sense).

- UpGradeClient = public API class, same as before
- ApiService = asynchronous data fetches
- DataService = storing data and data-related util functions
- DefaultHttpClient = module that binds to `IHttpClient
- Assignment = assignment class

Also took the opportunity to clean up:
- Types
- Naming consistency
- Webpack / exported bundles
- Added many unit tests (github copilot helps a lot with this, btw)

Still working regression testing and documenting it, please let me know of thoughts / concerns, I went a little crazy with this one but it was hard not to while reorganizing to try and tidy up also...


clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,311 @@
...
+  private setHttpClient(httpClient: UpGradeClientInterfaces.IHttpClientWrapper) {
+    console.log({ USE_CUSTOM_HTTP_CLIENT });
+    if (USE_CUSTOM_HTTP_CLIENT) {
+      if (httpClient) {
+        return httpClient;
+      } else {
+        throw new Error('Please provide valid httpClient.');
+      }
+    } else {
+      return new DefaultHttpClient();
```
kschaefe, line 63: We could figure a way to bundle this in another library is there's client demand.  But including a basic lightweight implementation is probably fine.

kschaefe (changes requested):
Did I miss or do we not have any documentation on the why we moved to an interface for the HTTP client and how to build a custom one?

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: hm my goal was to find some webpack way to create two different bundles depending on the USE_CUSTOM_HTTP_CLIENT boolean supplied by webpack, one with the DefaultHttpClient and one without it (`browser` v `browser-lite`), but I deceived myself into thinking this would work. The package sizes are the same, so yeah it's bundling with.

I don't want the library user to have to import two packages, `upgrade-client-lib` and `upgrade-client-default-http-client`, just to have to marry them together in `new UpGradeClient('blah', 'blah', 'blah', { httpClient: new DefaultHttpClient() }})`... I'd the like "full" bundle to just have it by default without having to pass it in as a param, and the "lite" bundle to omit it and throw an error it is not passed in. Is there some simpler way to accomplish that I'm overthinking?

danoswaltCL:
> Did I miss or do we not have any documentation on the why we moved to an interface for the HTTP client and how to build a custom one?

Documentation still to come, making sure this approach works while working out the details.

clientlibs/js/src/ApiService/ApiService.ts
kschaefe: Webpack configs are not my thing.  I've got no real advice to offer there.

Thinking about it differently, if you had the "other" package with the default client bundle everything together and the main package just have the basics.  Then, you're not using Webpack to carve anything out, you're using it to bring things together in the case of the "other" package.

VivekFitkariwala: You can use webpack externals https://webpack.js.org/configuration/externals/ for removing the DefaultHttpClient if USE_CUSTOM_HTTP_CLIENT flag is true and given that axios is used only inside this client it will also get excluded from the bundle.

danoswaltCL: great, I'll look into that

danoswaltCL:
will put up different PR when officially ready for review

PR #981, opened 2023-08-22
upgrade to node 18

bcb37:
This PR:
- sets all the CI values for node to 18
- updates the nix package to one that's compatible with node 18
- changes install commands to ignore peer dependency errors (npm got stricter about that)

### DIFF
--- a/.github/actions/backend-build/action.yml
+++ b/.github/actions/backend-build/action.yml
@@ -24,7 +24,7 @@ runs:
     - name: "[Backend] Install Package Dependencies"
       shell: bash
       working-directory: backend/packages/Upgrade
-      run: npm ci
+      run: npm ci --legacy-peer-deps
     - name: "[Backend] Install Build Dependencies"
       shell: bash
       working-directory: backend
--- a/.github/actions/generic-npm-build/action.yml
+++ b/.github/actions/generic-npm-build/action.yml
@@ -18,7 +18,7 @@ runs:
     - uses: tw3lveparsecs/github-actions-setvars@v0.1
     - uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version: 18
         check-latest: true
     - name: Set Version
       shell: bash
--- a/.github/actions/generic-npm-build/action.yml
+++ b/.github/actions/generic-npm-build/action.yml
@@ -12,13 +12,17 @@ inputs:
     description: Version to set in package.json
     required: false
     default: ""
+  node_version:
+    description: Version of nodejs to build with
+    required: false
+    default: 18
 runs:
   using: composite
   steps:
     - uses: tw3lveparsecs/github-actions-setvars@v0.1
     - uses: actions/setup-node@v3
       with:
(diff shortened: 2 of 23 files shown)


danoswaltCL:
There must be something in the scheduler function lambda code that Node 18 dislikes that we may need to resolve. https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported (the accepted answer is pretty bad! the next one offers a saner explanation).

Perhaps for the sake of getting this in, we can keep it at 16 in the github task-runner yaml for lambda build and open a separate issue to resolve this (this would be code found at `backend/packages/Scheduler`). This is a lower priority than other things, as this function is a nice-to-have that is rarely (never?) utilized.

bcb37:
> There must be something in the scheduler function lambda code that Node 18 dislikes that we may need to resolve. https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported (the accepted answer is pretty bad! the next one offers a saner explanation).
> 
> Perhaps for the sake of getting this in, we can keep it at 16 in the github task-runner yaml for lambda build and open a separate issue to resolve this (this would be code found at `backend/packages/Scheduler`). This is a lower priority than other things, as this function is a nice-to-have that is rarely (never?) utilized.

Ok, that makes sense. I suspect upgrading to webpack 5 might fix the problem, but that's not usually a trivial thing to do.

danoswaltCL:
ah couple notes, the terraform scripts (and any AWS-related configs in the repo) are no longer what SRE wants to use. These scripts are in purgatory, they should be removed from the open-source project (see https://carnegielearning.slack.com/archives/C051CUABJKA/p1680032439060529) for some conversation and apparently a task i never followed up on. These were only really used to launch new resources, and managed in AWS thereafter. So, no updates there will have any effect.

Also, the lambda build in github workflows should get set to 16 there, as the error is github-job-runner trying to build it using 18. Although setting dockerfiles to 18 will get us cooking for frontend / backend, I forgot that the lambdas we have will need updates in AWS manually for the actual runtimes, no docker there.

bcb37:
https://github.com/CarnegieLearningWeb/UpGrade/issues/982

> ah couple notes, the terraform scripts (and any AWS-related configs in the repo) are no longer what SRE wants to use. These scripts are in purgatory, they should be removed from the open-source project (see https://carnegielearning.slack.com/archives/C051CUABJKA/p1680032439060529) for some conversation and apparently a task i never followed up on. These were only really used to launch new resources, and managed in AWS thereafter. So, no updates there will have any effect.
> 
> Also, the lambda build in github workflows should get set to 16 there, as the error is github-job-runner trying to build it using 18. Although setting dockerfiles to 18 will get us cooking for frontend / backend, I forgot that the lambdas we have will need updates in AWS manually for the actual runtimes, no docker there.

I'm not seeing the error anymore. But if the AWS console (or cli) needs to be involved, I still don't have access.

danoswaltCL (approved):

PR #986, opened 2023-08-25
reorg ts lib as classes, webpack bundle updates, http-client-wrapper

danoswaltCL:
This change is a pretty substantial re-arrange of the TS Client Library in order support https://github.com/CarnegieLearningWeb/UpGrade/issues/931. Apologies for a large PR, not sure I could have broken this down into smaller bits.

The end-result to know is that we will now have 4 types of bundles; currently there is browser and node, now we will also have browser-lite and node-lite. The lite bundles will not bundle an http module. Instead, a lib consumer can opt to only import the functionality without a bundled http-provider. They will provide their own http-client by binding to anIHttpClientWrapper interface. See ([this](https://github.com/CarnegieLearningWeb/UpGrade/blob/feature/POC-custom-http-client-ts-lib/clientlibs/libTesters/client-lib-tester-frontend/src/app/mockFrontendClientAppComponents/general-test-for-version5.service.ts#L248) and [this](https://github.com/CarnegieLearningWeb/UpGrade/blob/feature/POC-custom-http-client-ts-lib/clientlibs/js/src/DefaultHttpClient/DefaultHttpClient.ts#L15) for examples).

This will be passed in constructor via new UpGradeClient('userId', 'url', 'context', { httpClient: <myHttpClient> })

The bundles are created using two new global boolean variables injected from webpack, USE_CUSTOM_HTTP_CLIENT and IS_BROWSER. The default bundle will automatically use the http client same as usual.

The public API remains unchanged and the logic remains the same, but organized into class modules in order to allow for bundling with OR without an http-client module. Hopefully this organization is also a little easier to follow (and is in part made to mimic the structure of the Java client more closely where it makes sense).

UpGradeClient = public API class, same as before
ApiService = asynchronous data fetches
DataService = storing data and data-related util functions
DefaultHttpClient = module that binds to `IHttpClient
Assignment = assignment class
Also took the opportunity to clean up:

Types
Naming consistency
Webpack / exported bundles
Added many unit tests (github copilot helps a lot with this, btw)


danoswaltCL:
@VivekFitkariwala @RidhamShah @ppratikcr7 can someone on playpower review / approve this change? I want to make sure it's not a surprise to anyone. There is a little bit of urgency to get it in but I'm happy to make changes if something looks doubtful in my approach. Thank you.

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,313 @@
...
+      getAllFeatureFlag: `${this.hostUrl}/api/${this.apiVersion}/featureflag`,
+      log: `${this.hostUrl}/api/${this.apiVersion}/log`,
+      logCaliper: `${this.hostUrl}/api/${this.apiVersion}/log/caliper`,
+      altUserIds: `${this.hostUrl}/api/${this.apiVersion}/useraliases`,
+      addMetrics: `${this.hostUrl}/api/${this.apiVersion}/metric`,
+    };
+    this.httpClient = this.setHttpClient(config.httpClient);
+  }
+
+  private setHttpClient(httpClient: UpGradeClientInterfaces.IHttpClientWrapper) {
```
bcb37, line 54: Here it seems like you could import a non-lite build (maybe by accident), and then supply an http client that would be silently ignored in favor of the default one. Probably not a huge issue, but maybe worth documenting?

clientlibs/js/src/ApiService/ApiService.types.ts
```diff
@@ -0,0 +1,22 @@
+import { UpGradeClientEnums } from 'types';
+
+export interface IApiServiceRequestParams {
+  path: string;
+  method: UpGradeClientEnums.REQUEST_METHOD;
+  options?: any;
```
bcb37, line 6: I don't see the param 'options' used anywhere; `sendRequest()`  takes the other three params and creates 'options' from `path`. Is it needed for some other reason?

bcb37 (commented):
Just a couple questions. Also, is the client-lib-tester at all a part of the CI, or is it a tool that's available during development? I ask because it seems much less formally audited (lots of commenting-out of stuff etc.)?

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: I think that may be true, that would be simple enough to catch an throw a proper error, i can add that.

clientlibs/js/src/ApiService/ApiService.types.ts
danoswaltCL: 👀 maybe not, I'll look

danoswaltCL:
@bcb37 client-lib tester very beta, and only for local testing right now. I would love to take another swing at simplifying the usability of it and build on it. I can imagine it or something like it as a deployed resource tied to our dev environments, but it's still cooking.

clientlibs/js/src/ApiService/ApiService.types.ts
danoswaltCL: was leftover from a refactor, removed

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: updated and removed console.log

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,318 @@
...
+  IFeatureFlag,
+  IFlagVariation,
+  IGroupMetric,
+  ILogInput,
+  ISingleMetric,
+} from 'upgrade_types';
+import { DataService } from 'DataService/DataService';
+import { IApiServiceRequestParams, IEndpoints } from './ApiService.types';
+import { IMarkDecisionPointParams } from 'UpGradeClient/UpGradeClient.types';
+import * as uuid from 'uuid';
```
VivekFitkariwala, line 15: As you are only using v4 you can change the import statement to
import { v4 as uuidv4 } from 'uuid';

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,318 @@
...
+  private apiVersion: string;
+  private clientSessionId: string;
+  private httpClient: UpGradeClientInterfaces.IHttpClientWrapper;
+  private api: IEndpoints;
+
+  constructor(config: UpGradeClientInterfaces.IConfig, private dataService: DataService) {
+    this.context = config.context;
+    this.hostUrl = config.hostURL;
+    this.token = config.token;
+    this.clientSessionId = config.clientSessionId;
```
VivekFitkariwala, line 35: It might be good idea to create your own sessionId in case user doesn't provide.

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,318 @@
...
+    if (!this.httpClient) {
+      throw new Error('HttpClient is not defined.');
+    }
+  }
+
+  private createOptions(url: string): UpGradeClientInterfaces.IHttpClientWrapperRequestConfig {
+    const options: UpGradeClientInterfaces.IHttpClientWrapperRequestConfig = {
+      headers: {
+        'Content-Type': 'application/json',
+        'Session-Id': this.clientSessionId || uuid.v4(), // set this here? Do we need to require a clientSessionId?
```
VivekFitkariwala, line 93: Creating uuid here will create new id for each request.

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,318 @@
...
+    return options;
+  }
+
+  private async sendRequest<RequestBodyType>({ path, method, body }: IApiServiceRequestParams): Promise<any> {
+    this.validateClient();
+
+    const options = this.createOptions(path);
+
+    if (method === UpGradeClientEnums.REQUEST_METHOD.GET) {
+      return await this.httpClient.doGet(path, options);
```
VivekFitkariwala, line 122: Why have you added await here? If you are not handling the error you should just return the promise.

clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -0,0 +1,318 @@
...
+    site,
+    target,
+    condition,
+    status,
+    uniquifier,
+    clientError,
+  }: IMarkDecisionPointParams): Promise<UpGradeClientInterfaces.IMarkDecisionPoint> {
+    const assignment = this.dataService.findExperimentAssignmentBySiteAndTarget(site, target);
+
+    if (!assignment) {
```
VivekFitkariwala, line 231: We can mark a decision point without an experiment also. I don't think we should throw error here. Correct me if I am missing something here.

clientlibs/js/src/ApiService/ApiService.types.ts
```diff
@@ -0,0 +1,21 @@
+import { UpGradeClientEnums } from 'types';
+
+export interface IApiServiceRequestParams {
+  path: string;
+  method: UpGradeClientEnums.REQUEST_METHOD;
+  body?: any;
```
VivekFitkariwala, line 6: any can be replaced with
type JSONValue =
    | string
    | number
    | boolean
    | { [x: string]: JSONValue }
    | Array<JSONValue>;

clientlibs/js/src/UpGradeClient/UpgradeClient.ts
```diff
@@ -0,0 +1,481 @@
...
+   * ```
+   */
+
+  constructor(userId: string, hostUrl: string, context: string, options?: IConfigOptions) {
+    const config: UpGradeClientInterfaces.IConfig = {
+      apiVersion: 'v5',
+      userId: userId,
+      hostURL: hostUrl,
+      context: context,
+      clientSessionId: options?.clientSessionId,
```
VivekFitkariwala, line 78: Create a session id in case client doesn't provide one

clientlibs/js/src/UpGradeClient/UpgradeClient.ts
```diff
@@ -0,0 +1,481 @@
...
+   * }
+   *
+   * const upgradeClient: UpgradeClient[] = new UpgradeClient(hostURL, userId, context);
+   * const upgradeClient: UpgradeClient[] = new UpgradeClient(hostURL, userId, context, options);
+   * ```
+   */
+
+  constructor(userId: string, hostUrl: string, context: string, options?: IConfigOptions) {
+    const config: UpGradeClientInterfaces.IConfig = {
+      apiVersion: 'v5',
```
VivekFitkariwala, line 74: Instead of hardcoding you can use the major version from package.json. I like the way you used the variable defined during webpack build time.

clientlibs/js/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "scripts": {
     "build:bundler": "webpack",
-    "build:types": "./node_modules/.bin/dts-bundle-generator -o dist/browser/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/node/index.d.ts src/index.ts",
+    "build:types": "./node_modules/.bin/dts-bundle-generator -o dist/browser/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/node/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/browser-lite/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/node-lite/index.d.ts src/index.ts",
```
VivekFitkariwala, line 10: You should add multiple scripts for build here. You can go with nps or something simple like concurrently to manage them.

VivekFitkariwala (dismissed):
I have added some comments. Please go through them. Thanks

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: updated

clientlibs/js/src/UpGradeClient/UpgradeClient.ts
danoswaltCL: what would be the simplest way of doing that so that?

clientlibs/js/package.json
danoswaltCL: yeah it's out of hand, i can break it into parts

danoswaltCL: updated

danoswaltCL:
@VivekFitkariwala thx for the comments, I've made some of the requested changes and will visit the rest of the feedback tomorrow.

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: hm yeah let me make sure this works if I remove it, this is preexisting from the within-subjects PR

clientlibs/js/src/ApiService/ApiService.types.ts
danoswaltCL: oops good catch, I tried this but it makes some other things downstream show type errors but I'll find a type for this

clientlibs/js/src/DefaultHttpClient/DefaultHttpClient.ts
```diff
@@ -0,0 +1,106 @@
...
+          return await this.fetchData({
+            url,
+            method,
+            body,
+            options,
+            retries: retries - 1,
+            backOff: backOff * 2,
+          });
+        } else {
+          return responseData;
```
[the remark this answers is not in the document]
danoswaltCL: updated

clientlibs/js/src/DefaultHttpClient/DefaultHttpClient.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+      } else {
+        return responseData;
+      }
+    } catch (error) {
+      throw new Error(JSON.stringify(error));
```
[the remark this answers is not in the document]
danoswaltCL: we are handling the error, if you are referring to not handling the error above, that was a mistake and it's handled now. I think we still want this to catch all other errors that are unspecified.

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: nice, I guess I thought it needed to be there. I was able to remove a ton of async/awaits in other places also that use this call.

clientlibs/js/src/ApiService/ApiService.ts
danoswaltCL: I addressed other comments but still need to look into this one tomorrow, definitely seems like you're right but need to look back at the within-subjects work and see if it's something I messed up or if there was a pre-existing bug (or both).

danoswaltCL: so I think there might be a pre-existing issue here, I don't know how we're supposed to mark a decision point without an assignment (aka no experiment running) in the v5 `/mark` API. If that is indeed that case, can that be handled in a separate bugfix PR?

I think there may need to be additional API and client-lib changes if so, currently this code is not right but the original code change for within-subjects before I rearranged things can be seen here: (https://github.com/CarnegieLearningWeb/UpGrade/commit/43d20a7cc16acb47555c2d5e483c9fe23a7918cf#diff-a4a1d40bc005562c8bb6ebb760ded881ff488ebd3bd499ce37213c9457728d2eL30) ... it would throw an exception if assignment is undefined, but that should not stop us from marking.

danoswaltCL:
@VivekFitkariwala i've updated for most comments except for the issues pointed out for `mark` experiment, which I think may need resolved in a separate PR. please take another look, ty!

clientlibs/js/package.json
```diff
@@ -7,7 +7,11 @@
   ],
   "scripts": {
     "build:bundler": "webpack",
-    "build:types": "./node_modules/.bin/dts-bundle-generator -o dist/browser/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/node/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/browser-lite/index.d.ts src/index.ts && ./node_modules/.bin/dts-bundle-generator -o dist/node-lite/index.d.ts src/index.ts",
+    "build:types": "concurrently \"npm run build:types-browser\" \"npm run build:types-node\" \"npm run build:types-browser-lite\" \"npm run build:types-node-lite\"",
```
VivekFitkariwala, line 10: If you want to further optimise you can use this
`concurrently \"npm:build:types-*\"`

clientlibs/js/src/UpGradeClient/UpgradeClient.ts
VivekFitkariwala: It would be the same way as setting the value of USE_CUSTOM_HTTP_CLIENT and IS_BROWSER during the webpack build.
You will need to read the `version` from `package.json` and give it to the webpack.

clientlibs/js/src/DefaultHttpClient/DefaultHttpClient.ts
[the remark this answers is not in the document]
VivekFitkariwala: What I mean by not handling the error is that here you are catching the error and then throwing a new error but not doing anything else with that error.

clientlibs/js/src/ApiService/ApiService.ts
VivekFitkariwala: @danoswaltCL can you create a separate issue for this and we can track it there?

danoswaltCL: @VivekFitkariwala #999 and #1000 created.

clientlibs/js/package.json
danoswaltCL: implemented, works great

clientlibs/js/src/UpGradeClient/UpgradeClient.ts
danoswaltCL: implemented, it is now parsing the version from package.json in the webpack file and injecting via a webpack variable

danoswaltCL:
@VivekFitkariwala please review changes made to resolve your comments now that bugfix code is in. thx!

VivekFitkariwala (approved):

PR #987, opened 2023-08-28
Feature: Experiment Archive State

ppratikcr7:
This PR contains Archived status feature.


danoswaltCL:
> This PR contains Archived status feature.

Can we get some more technical overview of the changes that were made to aid review and know how to test? If there is any finalized document that goes over the actual requirements that were decided on after all the years of discussion, can someone point to it?

ppratikcr7:
> > This PR contains Archived status feature.
> 
> Can we get some more technical overview of the changes that were made to aid review and know how to test? If there is any finalized document that goes over the actual requirements that were decided on after all the years of discussion, can someone point to it?

Sure. So we do have a [document](https://docs.google.com/document/d/1X-DBcyz7nTwk_MLg8kYfWg1Uspb9bx0y9OEgi_T0tLI/edit#heading=h.1z8xhfd43vmo).

Requirements:

1. The archived experiments can't be unarchived or changed to a different status.
2. The experiment will be in read-only mode after being in archived state.
3. The logs wont get stored for archived experiments, even for the user earlier enrolled in it and had some logs.
4. The logs gets stored in a new Archived table as a backup.
5. While updating state to Archived state, we get a confirmation checkbox, until it's checked, the 'Save' button would be disable.
6. The archived experiments wont be shown on home dashboard experiment list view while the filter is 'All' or any other option except 'Status'. If we select 'Status' option, then we can filter all experiments including 'Archived' using a mat-auto complete search input box.

Just wanted to confirm one thing, we don't want to have archived state when the experiment is in enrolling state as per the diagram? And wait for experiment to be in enrolment Complete state.

213239091-10d9c46f-04d5-4d27-b08e-9f77299a42f9

bcb37 (commented):
I've gone over this and it all looks good to me.

ppratikcr7:
> I've gone over this and it all looks good to me.

Okay, you can also approve this and it would't auto merge as I have disabled the option as Dan had also pointed out in slack. We can wait for multiple approvals before merging if needed.

backend/packages/Upgrade/src/api/models/ArchivedStats.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Query } from './Query';
+
+@Entity()
+export class ArchivedStats extends BaseModel {
+  @PrimaryColumn('uuid')
+  public id: string;
+
+  @IsNotEmpty()
+  @Column('jsonb')
+  public result: any;
```
danoswaltCL, line 13: sry gotta be annoying about `any`, is there any type that can be defined for this? What does "result" mean?

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.ts
```diff
@@ -50,6 +53,16 @@ export class ExperimentListComponent implements OnInit, OnDestroy, AfterViewInit
   @ViewChild('searchInput') searchInput: ElementRef;
 
   @ViewChild(MatSort, { static: true }) sort: MatSort;
+  searchControl = new FormControl();
+
+  get filteredStatusOptions(): string[] {
+    if (this.searchValue !== undefined) {
```
danoswaltCL, line 59: we really only want searchValue to specifically be a string, so I think we'd be better to be explicit `if (typeof this.searchValue === 'string')`.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.ts
```diff
@@ -119,7 +143,11 @@ export class ExperimentListComponent implements OnInit, OnDestroy, AfterViewInit
   applyFilter(filterValue: string) {
     this.filterExperimentPredicate(this.selectedExperimentFilterOption);
     if (filterValue !== undefined) {
```
danoswaltCL, line 145: same as other comment, filterValue needs to be a string, not just 'not undefined', a null value can slip in here, for instance.

frontend/projects/upgrade/src/assets/i18n/en.json
```diff
@@ -267,6 +268,8 @@
   "home.change-experiment-status.enrolling-info.text": "*Once the experiment is started you won’t be allowed to update Experiment’s Condition and Sites",
   "home.change-experiment-status.enrolling-error.text": "*Experiment start date should not be greater than experiment end date {{ endDate }}",
   "home.change-experiment-status.condition-count-error.text": "*Please have at least 2 conditions to move forward",
+  "home.change-experiment-status.archiving-confirmation.text": "Please acknowledge that you agree to permanently transition this experiment to archive status",
```
danoswaltCL, line 271: probably should have period at the end.

danoswaltCL:
Just a few small comments from me, there's just the one spot in `experiment-list` with the filters that feels like it could be refactored to be simpler, but as I looked at it I didn't have any good quick suggestions. Chatgpt also didn't give me any good suggestions.

backend/packages/Upgrade/src/api/controllers/QueryController.ts
```diff
@@ -50,4 +50,40 @@ export class QueryController {
   ): Promise<any> {
     return this.queryService.analyze(dataLogParams.queryIds, request.logger);
   }
+
```
VivekFitkariwala, line 53: Why do we need a separate http endpoint for archive? I thing We should use the /analyse endpoint and server should get the data from the archive table in case the experiment is in archive state. Given that we have the same response type we can use the same endpoint. This will also keep the frontend dumb.

backend/packages/Upgrade/src/api/repositories/MetricRepository.ts
```diff
@@ -32,7 +33,9 @@ export class MetricRepository extends Repository<Metric> {
   public async findMetricsWithQueries(ids: string[]): Promise<Metric[]> {
```
VivekFitkariwala, line 33: Changes over here are unnecessary. Because the query does more than what the function name suggests.

frontend/projects/upgrade/src/app/core/analysis/store/analysis.actions.ts
```diff
@@ -33,7 +34,10 @@ export const actionUpsertMetricsSuccess = createAction(
 
 export const actionUpsertMetricsFailure = createAction('[Analysis] Upsert Metrics Failure');
 
-export const actionExecuteQuery = createAction('[Analysis] Execute Query', props<{ queryIds: string[] }>());
+export const actionExecuteQuery = createAction(
```
VivekFitkariwala, line 37: You don't need to send the state information from the frontend. You can get that from the queryId itself.

VivekFitkariwala (changes requested):
The frontend doesn't need to know that we are storing the query result of the archive state in a different way. We can use the same endpoint to serve the result.

RidhamShah:
@VivekFitkariwala  Yes, we can merge analysis and achieve branch by giving one more variable to identify which call is sent.

backend/packages/Upgrade/src/api/models/ArchivedStats.ts
RidhamShah: the result stores the query result in a JSON object. Will update this to appropriate type

frontend/projects/upgrade/src/app/core/analysis/store/analysis.actions.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/QueryController.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/models/ArchivedStats.ts
RidhamShah: code updated

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.ts
ppratikcr7: yeah this was a very old code which used undefined. Nice catch, will update. :)

backend/packages/Upgrade/src/api/services/QueryService.ts
```diff
@@ -66,6 +66,12 @@ export class QueryService {
 
     const promiseResult = await Promise.all(promiseArray);
     const experiments: Experiment[] = [];
+
+    // checks for archieve state experiment
+    if (promiseResult[0].experiment?.state === EXPERIMENT_STATE.ARCHIVED) {
+      return await this.getArchivedStats(queryIds, logger);
```
VivekFitkariwala, line 72: You don't need to `await` here. Just return the promise

VivekFitkariwala:
The changes look good.

backend/packages/Upgrade/src/api/services/QueryService.ts
ppratikcr7: @RidhamShah Can you please resolve this.

VivekFitkariwala (approved):

PR #989, opened 2023-08-31
Stratification Factor Feature

Yagnik56:
This PR contain all changes of adding stratification factors feature from front-end and back-end. This PR also contains backend integration test cases. We will create separate PR's for backend unit test cases and frontend test cases. The details of the tickets covered in this PR are mentioned in the [document](https://docs.google.com/document/d/1zMA2yeBcfn2rJBHzC2gAdpW9nS1fvOzu05ncasAsy_E/edit?usp=sharing).


danoswaltCL (approved):

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.ts
```diff
@@ -262,14 +264,16 @@ export class ExperimentOverviewComponent implements OnInit, OnDestroy {
     }
   }
 
-  findStratificationFactorId(stratificationFactorName: string) {
+  findStratificationFactorId(stratificationFactorName: string, assignmentAlgorithm) {
```
ppratikcr7, line 267: @Yagnik56 Please mention the type for `assignmentAlgorithm`.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
```diff
@@ -0,0 +1,240 @@
...
+    @Res() res: express.Response
+  ): Promise<any> {
+    if (!factor) {
+      return Promise.reject(new Error(SERVER_ERROR.MISSING_PARAMS + ' : stratification Factor should not be null.'));
+    }
+
+    const data = await this.stratificatonService.getCSVDataByFactor(factor, request.logger);
+
+    // Convert JSON data to CSV
+    const parser = new Parser();
```
VivekFitkariwala, line 128: These two lines should be moved inside the service.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
```diff
@@ -0,0 +1,240 @@
...
+    const csv = parser.parse(data);
+
+    // return csv file with appropriate headers to request;
+    res.setHeader('Content-Type', 'text/csv; charset=UTF-8');
+    res.setHeader('Content-Disposition', 'attachment; filename="data.csv"');
```
VivekFitkariwala, line 133: Instead of just data.csv you can also add factor name in the csv file name. Something like this -> data-factorName.csv.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
```diff
@@ -0,0 +1,240 @@
...
+   *              $ref: '#/definitions/UserStratificationFactor'
+   *        '401':
+   *          description: Authorization Required Error
+   *        '500':
+   *          description: Internal Server Error, Insert Error in database, CSV file is not valid
+   */
+  @Post()
+  @UseBefore(upload.single('file'))
+  public insertStratification(@Req() request: AppRequest): Promise<UserStratificationFactor[]> {
+    const csvData = request.body[0].file;
```
VivekFitkariwala, line 168: Parsing logic should be moved inside the service. Controller should only pass the data to the service.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
```diff
@@ -0,0 +1,240 @@
...
+   *          description: Authorization Required Error
+   *        '500':
+   *          description: Internal Server Error, FactorName is not valid
+   */
+  @Delete('/:factor')
+  public deleteStratification(
+    @Param('factor') stratificationFactor: string,
+    @Req() request: AppRequest
+  ): Promise<StratificationFactor> {
+    if (!stratificationFactor) {
```
VivekFitkariwala, line 235: Can you add class-validator decorator with the @Param('factor') for not empty?

backend/packages/Upgrade/src/api/models/Experiment.ts
```diff
@@ -120,6 +129,12 @@ export class Experiment extends BaseModel {
   })
   public filterMode: FILTER_MODE;
 
+  @ManyToOne(() => StratificationFactor, (stratificationFactor) => stratificationFactor.experiment, {
+    onDelete: 'CASCADE',
+    nullable: true,
+  })
+  public stratificationFactor: StratificationFactor;
```
VivekFitkariwala, line 136: This should be optional.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1522,6 +1534,7 @@ export class ExperimentAssignmentService {
       }
       await Promise.all(promiseArray);
     } else {
+      const enrollmentCounts = await this.getEnrollmentCountPerCondition(experiment, user.id);
```
VivekFitkariwala, line 1537: You should not derive the condition during mark experiment point because the enrollment count will be different during assignment and enrollment. Instead use the condition passed by the client.

backend/package.json
```diff
@@ -12,11 +12,13 @@
     },
     "license": "ISC",
     "dependencies": {
+        "dayjs": "^1.11.10",
+        "json2csv": "^6.0.0-alpha.2",
```
VivekFitkariwala, line 16: Use stable version.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
Yagnik56: can't find the right syntax

danoswaltCL:
The backend build is failing, please check (@Yagnik56?)

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
```diff
@@ -0,0 +1,202 @@
...
+   *          description: Factor name not found
+   *        '500':
+   *          description: Internal Server Error, FactorName is not valid
+   */
+  @Get('/download/:factor')
+  public async getStratificationByFactor(
+    @Param('factor') factor: string,
+    @Req() request: AppRequest,
+    @Res() res: express.Response
+  ): Promise<any> {
```
danoswaltCL, line 119: please define a response type (Promise<any>)

backend/packages/Upgrade/src/api/controllers/validators/StratificationValidator.ts
```diff
@@ -0,0 +1,24 @@
+import { IsNotEmpty, IsObject, IsOptional, IsString } from 'class-validator';
+
+export class FactorStrata {
+  @IsNotEmpty()
+  @IsString()
+  public factor: string;
+
+  @IsObject()
+  public value: Record<string, number>;
```
danoswaltCL, line 9: what is "value" here? will it always be a number?

backend/packages/Upgrade/src/api/controllers/validators/StratificationValidator.ts
```diff
@@ -0,0 +1,24 @@
...
+  @IsString()
+  public userId: string;
+
+  @IsNotEmpty()
+  @IsString()
+  public factor: string;
+
+  @IsOptional()
+  @IsString()
+  public value?: string;
```
danoswaltCL, line 23: above we say value can be a number, and it is required, here we have string, and it's optional. I imagine these are closely related. It doesn't seem like this can be optional though, can someone explain these new models with an example or two?

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1741,6 +1766,21 @@ export class ExperimentAssignmentService {
     return experimentalCondition;
   }
 
+  private assignStratifiedRandom(
+    sortedExperimentCondition: ExperimentCondition[],
+    originalWeights: number[],
+    enrollmentCount: any[]
```
danoswaltCL, line 1772: please give type instead of `any`

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -0,0 +1,233 @@
...
+import { StratificationFactorRepository } from '../repositories/StratificationFactorRepository';
+import { ErrorWithType } from '../errors/ErrorWithType';
+@Service()
+export class StratificationService {
+  constructor(
+    @OrmRepository()
+    private stratificationFactorRepository: StratificationFactorRepository
+  ) {}
+
+  private calculateStratificationResult(results: any[]): FactorStrata[] {
```
danoswaltCL, line 20: please give type instead of `any`

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -0,0 +1,233 @@
...
+      ])
+      .innerJoin('sf.userStratificationFactor', 'usf')
+      .where('sf.stratificationFactorName = :factor', { factor })
+      .groupBy('sf.stratificationFactorName, value')
+      .getRawMany();
+
+    return this.calculateStratificationResult(queryBuilder)[0];
+  }
+
+  public async getCSVDataByFactor(factor: string, logger: UpgradeLogger): Promise<any> {
```
danoswaltCL, line 84: please give type instead of `any`

backend/packages/Upgrade/test/integration/Experiment/stratification/RandomAlgoCheck.ts
```diff
@@ -0,0 +1,74 @@
...
+import { UpgradeLogger } from '../../../../src/lib/logger/UpgradeLogger';
+
+export default async function RamdomAlgoCheck(): Promise<void> {
+  const experimentService = Container.get<ExperimentService>(ExperimentService);
+  // experiment object
+  const experimentObject = stratificationRandomExperimentAssignmentExperiment2;
+  const userService = Container.get<UserService>(UserService);
+
+  // creating new user
+  const user = await userService.upsertUser(systemUser as any, new UpgradeLogger());
```
danoswaltCL, line 18: please give type instead of `any`

backend/packages/Upgrade/test/integration/Experiment/stratification/SRSAlgorithmCheck.ts
```diff
@@ -0,0 +1,36 @@
...
+export default async function AlgorithmCheck(): Promise<void> {
+  // Testing for Factorial Experiment with different Decision Point
+
+  const experimentService = Container.get<ExperimentService>(ExperimentService);
+  // experiment object
+  const experimentObject = stratificationSRSExperimentAssignmentExperiment1;
+  const userService = Container.get<UserService>(UserService);
+
+  // creating new user
+  const user = await userService.upsertUser(systemUser as any, new UpgradeLogger());
```
danoswaltCL, line 18: please give type instead of `any`

backend/packages/Upgrade/test/integration/Experiment/stratification/RandomAlgoCheck.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  // setting condition-2 weight as 100%
+  conditions.sort((a, b) => (a.order > b.order ? 1 : b.order > a.order ? -1 : 0));
+  const updatedConditions = conditions.map((condition) => {
+    return condition.order === 2 ? { ...condition, assignmentWeight: 100 } : { ...condition, assignmentWeight: 0 };
+  });
+
+  // create experiment
+  await experimentService.create(
+    { ...experimentObject, conditions: updatedConditions } as any,
```
danoswaltCL, line 31: please give type instead of `any`

backend/packages/Upgrade/test/integration/Experiment/stratification/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  expect(groupExclusions.length).toEqual(0);
+
+  const individualAssignments = await checkService.getAllIndividualAssignment();
+  expect(individualAssignments.length).toEqual(0);
+
+  const individualExclusions = await checkService.getAllIndividualExclusion();
+  expect(individualExclusions.length).toEqual(0);
+
+  // create users over here
+  await userService.create(experimentUsers as any, new UpgradeLogger());
```
danoswaltCL, line 34: please give type instead of `any`

backend/packages/Upgrade/test/unit/controllers/mocks/StratificationServiceMock.ts
```diff
@@ -0,0 +1,29 @@
...
+export default class StratificationServieMock {
+  public getAllStratification(logger: UpgradeLogger): Promise<[]> {
+    return Promise.resolve([]);
+  }
+
+  public getStratificationByFactor(factor: string, logger: UpgradeLogger): Promise<[]> {
+    return Promise.resolve([]);
+  }
+
+  public getCSVDataByFactor(factor: string, logger: UpgradeLogger): Promise<any> {
```
danoswaltCL, line 15: please give type instead of `any`

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+    private stratificationFactorsDataService: StratificationFactorsDataService
+  ) {}
+
+  fetchStratificationFactors$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(StratificationFactorsActions.actionFetchStratificationFactors),
+      withLatestFrom(this.store$.pipe(select(selectAllStratificationFactors))),
+      switchMap(() =>
+        this.stratificationFactorsDataService.fetchStratificationFactors().pipe(
+          map((data: any) =>
```
danoswaltCL, line 26: please give type instead of `any`

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  deleteStratificationFactor$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(StratificationFactorsActions.actionDeleteStratificationFactor),
+      map((action) => action.factor),
+      filter((factor) => !!factor),
+      switchMap((factor) =>
+        this.stratificationFactorsDataService.deleteStratificationFactor(factor).pipe(
+          map((data: any) => {
```
danoswaltCL, line 44: please give type instead of `any`

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.model.ts
```diff
@@ -0,0 +1,20 @@
...
+  values: Record<string, number>;
+}
+
+export interface StratificationFactorSimple {
+  factorName: string;
+}
+
+export interface StratificationFactorsState extends EntityState<StratificationFactor> {
+  isLoading: boolean;
+  totalStratificationFactors: any;
```
danoswaltCL, line 19: please give type instead of `any`

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/delete-stratification/delete-stratification.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  selector: 'app-delete-stratification',
+  templateUrl: './delete-stratification.component.html',
+  styleUrls: ['./delete-stratification.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DeleteStratificationComponent {
+  message: boolean;
+  constructor(
+    public dialogRef: MatDialogRef<DeleteStratificationComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: any
```
danoswaltCL, line 17: please give type instead of `any`

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/import-stratifications/import-stratifications.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  selector: 'app-import-stratifications',
+  templateUrl: './import-stratifications.component.html',
+  styleUrls: ['./import-stratifications.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ImportStratificationsComponent {
+  file: any;
+  stratificationInfo: any;
+  isStratificationCSVValid = true;
+  csvData: any = [];
```
danoswaltCL, line 18: please give type instead of `any` for all of these

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/import-stratifications/import-stratifications.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  constructor(
+    private stratificationFactorsService: StratificationFactorsService,
+    public dialogRef: MatDialogRef<ImportStratificationsComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: any
```
danoswaltCL, line 24: please give type instead of `any`

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.router.navigate(['/participants']);
```
danoswaltCL, line 45: Not sure how we've done it elsewhere, but just eyeballing this it feel like we're navigating away before we're really finished executing code; I think this will may work out ok, but I think it would probably be cleanest to dispatch a separate action for navigating once everything else has finished executing. `actionDeleteStratificationFactorSuccess` seems like a more natural place to do navigation.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+            const stratificationFactor = {
+              ...data[0],
+              factor: data[0].stratificationFactorName,
```
danoswaltCL, line 48: we're accessing a property on a potentially undefined or null object. TypeScript surprisingly won't throw any warning squiggles for array indices that are potentially undefined (https://github.com/microsoft/TypeScript/issues/13778), so we have to be careful.

I think it's probably unlikely that we'd get something undefined here, but we have  also defined this data object as `any` also, so we're flying a little blind with type safety in this function.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.download(data);
+            return StratificationFactorsActions.actionExportStratificationFactorSuccess();
+          }),
+          catchError(() => [StratificationFactorsActions.actionExportStratificationFactorFailure()])
+        )
+      )
+    )
+  );
+
+  private download(csvData) {
```
danoswaltCL, line 93: I know I've seen this kind of "download" function plunked inside the effects file elsewhere in our codebase, but I don't think we should repeat it. The ngrx effect pattern would be to dispatch an action that would prompt the actual component to fire this. Plus we're doing DOM manipulation, that's component business, this file should just be for ngrx effects.

Can you create a 'downloadStratificationCSVData' action, then subscribe to it in the actual component(s) where this action should be received?

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.model.ts
```diff
@@ -0,0 +1,20 @@
+import { AppState } from '../../core.state';
+import { EntityState } from '@ngrx/entity';
+
+export interface State extends AppState {
+  stratificationFactors: StratificationFactorsState;
+}
+
+export interface StratificationFactor {
```
danoswaltCL, line 8: is this the same object as the `FactorStrata` model in the backend? Can we use a consistent naming if so? That object has `value` and this has `values`, but I think it's supposed to be the same.

I have same question as with that model, is `Record<string, number>` the right type? Maybe I'm not understanding the function of this particular object. It would probably help if the word `value` was more descriptive, value could mean anything.

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.scss
```diff
@@ -0,0 +1,91 @@
...
+  padding: 0;
+  height: 100%;
+
+  .header {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: space-between;
+
+    button {
```
danoswaltCL, line 15: Just a note here that using a native element tag as a css selector is asking for trouble, can a class be used instead? I know it's happening in many places in codebase but it's an easy thing to do to avoid very annoying css debugging behavior.

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.allStratificationFactorsSub = this.stratificationFactorsService.allStratificationFactors$.subscribe(
+      (allStratificationFactors) => {
+        this.allStratificationFactors = allStratificationFactors;
+        this.stratificationFactorsForTable = this.convertToTableformat();
+      }
+    );
+  }
+
+  convertToTableformat() {
+    const stratificationFactors = this.allStratificationFactors.map((element) => {
```
danoswaltCL, line 35: instead of "element", can we say "stratificationFactor" so it's obvious?

I'd like to see a type defined for this "table format", it has an implied return type of 

```
{
    factor: string;
    summary: string;
}[]
```

Which feel like an important thing to define an explicit type for. Maybe something like:

```
interface StratificationFactorsTableRow {
    factor: string;
    summary: string;
}

type StratificationFactorsTable = StratificationFactorTableRow[];
```

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    return stratificationFactors;
+  }
+
+  openImportStratificationsDialog() {
+    const dialogRef = this.dialog.open(ImportStratificationsComponent, {
+      panelClass: 'import-stratification-modal',
+    });
+    dialogRef.afterClosed().subscribe((isImportButtonClicked) => {
+      if (isImportButtonClicked) {
+        setTimeout(() => {
```
danoswaltCL, line 58: what is the setTimeout for here in the afterClosed callbacks?

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -0,0 +1,90 @@
...
+  ngOnInit(): void {
+    this.allStratificationFactorsSub = this.stratificationFactorsService.allStratificationFactors$.subscribe(
+      (allStratificationFactors) => {
+        this.allStratificationFactors = allStratificationFactors;
+        this.stratificationFactorsForTable = this.convertToTableformat();
+      }
+    );
+  }
+
+  convertToTableformat() {
```
danoswaltCL, line 34: "convertToTableFormat" (capitalization)

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.ts
```diff
@@ -284,30 +313,54 @@ export class ExperimentOverviewComponent implements OnInit, OnDestroy {
...
-        context: [context],
-        tags,
-        logging,
-      };
-      this.emitExperimentDialogEvent.emit({
-        type: eventType,
-        formData: overviewFormData,
-        path: NewExperimentPaths.EXPERIMENT_OVERVIEW,
-      });
+      let stratificationFactorValueToSend = stratificationFactor;
```
danoswaltCL, line 321: This function is already overloaded, lines 321 through 336 seem like a good place to slice out the logic into its own function.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/modal/delete-segment/delete-segment.component.html
danoswaltCL, line 1: what was the delete-segment modal? why is it being deleted as a part of the stratification random PR?

Yagnik56: It wasn't used in any component, we are using common delete dialog for segments and experiments.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.ts
```diff
@@ -284,30 +313,54 @@ export class ExperimentOverviewComponent implements OnInit, OnDestroy {
...
+          } else {
+            this.stratificationFactorNotSelectedMsg = 'home.new-experiment.overview.stratification-factor.error.text';
+          }
+        } else {
+          this.isStratificationFactorSelected = true;
+        }
+      } else {
+        stratificationFactorValueToSend = null;
+      }
+      if (this.isStratificationFactorSelected) {
```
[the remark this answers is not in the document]
Yagnik56: we will show warning that either stratificationFactor isn't selected or if there are no stratificationFactor to choose from then warning to import a one before creating SRS experiment.

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
Yagnik56: subscriber was updating value before reducer even update it in case of import or delete so we have to reload the page every time to get updated strata-factor list that's why added extra delay there.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.model.ts
Yagnik56: resolved

Yagnik56: we are using `Record<string, number>` coz we don't have specific key names in factor like for `favourite_food` it can have pizza, burger, pie etc... while for graduated it can have yes, no with count for that value. other issue you mentioned here are resolved.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: @danoswaltCL Okay, so can we move the download logic and DOM manipulations in the service and call that instead?

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.model.ts
danoswaltCL: ok i see, this number here literally for the _counts_, ok

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
danoswaltCL: I'm gonna take a deeper look at this when I get back in tomorrow, I want to give you a good recommendation. The reason this looks extra strange to me is that we're not really using the success and failure import / export actions for anything. it doesn't touch the reducer or the state, we're just firing a DOM event and we don't do anything with the logic in the effect, so as it is currently written, it doesn't even make sense to dispatch an action, the download action could happen in the component by itself.

backend/packages/Upgrade/src/api/controllers/validators/StratificationValidator.ts
RidhamShah: `Value` is the object containing all factorValues and the number of its occurrence for a `stratificationFactor`.
Ex:
``` 
{
  factor: FavoriteFood,
  value: {
    burger: 5,
    pizza: 2,
  }
}
```

backend/packages/Upgrade/src/api/controllers/validators/StratificationValidator.ts
RidhamShah: This value is renamed `stratificationFactorValue` which gets stored in DB.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: We have been navigating in effects at all other places as well. If we want to move it to actions, we will have to do it for all.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/modal/delete-segment/delete-segment.component.html
danoswaltCL: yeah that's why I asked, I assumed it wasn't really related but I couldn't tell. separate PRs are ideal but it's okay, getting rid of dead code appreciated.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
danoswaltCL: navigating in the effects is fine, I'm just asking for that to happen last, when we're executing "success" actions; it's confusing to navigate away before we're finished executing other code.

ppratikcr7: Okay, got it

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: Okay, let's keep it as it is for now as we discussed on slack. We can later create a separate enhancement PR to hadle this for all download related components at once.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: So actually we are returning an array of object from the backend and `data[0]` refers to the first object, which is the deleted row object. I don't think this can be a null object, still if you think, we can add a conditional check. If we have `data.stratificationFactorName`, then `stratificationFactorName` wont exist on object type.
Screenshot 2023-10-19 at 11 08 11 AM

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
VivekFitkariwala: Can you try @Params decorator it has a validation option. The only drawback is that you will get an object of all the params.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: @danoswaltCL So, I would update the other component's navigations in a separate PR. Hope this is what you want:

Screenshot 2023-10-19 at 12 30 23 PM

ppratikcr7: @danoswaltCL Are we good here?

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: @danoswaltCL I have added the conditional check for this.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  deleteStratificationFactor$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(StratificationFactorsActions.actionDeleteStratificationFactor),
+      map((action) => action.factor),
+      filter((factor) => !!factor),
+      switchMap((factor) =>
+        this.stratificationFactorsDataService.deleteStratificationFactor(factor).pipe(
+          map((data) => {
+            if (data[0]) {
```
danoswaltCL, line 46: If `data` comes back successfully but `data[0]` is unexpectedly falsey, what should happen? I think this would return whatever this unexpected `data` data is and not navigate anywhere, which is probably not wanted.

I guess what should happen would be same as the `catchError` scenarios, `StratificationFactorsActions.actionDeleteStratificationFactorFailure()`, but does failure also need to do `router.navigate`?

Yagnik56: Giving data a type there will resolve the first point you mentioned right? 
Regarding navigation don't you think success or failure in both cases we should route back to stratification tab and not at home page?

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  private calculateStratificationResult(
+    results: { factor: string; value: string; count: string; experimentIds: string[] }[]
+  ): FactorStrata[] {
+    const formattedResults = results.reduce((formatted, result) => {
+      const { factor, value, count, experimentIds } = result;
+      if (!formatted[factor]) {
+        formatted[factor] = { factor, factorValue: {}, experimentIds };
+      }
+      if (value !== 'N/A') {
```
danoswaltCL, line 28: will this 'N/A' string for sure come this way? and never like `n/a` or `n / a` or whatever?

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
danoswaltCL: what i'm saying is that this is an `if` block without an else. So in the case where this is "successful" response from the db call but `data[0]` is null or undefined (such as if the `data` is an empty array), there is no `else` block to handle what to do.

danoswaltCL (commented):
I think we're close, I think I've given this as much a review as I can and most of the comments have been resolved, ty guys, just one or two questions left open.

backend/packages/Upgrade/src/api/controllers/StratificationController.ts
danoswaltCL: is this resolved?

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
danoswaltCL: I tried running the unit tests for `deleteStratificationFactor$` to see if this was covered, but they are not really working; try making them fail, they will keep passing. I'll put the reason why in the comments for tests.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+      });
+
+      service.deleteStratificationFactor$.subscribe((result) => {
+        expect(result).toEqual(expectedAction);
+        expect(router.navigate).toHaveBeenCalledWith(['/participants']);
+      });
+
+      actions$.next(
+        StratificationFactorsActions.actionDeleteStratificationFactor({ factor: { ...mockData[0] }.factor })
+      );
```
danoswaltCL, line 82: Firstly, thank you for writing unit tests, that's great. I found a weird thing here though, I found that this test is not actually running the expectations in the `subscribe` block. If I change `expectedAction` to null or `/participants` to something else, it still passes.

This is because the observable never emits. I'd think that this would fail the test, but apparently it does not, which is very confusing but here's what I found is happening:

The reason the observable never emits is that you're passing in the data param like this when dispatching `actionDeleteStratificationFactor`:

```
{ factor: { ...mockData[0] }.factor }
```

Which makes no sense, mockData is an object, so`.factor` is undefined here, and thus it is getting filtered out. you want `{ factor: mockData.factor}`.

When you pass the data in that way, the test will actually run the expect code in the `subscribe` block.

The reason I came into the unit tests in the first place is on the comment elsewhere about this, we should be able to use the unit tests here to test what happens when `data[0]` is undefined. When I test this, the navigation is not called and the action dispatched is `undefined`.

```
it('this should do something if the data coming back is unexpected or empty array, but it will neither navigate nor dispatch anything', fakeAsync(() => {
      stratificationFactorsDataService.deleteStratificationFactor = jest.fn().mockReturnValue(of([])); // <---return empty array from mock API response to simulate "success" but unexpected data.

      const expectedAction = StratificationFactorsActions.actionDeleteStratificationFactorSuccess({
        stratificationFactor: { ...mockData },
      });

      service.deleteStratificationFactor$.subscribe((result) => {
        expect(result).toEqual(expectedAction);
        expect(router.navigate).toHaveBeenCalledWith(['/participants']);
      });

      actions$.next(StratificationFactorsActions.actionDeleteStratificationFactor({ factor: mockData.factor }));

      tick(0);
    }));
```

backend/packages/Upgrade/src/api/services/StratificationService.ts
amurphy-cl: Good point about this string. I'd suggest if we make a default string value it should be just `NA`. We need to make sure this is described in the import UI and handled well for MVP.

ppratikcr7: This `if` condition should be removed. It's not used now.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.spec.ts
ppratikcr7: Okay, got it. Will add failure actions and update the backend to return the object instead of array of object.

frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
ppratikcr7: Okay, got it. Will resolve this.

danoswaltCL (approved):

PR #1003, opened 2023-09-12
Bugfix/level payload type error

Yagnik56:
Issue was that payload for `levelelements` in condition was expected as `{type: payload type, value: string}` but it was sending it as `string`, we were converting payload from `string` to  `{type: payload type, value: string}` when we go next or save the data but because `conditionTable` was updating live on every change we perform in `factorTable` it was reading it directly from form as `string`. I have changed format in form only to `{type: payload type, value: string}` format so this issue is resolved.


danoswaltCL:
@Yagnik56 please open this PR against release/5.0 branch as the base instead, it should go into version 5

ppratikcr7:
@Yagnik56 You will need to create another PR as the dev branch latest commits are pulled in this PR as old destination branch was dev. Later we can close this PR.

PR #1011, opened 2023-09-14
Fix/issue-1000 Updated mark call related changes

RidhamShah:
Fix/Issue-1000
Remove Assignment if statement which throws Assignment when not found, instead send the Assignment object with no condition details. This way, the user can mark this Assignment.

Fix/Issue-999
Remove unnecessary experimentType property from Assign & Mark call of V5


danoswaltCL (dismissed):
@RidhamShah looks right to me but changes need to made for v4 mark validator / controller as well

danoswaltCL:
v4 and v5 look like they should have `payload` removed as a parameter in `data.assignedCondition` object, it causes an error in v4 library mark calls but the mark call in the controllers doesn't even utilize this property as far as I can tell.

PR #1014, opened 2023-09-15
make 'status' a string in Java mark requests

bcb37:


danoswaltCL:
@bcb37 looks good to me, I checked in release/v5 branch and it seems to need be fixed there also, can you make this change against the release/v5 branch instead, and we'll merge it down to dev?

PR #1025, opened 2023-09-25
allow null in type for condition

danoswaltCL:


clientlibs/js/src/UpgradeClient.ts
```diff
@@ -302,7 +302,7 @@ export default class UpgradeClient {
 
   async markExperimentPoint(
     site: string,
-    condition: string = null,
+    condition: string | null = null,
```
bcb37, line 305: I don't think we need explicit `null` here. We can test for truthiness of this parameter and then make the API with `null` if there's no value.

danoswaltCL: the user should explicitly send in null or a string, we want them to knowingly decide to indicate no condition with `null` and not undefined, which could be accidental. If the type is just `string`, typescript will complain in development if code implementing our library wants to pass in a nullable value, which should be allowed.

Using `null` as the default in this context to treat the param as optional is pointless because it's logically impossible for a condition value to not get sent in when `status` is after it in index order and required.

bcb37 (approved):

PR #1032, opened 2023-10-03
Added payload migration file

RidhamShah:
Some unnecessary details of `Payload` data are created while creating a new Migration file, so adding a migration file of such at first so no other migration files will get affected.


ppratikcr7:
@danoswaltCL This is kind of dummy migration file which is renaming the "condition_payload_type_enum" to "condition_payload_type_enum_old" and few level and condition_payload changes, that are just to satisfy some typeorm constraints. We get these migrations in all our future migration files which are ot required and are already kind of present in the dev branch DB schema. We can directly push this file to dev for now to satisfy the typeorm constraints, so that we are able to create a clean migration file for the SRS PR.

VivekFitkariwala (approved):

PR #1037, opened 2023-10-09
Stratification factor usage status feature

ppratikcr7:
This PR has enhancements to the major stratification factor PR, where we add a status column in stratification table, to show if the stratification factor is being used in any experiment or not. Hovering the statuses will show a tooltip to view the list of experiments where the stratification factor is currently being used. We will disable the delete trash icon for used stratification factors.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/StratificationService.ts
+++ b/backend/packages/Upgrade/src/api/services/StratificationService.ts
@@ -50,22 +50,7 @@ export class StratificationService {
       .groupBy('sf.stratificationFactorName, value')
       .getRawMany();
 
-    const allStratificationFactors = await this.stratificationFactorRepository.find();
-    const remainingFactors = allStratificationFactors
-      .filter((factor) => {
-        return !queryBuilder.some((result) => result.factor === factor.stratificationFactorName);
-      })
-      .map((factors) => {
-        return {
-          factor: factors.stratificationFactorName,
-          value: 'N/A',
-          count: 0,
-          experimetIds: factors.experiment[0].id,
-        };
-      });
-
-    const temp = this.calculateStratificationResult([...queryBuilder, ...remainingFactors]);
-    return temp;
+    return this.calculateStratificationResult(queryBuilder);
   }
 
   public async getStratificationByFactor(factor: string, logger: UpgradeLogger): Promise<FactorStrata> {
--- a/backend/packages/Upgrade/src/api/services/StratificationService.ts
+++ b/backend/packages/Upgrade/src/api/services/StratificationService.ts
@@ -18,9 +18,9 @@ export class StratificationService {
 
   private calculateStratificationResult(results: any[]): FactorStrata[] {
     const formattedResults = results.reduce((formatted, result) => {
-      const { factor, value, count } = result;
+      const { factor, value, count, experimentIds } = result;
       if (!formatted[factor]) {
-        formatted[factor] = { factor, values: {} };
+        formatted[factor] = { factor, values: {}, experimentIds};
       }
       if (value !== 'N/A') {
         formatted[factor].values[value] = parseInt(count);
@@ -40,13 +40,15 @@ export class StratificationService {
         'sf.stratificationFactorName AS factor',
         `COALESCE(usf.stratificationFactorValue, 'N/A') AS value`,
         'COUNT(*) AS count',
+        'ARRAY_AGG(experiments.id) AS "experimentIds"', // Aggregate experiment ids into an array
       ])
       .innerJoin('sf.userStratificationFactor', 'usf')
+      .leftJoin('sf.experiment', 'experiments') // Left join with the Experiment entity
       .groupBy('sf.stratificationFactorName, value')
       .getRawMany();
 
     const allStratificaitonFactors = await this.stratificationFactorRepository.find();
-
+    console.log('allStratificaitonFactors: ', allStratificaitonFactors);
     const remainingFactors = allStratificaitonFactors
       .filter((factor) => {
         return !queryBuilder.some((result) => result.factor === factor.stratificationFactorName);
@@ -56,10 +58,13 @@ export class StratificationService {
           factor: factors.stratificationFactorName,
           value: 'N/A',
           count: 0,
+          experimetIds: factors.experiment[0].id,
         };
       });
 
-    return this.calculateStratificationResult([...queryBuilder, ...remainingFactors]);
+    const temp = this.calculateStratificationResult([...queryBuilder, ...remainingFactors]);
+    console.log('temp: ', temp);
+    return temp;
   }
 
   public async getStratificationByFactor(factor: string, logger: UpgradeLogger): Promise<FactorStrata> {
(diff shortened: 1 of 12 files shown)


backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -40,13 +40,15 @@ export class StratificationService {
...
+        'ARRAY_AGG(experiments.id) AS "experimentIds"', // Aggregate experiment ids into an array
       ])
       .innerJoin('sf.userStratificationFactor', 'usf')
+      .leftJoin('sf.experiment', 'experiments') // Left join with the Experiment entity
       .groupBy('sf.stratificationFactorName, value')
       .getRawMany();
 
     const allStratificaitonFactors = await this.stratificationFactorRepository.find();
-
+    console.log('allStratificaitonFactors: ', allStratificaitonFactors);
```
VivekFitkariwala, line 51: Remove this

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -56,10 +58,13 @@ export class StratificationService {
           factor: factors.stratificationFactorName,
           value: 'N/A',
           count: 0,
+          experimetIds: factors.experiment[0].id,
```
VivekFitkariwala, line 61: Why returning only single experimentId here?

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -56,10 +58,13 @@ export class StratificationService {
...
         };
       });
 
-    return this.calculateStratificationResult([...queryBuilder, ...remainingFactors]);
+    const temp = this.calculateStratificationResult([...queryBuilder, ...remainingFactors]);
```
VivekFitkariwala, line 65: Refoactor this

backend/packages/Upgrade/src/api/services/StratificationService.ts
ppratikcr7: This is part of unused code now, I will remove the remainingFactors variable, its not used now.

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -90,6 +105,14 @@ export class StratificationComponent implements OnInit {
     });
   }
 
+  getExperimentIdsTooltip(experimentIds: any[]): string {
+    return 'Experiment IDs: [' + experimentIds.join(', ') + ']';
+  }
+
+  checkStratificationFactorUsageStatus(status: string): boolean {
+    return status === STRATIFICATION_FACTOR_STATUS.USED ? true : false;
```
danoswaltCL, line 113: ternary isn't necessary, `status === STRATIFICATION_FACTOR_STATUS.USED` returns a boolean, leave off the `? true : false`

backend/packages/Upgrade/test/integration/Experiment/withinSubject/RandomAlgoCheck.ts
```diff
@@ -68,5 +68,6 @@ export default async function RamdomAlgoCheck(): Promise<void> {
     new UpgradeLogger(),
     context
   );
+  console.log("experimentConditionAssignments[0].assignedCondition: ", experimentConditionAssignments[0].assignedCondition)
```
danoswaltCL, line 71: remove console log

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -48,8 +52,19 @@ export class StratificationComponent implements OnInit {
             : key + '=' + stratificationFactor.factorValue[key];
           totalUsers += stratificationFactor.factorValue[key];
         });
+
         factorSummary = totalUsers.toString() + ' ' + factorSummary + ' (' + tempSummary + ')';
-        return { factor: stratificationFactor.factor, summary: factorSummary };
+        const status =
```
danoswaltCL, line 57: I think the enums are not useful here? we just need a boolean. I think we've got some things we can smooth out here:

The intent of `stratificationFactor.experimentIds[0] !== null` looks to be an existence check for at least one id, which will fail if it's `undefined` for instance, which seems possible. `!stratificationFactor.experimentIds[0]` would be safer to catch all falsey values, or even better is probably `stratificationFactor?.experimentIds.some((experimentId) => experimentId);` to truly capture what you're really wanting: if any member of experimentIds is truthy, then the "status" is true.

I think you can eliminate these enums and change `checkStratificationFactorUsageStatus`, they are only used in the frontend and only here. Then you also don't need to call `checkStratificationFactorUsageStatus` in the html, you can just check the boolean.

Given the above, since `status` is really at the end of the day a boolean indicator, it would be more clear if we called this `isUsed` on the `StratificationFactorTableRows` type interface. `status: true` would not be obvious if you didn't already know what it meant.

In summary, here are the suggestions (I have not checked this, i'm just coding in the comment box, this is just a suggestion):

```
interface StratificationFactorsTableRow {
  factor: string;
  summary: string;
  isUsed: boolean;
  experimentIds: string[];
}
```

```
checkStratificationFactorUsageStatus(experimentIds: string[]) {
   if (Array.isArray(experimentIds) {
       return experimentIds.some(id => id);
   }
   
   return false;
}
```

return {
          factor: stratificationFactor.factor,
          summary: factorSummary,
          status: this.checkStratificationFactorUsageStatus(stratificationFactor.experimentIds);
          experimentIds: stratificationFactor.experimentIds,
        };
```

**Review by VivekFitkariwala — APPROVED** [review]

File: frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts (hunk 6 above)
Comment (ppratikcr7) [reply frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts:57]: We already use a similar enum for segment usage, so keeping it in sync. Should I go ahead and make changes as per your suggestion, as the enum has kind of 2 values only, which can definitely handled with a true/false boolean.

Screenshot 2023-10-23 at 1 23 36 PM

Comment (danoswaltCL) [reply frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts:57]: If it's only the two possible states, which are basically on/off, I think it would make sense to do it as a boolean, would be simpler to me. What do the 'locked' and 'unlocked' states of the segments mean? I don't think I've seen that.

Comment (ppratikcr7) [reply frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts:57]: Okay, will have boolean for this. The segment 'locked' and 'unlocked' states was mainly to lock the edit mode but used/unused enums do that now. So, will remove that part as well in a separate PR.

Comment (ppratikcr7) [reply frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts]: @danoswaltCL Done. Can you please review again?

File: types/src/Experiment/enums.ts [hunk 7]
```diff
@@ -170,6 +170,11 @@ export enum SEGMENT_STATUS {
   UNLOCKED = 'Unlocked',
 }
 
+export enum STRATIFICATION_FACTOR_STATUS {
```
danoswaltCL, line 173: @ppratikcr7 i think this just needs removed now if it's now unused right? Looks good to me otherwise, can approve and merge after that

ppratikcr7: Done

danoswaltCL:
@ppratikcr7 looks like frontend spec test is still referencing the enums that were removed. I approve, feel free to merge this if you'd like once the build completes. 👍

Issue #1047, opened 2023-10-09
Update Java retry handling (no 4xx retries, retry others once)


danoswaltCL:
This is in response to investigating prod issues. This is an attempt to reduce the number of meaningless retries that may occur when Mathia gets a bad response from our API, which is potentially causing a number of side-effects that are worse than the benefit of retries.

The "fix" will be to adjust to only retry once instead of 3 times, and 4xx status code errors will not be retried at all, as they represent a class of errors with the request from the client that should not be likely to succeed if retried. (aka bad request or 404).

Please see my debut novel: https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1696815544069419?thread_ts=1696815538.859709&cid=C055CGCRP6Z

Note: The fix will need to be merged down/cherry-picked into all published Java versions, so each 3x, 4x, and current 5x library should get this "patch".

PR #1048, opened 2023-10-09
no retries on 4xx, max_retries=1

danoswaltCL:
#1047

This will need to get merged down into release/v4.1 then into release/v5.0, with new Java libs published for each.


danoswaltCL:
@jerith666 tagging you on this for awareness. Java lib is throwing retries without any strategy, which is sometimes generating a lot of extra meaningless error traffic and all the side-effects of it. This was spawned out of difficultly triaging the prod issues from last week, I am putting up a few changes in general to tamp down on snowballing error-state that our system sometimes gets into. I think this could be a major help.

jerith666 (approved):
Yep, should help.

PR #1057, opened 2023-10-11
sub-segments of used segment will show as used in segments list

Yagnik56:
This PR will resolve the issue of sub-segments of used segments not showing as used.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/SegmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/SegmentService.ts
@@ -110,6 +110,24 @@ export class SegmentService {
         });
       }
 
+      segmentsData.forEach((segment) => {
+        segmentsUsedLockedList.forEach((usedSegment) => {
+          if (segment.id === usedSegment) {
+            segment.subSegments.forEach((subSegment) => {
+              segmentsUsedLockedList.push(subSegment.id);
+            });
+          }
+        });
+
+        segmentsUsedUnlockedList.forEach((usedSegment) => {
+          if (segment.id === usedSegment) {
+            segment.subSegments.forEach((subSegment) => {
+              segmentsUsedUnlockedList.push(subSegment.id);
+            });
+          }
+        });
+      });
+
       const segmentsDataWithStatus = segmentsData.map((segment) => {
         if (segment.id === globalExcludeSegment.id) {
           return { ...segment, status: SEGMENT_STATUS.GLOBAL };


backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -110,6 +110,24 @@ export class SegmentService {
         });
       }
 
+      segmentsData.forEach((segment) => {
+        segmentsUsedLockedList.forEach((usedSegment) => {
+          if (segment.id === usedSegment) {
```
bcb37, line 115: We can replace this third-level loop with something like:          `segmentsUsedLockedList.push(...segment.subSegments.map(subSegment => subSegment.id))`
and also below for the unlocked list.

bcb37 (commented):
One suggestion

backend/packages/Upgrade/src/api/services/SegmentService.ts
VivekFitkariwala: the second level loop can be replaced with an if condition
`if(segmentUsedList.includes(segment.id))`

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -85,38 +85,35 @@ export class SegmentService {
...
+      });
+
       const segmentsDataWithStatus = segmentsData.map((segment) => {
         if (segment.id === globalExcludeSegment.id) {
           return { ...segment, status: SEGMENT_STATUS.GLOBAL };
-        } else if (segmentsUsedLockedList.find((segmentId) => segmentId === segment.id)) {
-          return { ...segment, status: SEGMENT_STATUS.USED }; // TODO change to locked
-        } else if (segmentsUsedUnlockedList.find((segmentId) => segmentId === segment.id)) {
-          return { ...segment, status: SEGMENT_STATUS.USED }; // TODO change to unlocked
+        } else if (segmentsUsedList.find((segmentId) => segmentId === segment.id)) {
```
VivekFitkariwala, line 115: Use `includes` over here also.

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -85,38 +85,35 @@ export class SegmentService {
       const allExperimentSegmentsInclusion = await this.getExperimentSegmenInclusionData();
       const allExperimentSegmentsExclusion = await this.getExperimentSegmenExclusionData();
```
VivekFitkariwala, line 86: These three sequential awaits should be executed parallelly.

VivekFitkariwala (changes requested):
I have added some comments

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: resolved

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: resolved

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: resolved

VivekFitkariwala (approved):

PR #1062, opened 2023-10-13
Moving data and query in within subject code block for latency issue in QA

ppratikcr7:
This PR is a hotfix for the latency issues observed in current QA (v5.0.9) due to the `monitoredLogCounts` query and its required sites and targets data used outside within subjects block, causing unnecessary calls to the query for non-within subjects experiments.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -650,21 +650,23 @@ export class ExperimentAssignmentService {
           );
         })
       );
-
-      const allSites = [],
-        allTargets = [];
-      filteredExperiments.forEach((exp) => {
-        exp.partitions.forEach((partition) => {
-          allSites.push(partition.site);
-          allTargets.push(partition.target);
+      let monitoredLogCounts: any[];
+      if (filteredExperiments.some((e) => e.assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS)) {
+        const allWithinSubjectsSites = [],
+          allWithinSubjectsTargets = [];
+        filteredExperiments.forEach((exp) => {
+          exp.partitions.forEach((partition) => {
+            allWithinSubjectsSites.push(partition.site);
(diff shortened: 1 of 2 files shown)


ppratikcr7:
@danoswaltCL @bcb37 We don't see any possibility for optimization of the below query as it uses only single left join. The hotfix in this PR should suffice for the current latency issue. In case we still observe this, we can look for some more possibilities.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -650,21 +650,23 @@ export class ExperimentAssignmentService {
...
         })
       );
-
-      const allSites = [],
-        allTargets = [];
-      filteredExperiments.forEach((exp) => {
-        exp.partitions.forEach((partition) => {
-          allSites.push(partition.site);
-          allTargets.push(partition.target);
+      let monitoredLogCounts: any[];
```
danoswaltCL, line 653: it would be safer to initialize this as an empty array, as later we call `.find` on it. It probably is not going to be undefined there, but why risk it. no point in giving type of it's `any`.

`let monitoredLogCounts = [];`

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -650,21 +650,23 @@ export class ExperimentAssignmentService {
...
+      if (filteredExperiments.some((e) => e.assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS)) {
+        const allWithinSubjectsSites = [],
```
danoswaltCL, line 655: small thing, but since I'm already asking for a change, for consistency let's avoid chaining declarations with commas:

```
const allWithinSubjectsSites = [];
const allWithinSubjectsTargets = [];
```

danoswaltCL (approved):

ppratikcr7:
@danoswaltCL I was going through within subjects experiment testing for exclusions logic and noticed that the within subject unitOfAssignmennt was removed from the UI in your last commit here. When do we plan to get this back in UI?

Issue #1078, opened 2023-10-25
Implement caching in getAllExperimentConditions to reduce latency


ppratikcr7:
**Is your feature request related to a problem? Please describe.**
Implement caching in getAllExperimentConditions function for the large queries to get valid experiments data.

**Describe the solution you'd like**
We should cache the result of the queries so that it can be used in future from the cache and save response time.

**Describe alternatives you've considered**
Apart from this we are also going to split the large queries into smaller chunks to get better latency.

Issue #909, opened 2023-07-05
Optimize query for post rule 


amurphy-cl:
**Describe the bug**
Completed experiments are still providing network traffic by querying the post-experiment rule. We noticed that this throttled UpGrade 4.x performance in prod, so the temporary solution was to delete the completed experiments. In the future, we'll want to archive rather than delete many if not most completed experiments (queries are not conducted on archived experiments). One solution for optimization is to disable queries when the post rule is set to "default".

danoswaltCL:
I have had trouble getting Playpower usable production db code to try and duplicate this, so I'm putting this issue on the board for "cycle 2" to try and create a reproducible example for this that can be shared.

danoswaltCL:
@ppratikcr7 I was able to load the prod clone sql file into empty local db and then start up backend services and use it normally, I don't know if the user 'readonly' error really matters?

I tried briefly to recreate the bug under locust load tests but I think it will require some more thought about how to reproduce, I will need to get used to how to modify these locust python scripts to get what we want.

I believe what might work is to find a stable baseline under load-testing with all experiments enrolling, then try setting experiments to "enrollment-complete" and see if anything starts to change in the response-times. That is about as much as I know about this bug at this point, I wish I had the exact data from when it happened, will try to catch it in the act if it happens again.

ppratikcr7:
> 

@danoswaltCL Hey, thanks for checking. I will try again to import it. For me, after taking about an hour or so, the error showed up and there was no data in the empty db I had created.

Yeah, right. We need to try load testing scripts as per above. Do let me know in case you need help with the script changes.

amurphy-cl:
Reopening this so we can validate in prod, tentatively scheduled for 01/02/2024

PR #1083, opened 2023-10-25
🐞  Hotfix/Split getValidExperiment queries into 3 sub queries to reduce latency against Release V5

ppratikcr7:
This PR contains the following changes:
1.  Splitting the large query in getValidExperiments() and getValidExperimentsWithPreview() called within the `assign` api call for release V5. Please refer the load testing [results](https://docs.google.com/document/d/14F1XxB6QifEW-q9XhqVgWg7Pgb0qJbnqlNg6TxgGxNE/edit?usp=sharing).
2. Caching the getValidExperiments(). We are not caching getValidExperimentsWithPreview().
3. Change to exclude users before fetching the experiments in `getAllExperimentConditions`.


amurphy-cl:
Please put #909 and #1078 into separate PRs. 
cc @mswartzCL

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -94,6 +126,14 @@ export class ExperimentRepository extends Repository<Experiment> {
         })
       );
 
+    const result3 = await query3.getMany();
```
VivekFitkariwala, line 129: You can run all the 3 queries in parallel using `Promise.all`

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -120,11 +160,7 @@ export class ExperimentRepository extends Repository<Experiment> {
         })
       );
 
-    const [experimentData, experimentSegmentData] = await Promise.all([
-      experiment.getMany().catch((errorMsg: any) => {
-        const errorMsgString = repositoryError('ExperimentRepository', 'find', {}, errorMsg);
-        throw errorMsgString;
-      }),
+    const [experimentSegmentData] = await Promise.all([
```
VivekFitkariwala, line 163: Run all the queries here

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -468,7 +468,7 @@ export class ExperimentAssignmentService {
       ]);
     } else {
       [experiments, [userExcluded, groupExcluded]] = await Promise.all([
-        this.experimentRepository.getValidExperiments(context),
+        this.experimentService.getCachedValidExperiments(context),
         this.checkUserOrGroupIsGloballyExcluded(experimentUser),
```
VivekFitkariwala, line 472: `this.checkUserOrGroupIsGloballyExcluded(experimentUser)`
Can you move this before fetching the valid experiment and end the request if the user is excluded? And because this is cached doing query serially won't impact the performance.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -437,6 +459,7 @@ export class ExperimentService {
     user: User,
     logger: UpgradeLogger
   ): Promise<ExperimentDTO[]> {
+    this.cacheService.resetCache();
```
VivekFitkariwala, line 462: You should not reset the whole cache instead find the keys with the pattern and reset only those.

VivekFitkariwala (dismissed):
Some comments need to be addressed before it can be merged.

VivekFitkariwala:
Replace `getValidExperiments` inside the `markExperimentPoint` function with the cached query `getCachedValidExperiments`.

ppratikcr7:
> Please put #909 and #1078 into separate PRs. cc @mswartzCL

We are not handline #909 in this PR. It would be in a separate PR. This PR contains changes #1078 (caching) and #1089 (optimizing getValidExperiment query for assign call)

backend/packages/Upgrade/src/api/services/ExperimentService.ts
RidhamShah: code updated

amurphy-cl:
My error, I linked to the wrong issue, but the sentiment is the same -- we don't really know whether it is caching, the getValidExperiment query, something else, or multiple issues that are causing bottlenecks in production. While it's great to have quick attention to the issues, we need to be methodical about doing fixes that we have good evidence that are the root cause of the problems, and put (even small!) fixes in separate PRs -- that way we can address them separately. For instance, if there is a problem with the caching fix but not the optimization one, we don't have the reject the entire PR. Thanks!

ppratikcr7:
> My error, I linked to the wrong issue, but the sentiment is the same -- we don't really know whether it is caching, the getValidExperiment query, something else, or multiple issues that are causing bottlenecks in production. While it's great to have quick attention to the issues, we need to be methodical about doing fixes that we have good evidence that are the root cause of the problems, and put (even small!) fixes in separate PRs -- that way we can address them separately. For instance, if there is a problem with the caching fix but not the optimization one, we don't have the reject the entire PR. Thanks!

@amurphy-cl Got it. We discussed this in today's dev sync-up call. We can cherry-pick the single commit for splitting of query or create a PR from the same. One more way is to setting the `CACHING_ENABLED` environment variable to false so that caching wont be effective during testing.

@danoswaltCL Let us know whatever you feel is okay, so that if a separate PR is needed, will create one.

danoswaltCL:
@ppratikcr7 @amurphy-cl yes we can work with this for testing given that we can isolate it.

ppratikcr7:
@danoswaltCL @bcb37 This one is now ready to be reviewed.

backend/packages/Upgrade/src/api/services/CacheService.ts
```diff
@@ -17,25 +17,32 @@ export class CacheService {
...
-    return this.memoryCache.del(id);
+  public delCache(id: string): Promise<void> {
+    return this.memoryCache ? this.memoryCache.del(id) : Promise.resolve();
   }
 
-  public resetCache(): void {
-    this.memoryCache.reset();
+  public async resetPrefixCache(prefix: string): Promise<void> {
+    const keys = this.memoryCache ? await this.memoryCache.store.keys() : [];
+    const filteredKeys = keys.filter((str) => str.startsWith(prefix));
```
ppratikcr7, line 33: @RidhamShah Do we need this filteredKeys? It is just getting logged in next line. If not required, you can remove this and the console log as well.

backend/packages/Upgrade/src/api/services/CacheService.ts
```diff
@@ -17,25 +17,32 @@ export class CacheService {
...
+    console.log(filteredKeys);
```
ppratikcr7, line 34: @RidhamShah Remove this console.log() if not required.

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -196,17 +269,31 @@ export class ExperimentRepository extends Repository<Experiment> {
...
+        const errorMsgString = repositoryError('ExperimentRepository', 'getValidExperiments-experimentConditionLevelPayloadQuery', {}, errorMsg);
         throw errorMsgString;
       }),
-      experimentSegment.getMany().catch((errorMsg: any) => {
-        const errorMsgString = repositoryError('ExperimentRepository', 'find', {}, errorMsg);
+      experimentFactorPartitionLevelPayloadQuery.getMany().catch((errorMsg: any) => {
+        const errorMsgString = repositoryError('ExperimentRepository', 'getValidExperiments-experimentFactorPartitionLevelPayloadQuery', {}, errorMsg);
+        throw errorMsgString;
+      }),
+      experimentMetricQuery.getMany().catch((errorMsg: any) => {
```
danoswaltCL, line 281: why do we need to get experiment metrics data during assign or mark calls where this is used

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -196,17 +269,31 @@ export class ExperimentRepository extends Repository<Experiment> {
...
+        const errorMsgString = repositoryError('ExperimentRepository', 'getValidExperiments-experimentMetricQuery', {}, errorMsg);
+        throw errorMsgString;
+      }),
+      experimentSegmentQuery.getMany().catch((errorMsg: any) => {
+        const errorMsgString = repositoryError('ExperimentRepository', 'getValidExperiments-experimentSegmentQuery', {}, errorMsg);
         throw errorMsgString;
       }),
     ]);
 
+    const experimentData = experimentConditionLevelPayloadData.map((data) => {
```
danoswaltCL, line 291: this seems like this nested array could be problematic if these arrays are long, maybe it's not important because if it will get cached, but just wanted to think it through. I'm seeing a lot of places where factorial assignments would be doing a lot of nested looping while processing factor / level / payload data.

backend/packages/Upgrade/src/api/services/CacheService.ts
RidhamShah: solved this

backend/packages/Upgrade/src/api/services/CacheService.ts
RidhamShah: removed this

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -458,19 +458,19 @@ export class ExperimentAssignmentService {
...
+      // return null if the user or group is excluded from the experiment
+      return [];
+    }
+
     if (previewUser) {
-      [experiments, [userExcluded, groupExcluded]] = await Promise.all([
-        this.experimentRepository.getValidExperimentsWithPreview(context),
-        this.checkUserOrGroupIsGloballyExcluded(experimentUser),
-      ]);
+      [experiments] = await Promise.all([this.experimentRepository.getValidExperimentsWithPreview(context)]);
```
danoswaltCL, line 471: These don't need to be a `Promise.all` array if there's only one promise now.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -458,19 +458,19 @@ export class ExperimentAssignmentService {
...
     // query all experiment and sub experiment
     // check if user or group is excluded
-    let experiments: Experiment[] = [],
-      userExcluded: string,
-      groupExcluded: string[];
+    let experiments: Experiment[] = [];
+
+    const [userExcluded, groupExcluded] = await this.checkUserOrGroupIsGloballyExcluded(experimentUser);
+
+    if (userExcluded || groupExcluded.length > 0) {
```
danoswaltCL, line 465: This change to exclude before fetching the experiments makes sense, but it was never mentioned that it would be part of this hotfix as far as I can recall. It's probably fine, but please mention all changes in the comments, this is a significant change to test for that I don't recall discussing making within the hotfix branch.

backend/packages/Upgrade/src/api/services/CacheService.ts
```diff
@@ -17,25 +17,32 @@ export class CacheService {
...
+  public delCache(id: string): Promise<void> {
+    return this.memoryCache ? this.memoryCache.del(id) : Promise.resolve();
   }
 
-  public resetCache(): void {
-    this.memoryCache.reset();
+  public async resetPrefixCache(prefix: string): Promise<void> {
+    const keys = this.memoryCache ? await this.memoryCache.store.keys() : [];
+    const filteredKeys = keys.filter((str) => str.startsWith(prefix));
+    this.memoryCache.store.del(filteredKeys);
```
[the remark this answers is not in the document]
danoswaltCL: I need a list of situations to test to make sure cache gets invalidated as expected for both valid experiments and segments caching.

- On experiment creation
- On experiment delete
- On experiment modify

- On segment creation
- On segment delete
- On segment modify

Are these the situations to check? Any others that are handled differently but would still need cache cleared?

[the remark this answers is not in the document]
VivekFitkariwala: @danoswaltCL Earlier only segments were getting cached and resetting the cache would have not caused problems. But now we have two different types of cache so they need to be invalidated in a proper way. We should add a test case for cache invalidation. I suggest a separate PR for that.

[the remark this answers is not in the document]
VivekFitkariwala: @danoswaltCL the above list looks good
CRUD for experiments (State change for experiment)
CRUD for segments

backend/packages/Upgrade/src/api/services/CacheService.ts
```diff
@@ -17,25 +17,32 @@ export class CacheService {
...
   }
 
-  public async wrapFunction<T>(keys: string[], functionToCall: () => Promise<T[]>): Promise<T[]> {
+  // Use this to wrap the function that you want to cache
+  public wrap<T>(key: string, fn: () => Promise<T>): Promise<T> {
+    return this.memoryCache ? this.memoryCache.wrap(key, fn) : fn();
+  }
+
+  public async wrapFunction<T>(prefix, keys: string[], functionToCall: () => Promise<T[]>): Promise<T[]> {
```
VivekFitkariwala, line 42: Add type for prefix

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -283,6 +305,7 @@ export class ExperimentService {
     }
     return getConnection().transaction(async (transactionalEntityManager) => {
       const experiment = await this.findOne(experimentId, logger);
+      this.cacheService.delCache(CACHE_PREFIX.EXPERIMENT_KEY_PREFIX + experiment.context[0]);
```
VivekFitkariwala, line 308: You should await as this returns a promise

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -311,7 +311,7 @@ export class SegmentService {
       }
 
       // reset caching
-      this.cacheService.resetCache();
+      this.cacheService.resetPrefixCache(CACHE_PREFIX.SEGMENT_KEY_PREFIX);
```
VivekFitkariwala, line 314: Await here

VivekFitkariwala, line 314: Await here

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
ppratikcr7: You are correct, this was not discussed and we found this improvement while discussion an issue in our scrum meet and thought of rectifying this in the current hotfix, as the same code section was being updated. We will keep this in mind for future. I will add this in PR description.

backend/packages/Upgrade/src/api/services/CacheService.ts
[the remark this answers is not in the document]
RidhamShah: @danoswaltCL In this portion of the code, I was supposed to clear the cached data. However, during testing, I mistakenly left in `console.log` statements instead of the deletion code. This commit rectifies this oversight. Yes, the commit message may not be entirely clear, rest assured that the code has been refined and is ready to go.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/services/CacheService.ts
RidhamShah: code updated

danoswaltCL (approved):

PR #1091, opened 2023-10-30
1090 - Fixing npm ci command

VivekFitkariwala:
Fixing npm ci command

Added override for conflicting dependencies and removed `legacy-peer-deps` command


danoswaltCL:
@bcb37 since you put in the --legacy-peer-deps flag when node 18 was updated, I wondered if you could review this one; it looks fine to me.

backend/packages/Upgrade/package.json
```diff
@@ -115,5 +114,9 @@
         "uuid": "^3.3.3",
         "winston": "^3.2.1",
         "winston-splunk-httplogger": "^3.1.0"
+    },
+    "overrides": {
+        "class-validator": "^0.14.0",
+        "typeorm": "^0.2.20"
     }
```
bcb37, line 121: Can we use the references to our specified library versions, so that the overrides don't get out of sync when we upgrade them?

```
"overrides": {
        "class-validator": "$class-validator",
        "typeorm": "$typeorm"
}
```

bcb37 (changes requested):
Looks good; just one thing

bcb37 (approved):

PR #1099, opened 2023-11-03
fix loop in log call

bcb37:
Refactored nested loops that set later elements to `null`.


VivekFitkariwala (approved):

VivekFitkariwala:
PR looks good. But I am not sure if this will fix the issue it is associated with because earlier we had this line to filter out the null logs
`if (logData !== null) {`
Were you able to recreate the issue?

bcb37:
> PR looks good. But I am not sure if this will fix the issue it is associated with because earlier we had this line to filter out the null logs `if (logData !== null) {` Were you able to recreate the issue?

Yeah, I was able to recreate it. The test for null was only in the outer loop - I included the unit test to demonstrate an example of input that would cause the null value to create the error.

PR #1127, opened 2023-11-16
🐛 Fix for archive state not working in QA testing #30

ppratikcr7:
This PR contains the below fixes:

1. Now, the `Save` button on the experiment status change modal will be disabled on unchecking the archived status confirmation checkbox again.
2. Fixes the error when we used to change the experiment status to archived state. This was in case of experiments not containing any metric queries. So, handled that case for analyze() function.


danoswaltCL (approved):

danoswaltCL:
Re-QA: Zack

zackcl:
QA: Passed

PR #1129, opened 2023-11-16
Fix QueryFailedError when sorting experiments by status

### LINKED ISSUE #1128: Sorting experiments by status causes "QueryFailedError"
5.1.0

**Describe the bug**
Clicking the "Status" column header in the Experiment list page to sort the experiments by status results in a "QueryFailedError: function lower(experiment_state_enum) does not exist" error in the backend.

**To Reproduce**
1. Go to the Experiment list page (need to have at least 2 experiments)
2. Click on the 'Status' column header to sort the experiments
3. See the backend console for the "QueryFailedError" error (Now, you might need to log out and log in again)

**Expected behavior**
Sorting the experiments by status should not cause the error.

**Screenshots**
Screenshot 2023-11-16 at 4 07 03 PM

zackcl:
Resolves #1128

This issue was caused by applying the `LOWER()` function to an enum type when sorting the experiments by status. We can fix this by casting the enum type to a text type with `LOWER(CAST(experiment.${sortParams.key} AS TEXT))`.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentService.ts
@@ -177,7 +177,10 @@ export class ExperimentService {
       .whereInIds(expIds);
 
     if (sortParams) {
-      queryBuilderToReturn = queryBuilderToReturn.addOrderBy(`LOWER(experiment.${sortParams.key})`, sortParams.sortAs);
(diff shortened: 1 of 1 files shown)


zackcl:
@ppratikcr7 Please check if this PR looks correct and feel free to approve it.

ppratikcr7 (approved):

ppratikcr7:
QA: Done. Working as expected.

Issue #1141, opened 2023-11-21
Export include/exclude csv files for segmentation


ppratikcr7:
**Is your feature request related to a problem? Please describe.**
Currently the UpGrade UI allows for include/exclude members to be exported on a by-individual basis or as a json file. Since we often have large lists (in csv format) of members that we want to import, we need an export function to more easily export list members in a format that most users will have on-hand. 

**Describe the solution you'd like**
The ability to export a csv file, perhaps with fixed headers, in the Segments tab for either include or exclude lists.

bcb37:
QA: @zackcl

zackcl:
@amurphy-cl @danoswaltCL Currently, the format of the expected CSV file looks like the following:

```
UUID,Type
student1,individual
segment1,segment
school1,schoolId
```

Do you think this format is OK, or would you prefer the two columns to be swapped to match the table in the UI?
<img width="750" alt="Screenshot 2024-01-18 at 8 15 21 AM" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/8f0b712d-de3a-428e-893f-87b5585cec4a">

Currently, the type 'individual' and 'segment' are exported in lowercase, but they appear capitalized in the UI. I believe they should also be capitalized when exported. Additionally, we should check the case sensitivity during import. Let me know your thoughts on this.

zackcl:
Currently, when importing from a CSV, the list is appended to the existing ones. It's also possible to import members that are already present in the table as I tested.
<img width="750" alt="Screenshot 2024-01-18 at 8 29 40 AM" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/5ebb73e1-8123-4adc-86ed-10a97190111c">


@amurphy-cl, do you think appending the list to the existing ones is better than replacing them? I think it might be less confusing to allow CSV imports only when the table is empty. However, there could be use cases for importing/upserting CSVs on top of existing data. Let me know your thoughts on this.

zackcl:
Here is the summary of what we discussed today:

1. The order of the `UUID` and `Type` columns needs to be swapped to match the order shown in the table UI.
2. It's acceptable to have "individual" and "segment" in lowercase when exported, and not to check case sensitivity during import. However, ideally, we would like all these types to be capitalized in UpGrade, which we can discuss later.
[...]

PR #1142, opened 2023-11-21
Import segment csv feature

Yagnik56:
This PR add Import segment csv feature, moving validation of importing segment to backend and related testcase changes.


backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
...
+        const segmentInfo = JSON.parse(segment.fileContent);
+        const userIds = segmentInfo.individualForSegment.map((individual) =>
+          individual.userId ? individual.userId : null
+        );
+        const subSegmentIds = segmentInfo.subSegments.map((subSegment) => (subSegment.id ? subSegment.id : null));
+        const groups = segmentInfo.groupForSegment.map((group) => {
+          return group.type && group.groupId ? { type: group.type, groupId: group.groupId } : null;
+        });
+
+        const segmentTemp = { ...segmentInfo, userIds: userIds, subSegmentIds: subSegmentIds, groups: groups };
```
RidhamShah, line 182: Change variable name

ppratikcr7:
@Yagnik56 When I click on the link to download the sample segment csv, it redirects me to the home page. I guess we should still stay on the same segment import modal page.

Screenshot 2023-11-22 at 9 22 22 PM

@danoswaltCL Hope thats the correct route?

frontend/projects/upgrade/src/assets/i18n/en.json
```diff
@@ -399,6 +399,8 @@
   "segments.global-members.segments-count-members-error.text": "Please have at least 1 valid member to move forward",
   "segments.import-segment.text": "IMPORT SEGMENT",
   "segments.import-segmemt.message.text": "Select the JSON to import segment:",
+  "segments.import-segmemt-csv.message.text": "Select the CSV to import segment:",
+  "segments.download-segmemt-csv-sample.message.text": "(Download the CSV import segment format.)",
```
ppratikcr7, line 403: @Yagnik56 Please remove the full stop at the end of the text.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/modal/import-segment/import-segment.component.html
```diff
@@ -2,31 +2,25 @@
...
-          </mat-cell>
-        </ng-container>
-
-        <mat-header-row *matHeaderRowDef="displayedColumns; sticky: true"></mat-header-row>
-        <mat-row *matRowDef="let row; columns: displayedColumns"></mat-row>
-      </mat-table>
+    <ng-container>
+      <mat-slide-toggle class="toggle-button" color="primary" [(ngModel)]="toggleValue" (change)="clearSegmentArray()">Import CSV</mat-slide-toggle>
+      <div *ngIf="toggleValue">
+        <a href="#" (click)="downloadSampleCSV()">
```
ppratikcr7, line 8: This href routes to home page?

Yagnik56:
> @Yagnik56 I tried importing the downloaded sample segment csv, but it throws the below error:
> 
> Screenshot 2023-11-22 at 9 25 56 PM
> Can you please look into this?



> @Yagnik56 I tried importing the downloaded sample segment csv, but it throws the below error:
> 
> Screenshot 2023-11-22 at 9 25 56 PM
> Can you please look into this?

coz you don't have segment named Subseg1, you have to create it before you import this CSV file.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/modal/import-segment/import-segment.component.html
Yagnik56: probably I will change that,

ppratikcr7:
> > @Yagnik56 I tried importing the downloaded sample segment csv, but it throws the below error:
> > Screenshot 2023-11-22 at 9 25 56 PM
> > Can you please look into this?
> 
> coz you don't have segment named Subseg1, you have to create it before you import this CSV file.

Interesting. So, what if someone tries to directly import this file for checking the import functionality. Maybe, we should create the subsegment as well. @danoswaltCL What do you think?

Yagnik56:
> 
> Interesting. So, what if someone tries to directly import this file for checking the import functionality. Maybe, we should create the subsegment as well. @danoswaltCL What do you think?

I don't think it will be good idea, everyone might not have mathia as context and group1 and group2 as their groups. We can't create them right? This should be used as format reference only isn't it.

danoswaltCL:
If I'm understanding, I think that we should simply show a notification to the user that they can't import a csv that references non-existent segment ids.

So when exporting a list that does contain sub-segments, the user could also be notified that they will need to also export those referenced sub-segment lists if they want to be able to import the same list later?

@amurphy-cl

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
     return await this.segmentRepository.deleteSegment(id, logger);
   }
 
-  public async importSegments(segments: SegmentInputValidator[], logger: UpgradeLogger): Promise<Segment[]> {
+  public async importSegments(segments: SegmentFile[], logger: UpgradeLogger): Promise<SegmentReturnObj> {
+    const importFileErrors: SegmentImportError[] = [];
+    if (this.isJSONString(segments[0].fileContent)) {
```
VivekFitkariwala, line 169: You are checking only the first segment and that also without the confirmation that the `segments` length is greater than zero

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
...
+      const parsedJsonData: SegmentInputValidator[] = [];
+      segments.forEach((segment) => {
+        const fileName = segment.fileName;
```
VivekFitkariwala, line 172: This part of the code is written with an assumption that the values you are expecting will be inside the JSON. What if these values are not present? like `fileName`, `fileContent` etc.

backend/packages/Upgrade/src/api/services/SegmentService.ts
VivekFitkariwala: Can you change the code flow to this
```
for each segment
 if type json
  parseJson
 else
  parseCsv
importSegment
```

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
...
+    try {
+      JSON.parse(data);
+      return true;
+    } catch (error) {
+      return false;
+    }
+  }
+
+  validateSegmentJSON(segment: SegmentInputValidator): boolean {
+    const segmentSchema: Record<keyof any, string> = {
```
VivekFitkariwala, line 217: Use class validator here.

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
...
+        }
+
+        const memberType = rowValues[1].trim().toLowerCase();
+        if (memberType === 'individual') {
+          segmentUserIds.push(id);
+        } else if (memberType === 'segment') {
+          subSegmentIds.push(id);
+        } else if (memberType === 'context') {
+          segmentContext = id;
+        } else {
```
VivekFitkariwala, line 281: You should check the `memberType` of `group` before pushing it into the `segmentGroups`

VivekFitkariwala: Don't we need any validation for the data coming from CSV? Like context should exist and other?

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -149,7 +164,134 @@ export class SegmentService {
...
+        name: fileNames[index],
+        context: segmentContext,
+        type: SEGMENT_TYPE.PUBLIC,
+        userIds: segmentUserIds,
+        groups: segmentGroups,
+        subSegmentIds: subSegmentIds,
+      };
+    });
+
+    return this.importJsonSegments(segmentData, logger);
```
VivekFitkariwala, line 296: Rename this function to `importSegment`

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -160,8 +302,20 @@ export class SegmentService {
     });
     const allSegmentsData = await this.getSegmentByIds(allSegmentIds);
     const duplicateSegmentsIds = allSegmentsData?.map((segment) => segment?.id);
+    const contextMetaData = env.initialization.contextMetadata;
+    const contextMetaDataOptions = Object.keys(contextMetaData);
 
     for (const segment of segments) {
```
VivekFitkariwala, line 313: This is not purely importing the segments but also doing validation like context should exist and all the sub-segments should exist before the actual import. These errors should be sent back to the user.

frontend/projects/upgrade/src/app/core/segments/store/segments.effects.spec.ts
```diff
@@ -12,6 +12,7 @@ describe('SegmentsEffects', () => {
   let actions$: ActionsSubject;
   let segmentsDataService: any;
   let router: any;
+  let notificationService: any;
```
VivekFitkariwala, line 15: Don't use any if possible

VivekFitkariwala (changes requested):
The core logic of import need changes

backend/packages/Upgrade/src/api/controllers/SegmentController.ts
```diff
@@ -363,9 +363,9 @@ export class SegmentController {
    */
   @Post('/import')
   public importSegments(
-    @Body({ validate: false }) segments: SegmentInputValidator[],
+    @Body({ validate: false }) segments: SegmentFile[],
```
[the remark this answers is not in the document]
Yagnik56: no idea, @RidhamShah @danoswaltCL ?

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: we already have func `importSegments` which we call in container, don't you think this will be confusing?

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: these validations are added in importJsonSegment as it didn't had these so wrong context or group name segments were getting created.

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: we won't throw error if wrong group name is added, we decided to just ignore that row and create that segment although we throw error on wrong subsegment so I think we should have same approach for both. Which one do you suggest?

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: We are converting data we sent from frontend to this format before sending.

backend/packages/Upgrade/src/api/controllers/SegmentController.ts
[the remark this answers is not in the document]
RidhamShah: We can now test this validator for both JSON and CSV objects and should make it true.

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: changed

zackcl:
Here are the 2 things that need to be updated in this PR based on our discussion today  (issue: https://github.com/CarnegieLearningWeb/UpGrade/issues/1141):

1. The order of the `UUID` and `Type` columns needs to be swapped to match the order shown in the table UI.
2. The "Import CSV" button should only be enabled if the table is empty or contains only blank rows; otherwise, the button should be greyed out using the color `var(--light-blue)`.

backend/packages/Upgrade/src/api/controllers/validators/SegmentInputValidator.ts
```diff
@@ -46,3 +47,18 @@ export class SegmentInputValidator {
   @IsString({ each: true })
   public subSegmentIds: string[];
 }
+
```
VivekFitkariwala, line 50: Why add interfaces in the validator file?

backend/packages/Upgrade/src/api/controllers/validators/SegmentInputValidator.ts
```diff
@@ -46,3 +47,18 @@ export class SegmentInputValidator {
...
+  segments: Segment[];
+  importErrors: SegmentImportError[];
+}
+
+export interface SegmentImportError {
+  fileName: string;
+  error: string;
+}
+
+export interface SegmentFile {
```
VivekFitkariwala, line 61: If you are using this as a validator then why an interface?

VivekFitkariwala (changes requested):

VivekFitkariwala (approved):

Issue #1137, opened 2023-11-20
Bug (Enhancement?): Shared URLs should not always redirect to /home


danoswaltCL:
We should be able to share and open up routed links, such as an experiment page.

Currently we always redirect to the /home page.

Current Behavior:
1. Click into an experiment detail page, which will look like `/home/detail/<experiment-id>`.
2. Copy that url
3. Log out (or open up firefox or an ingognito tab that has no current session data for upgrade)
4. Paste that url back into browser (as if you received this link from someone else but were not logged in)
5. Log in

You will be redirected to `/home`.

Expected / Desiered behavior

- The user should be navigated back to the page indicated by the original url, not `/home`.

danoswaltCL:
QA: Pratik

PR #1143, opened 2023-11-21
Original url destinations like experiment detail page will not redirect to home after login

danoswaltCL:
#1137

Most routes will redirect to `/home` if you click on a shared url such as an experiment detail page like `/home/detail/abc123...` because the original destination url gets lost after navigating to `/login`.

This should allow one to open up a shared link, login if needed, and return to the intended destination page.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/meta-reducers/clear-state.reducer.spec.ts
+++ b/frontend/projects/upgrade/src/app/core/meta-reducers/clear-state.reducer.spec.ts
@@ -1,5 +1,5 @@
 import { createReducer } from '@ngrx/store';
-import { actionLogoutSuccess } from '../auth/store/auth.actions';
+import { actionLogoutSuccess, actionLoginStart } from '../auth/store/auth.actions';
 import { LocalStorageService } from '../local-storage/local-storage.service';
 import { ThemeOptions } from '../settings/store/settings.model';
 import { clearState } from './clear-state.reducer';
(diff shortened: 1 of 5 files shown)


frontend/projects/upgrade/src/app/core/meta-reducers/clear-state.reducer.spec.ts
```diff
@@ -10,6 +10,13 @@ describe('clearState', () => {
     settings: {
       theme: ThemeOptions.DARK_THEME,
     },
+    auth: {
+      isLoggedIn: false,
+      isAuthenticating: false,
+      user: null,
+      googleCredential: null,
+      redirectUrl: '/home',
+    },
```
bcb37, line 19: Do we need to set any attributes (besides `redirectUrl`) here, to test that they get set to the default values when `clearState` is called? The other two attributes in 'settings', for instance, aren't set here. If we do set them, it might be a more definitive test if they're initially set to different values.

frontend/projects/upgrade/src/app/core/meta-reducers/clear-state.reducer.spec.ts
```diff
@@ -25,6 +32,13 @@ describe('clearState', () => {
         toCheckAuth: null,
         toFilterMetric: null,
       },
+      auth: {
```
bcb37, line 35: Should the 'it' string read 'should reset settings and auth state...'?

frontend/projects/upgrade/src/app/core/meta-reducers/clear-state.reducer.spec.ts
danoswaltCL: i updated the specs to make this a little more explicitly tested

bcb37 (approved):

PR #1145, opened 2023-11-21
Bugfix/1136 google auth resets after refresh

danoswaltCL:
This solves for production environment always requiring a login every time you refresh open a link (aka when AUTH_CHECK is on for the UI, as it is in production but not lower environments.)

Funnily, this should also solve the production errors with "Authorization required" that get spit out as noise in backend because the wrong action was being dispatched. It doesn't hurt anything but it causes some pointless calls to fire too early. Now they won't.


ppratikcr7:
@danoswaltCL I checked in your branch and I found that it seems to kind of resolve the re-login issue on refresh. But, when I tried refreshing a few more times (I guess after 3 times), the login seems to be gone. I no longer see the user icon on left bottom and no import or add experiments options on home screen.
Screenshot 2023-11-24 at 4 17 59 PM

danoswaltCL:
@ppratikcr7 weird, okay I'll see if what's up

danoswaltCL:
@ppratikcr7 in your local, were you able to login at all on my branch? (sign out, then try logging in)?

danoswaltCL:
I do see this happening now, with AUTH_CHECK false, so I'll need to fix that. If you have AUTH_CHECK as true, and the GOOGLE_CLIENT_ID is filled out correctly in the env file (to match the one in environment.ts for the frontend dev build), it should work as expected?

ppratikcr7:
> @ppratikcr7 in your local, were you able to login at all on my branch? (sign out, then try logging in)?

@danoswaltCL Yes, I was able to sign out and login again in your branch.

danoswaltCL:
@ppratikcr7 should be fixed now, found that it also happened when on a routed detail page like the segments page and then click refresh, the profile name will disappear every time.

Please re-check, I made some changes to handle the login sequence in the ngrx effects in a little cleaner way, there was some out-of-order thing happening.

ppratikcr7 (approved):

Issue #1120, opened 2023-11-13
500 error when importing duplicate segment to associate with another context: need to support segments associated with multiple app contexts


amurphy-cl:
**Version where bug was found:**
5.x

**Describe the bug**
An existing segment is currently being used by the `assign-prog` app context. I want to use the same segment in another app context (`mathia-student-ui`). Currently segments are only associated with a single app context, so I planned to export the original segment, then import and associate it with the new context. I exported the original segment, but when importing the new segment, got a 500 error on import. 

**Expected behavior**
We need to be able to either support associating individual segments with multiple contexts, or be able to create or import a "duplicate" segment to then associate it with a new context. 

**Screenshots**
<img width="814" alt="image" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/65311369/93a5c498-8ddb-407f-9217-9cbab5a8bccc">

danoswaltCL:
@amurphy-cl We need a little clarification on what the requirement should be: should these be separate segments in the list per context? Or should we allow multiple contexts per segment?

amurphy-cl:
I think the idea of multiple contexts per segment seems more efficient, at least as displayed as such in the UI. Something to think about might be how displaying this info might change (or not change) in the UI refresh.

danoswaltCL:
Code review: Dan

PR #1150, opened 2023-11-29
🐞 Allow multiple contexts for duplicate segments and make mat-chip rounded

ppratikcr7:
This PR contains changes to handle duplicate segment import for multiple contexts. (#1120). We would still throw network call error with duplicate segment error message in case the duplicate segment has the same context. 

There is also a very small frontend change to show rounded centered mat-chips for contexts in segment list. (Earlier it was left aligned and rectangular)


ppratikcr7:
@danoswaltCL Can we merge this?

Issue #967, opened 2023-08-11
[UI] Add POST Imported CSV data functionality (import button click)


danoswaltCL:
**User Story**
As an experimenter, I want to send a csv of stratification data to backend via UI

**Dependencies**:
[🟠 API Creation story ]
#958
#966

**Acceptance Criteria**
GIVEN: I've selected a valid csv from my computer
AND: I've clicked on the import button
THEN: I should see a POST request to [🟠 add actual endpoint] with a request body [🟠 add request body details for csv import]

---

GIVEN: The network request is pending
THEN: I should not be able to click the import button again
AND: I should not be able to navigate away from the page while pending
AND: I should see some kind of spinner (🟠 figma?)

[Success and error handling could be a separate story]

---

GIVEN: The request is successful:
THEN: [ 🟠 What ? The show a "success" message, close modal, refetch the csv data so the table is refreshed with the new row?]

---

GIVEN: The request errors:
THEN: [ 🟠 What should the error messaging look like? What scenarios should be considered for error-states on this call? Do we close the modal or leave it open for trying again?]

zackcl:
The feature works fine in terms of making requests to the backend, but the success or error snackbar is not displayed when a CSV file is imported. This needs a discussion. (Related to https://github.com/CarnegieLearningWeb/UpGrade/issues/974#issuecomment-1820146209))

<br>

If I import a CSV file of the following content:
```
UUID,frl
stud1,true
stud2,false
stud3,true
stud4,false
stud5,false
stud6,false
```

The client makes POST request to `/api/stratification` with the following body:
```
[
    {
        "file": "UUID,frl\nstud1,true\nstud2,false\nstud3,true\nstud4,false\nstud5,false\nstud6,false"
    }
]
```
And then makes GET request to `/api/stratification` which retrieves the following:
```
[
    {
        "factor": "frl",
        "factorValue": {
            "false": 4,
            "true": 2
        },
        "experimentIds": [
            null,
            null,
            null,
            null
        ]
    }
]
```

This results in the following UI:
<img width="750" alt="Screenshot 2023-11-20 at 10 55 56 PM" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/903e6109-6123-44b6-a3b0-bde141e4f44d">

bcb37:
Reopening

zackcl:
QA: Passed

PR #1167, opened 2023-12-05
SRS factor delete and download bugfix

Yagnik56:
This PR fix issue of table not updating on deleting srs factor and added notification for successful download, delete and import operation of SRS factor.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/StratificationService.ts
+++ b/backend/packages/Upgrade/src/api/services/StratificationService.ts
@@ -92,7 +92,6 @@ export class StratificationService {
   }
 
   public async insertStratification(csvData: string, logger: UpgradeLogger): Promise<UserStratificationFactor[]> {
-    // const csvData = request.body[0].file;
     const rows = csvData.replace(/"/g, '').split('\n');
     const columnNames = rows[0].split(',');
 
--- a/frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
+++ b/frontend/projects/upgrade/src/app/core/stratification-factors/store/stratification-factors.effects.ts
@@ -8,6 +8,8 @@ import * as StratificationFactorsActions from './stratification-factors.actions'
 import { StratificationFactorsDataService } from '../stratification-factors.data.service';
 import { selectAllStratificationFactors } from './stratification-factors.selectors';
 import { StratificationFactor, StratificationFactorDeleteResponse } from './stratification-factors.model';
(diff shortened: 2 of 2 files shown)


danoswaltCL:
Code Review: @danoswaltCL

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -22,8 +22,12 @@ export class StratificationService {
   ): FactorStrata[] {
     const formattedResults = results.reduce((formatted, result) => {
       const { factor, value, count, experimentIds } = result;
+      const expIds: string[] = [];
+      experimentIds.forEach((expId) => {
+        expId ? expIds.push(expId) : null;
```
danoswaltCL, line 27: This is a long way of doing a `map`, but why exactly are we setting falsey values to `null`? Do we want `null` for some special reason somewhere else?

frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -101,15 +101,15 @@ export class StratificationComponent implements OnInit {
   }
 
   getExperimentIdsTooltip(experimentIds: any[]): string {
-    return 'Experiment IDs: [' + experimentIds.join(', ') + ']';
+    return 'Experiment IDs: [' + experimentIds?.join(', ') + ']';
```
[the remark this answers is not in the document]
Yagnik56: We were getting array of null value from query when SRS factor isn't assigned to any experiment but there are many experiment existing, you can look at the screenshot zack shared in this QA https://github.com/CarnegieLearningWeb/UpGrade/issues/967 where experimentIds getting multiple null values in array so I am filtering all the null values from array so if there is no experiment assign to particular factor expIds will be empty array `[ ]`. 
image

backend/packages/Upgrade/src/api/services/StratificationService.ts
danoswaltCL: It's a bit confusing to use a `forEach` and a ternary operator that is returning null to nowhere, all we're trying to do is a normal filter. I think `const expIds = experimentIds.filter(experimentId => !!experimentId)` works fine, right?

Yagnik56:
@danoswaltCL resolved your previous cmt, you can review this PR again and let me know if any changes required otherwise you can merge this PR.

PR #1177, opened 2023-12-08
⚡ Single command Dockerization without nix OS dependency to Quick Start Upgrade Locally 

ppratikcr7:
This PR contains an alternate docker setup without using NixOS dependency which will make it even more easier for others to try out `Upgrade` locally just depending upon Docker. Also, this PR will now have only 2 containers. One for DB and one for Frontend+Backend. There is a new dockerfile at the root level that will install all the packages and run frontend and backend concurrently.


dev.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY ./frontend/package*.json ./frontend/
+RUN cd frontend && npm ci
+
+# Install backend dependencies
+COPY ./backend/packages/Upgrade/package*.json ./backend/packages/Upgrade/
+RUN cd backend/packages/Upgrade && npm ci
+
+# Copy frontend and backend source
+COPY ./frontend ./frontend
+COPY ./backend/packages/Upgrade ./backend/packages/Upgrade
```
VivekFitkariwala, line 30: Why copy the code after installation?

docker-compose.yml
```diff
@@ -1,34 +1,21 @@
...
-      context: .
-      dockerfile: ./backend/dev.Dockerfile
-    image: upgrade-backend:latest
+      dockerfile: ./dev.Dockerfile
+    image: upgrade-app:latest
     user: "node"
     env_file:
       - ./backend/packages/Upgrade/.env.docker.local
     environment:
       - NODE_ENV=development
```
VivekFitkariwala, line 11: Not sure if we should keep this as development or production?

VivekFitkariwala (changes requested):

dev.Dockerfile
ppratikcr7: Done

docker-compose.yml
ppratikcr7: I feel we are using this docker for development, as it was in previous dockerfile as well. @danoswaltCL @kawcl Should we have a production version for the docker image that gets pushed automatically to dockerhub using github actions?

Makefile
danoswaltCL, line 1: why is this deleted? I find it very helpful.

docker-compose.yml
danoswaltCL: Maybe? Originally the setup was only meant to be a dockerized local setup. I like the idea of having a production-ready docker image that we pop into dockerhub, and maybe they can be the same basic image but I'm probably not the right person to review it, I'd like to run these ideas by our SRE guy @kcalvo-cli and @kawcl when we get back after the holidays before we merge this in. We definitely talked about that kind of thing when we were helping another team implement UpGrade on their prod env.

dev.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Install concurrently globally
+RUN npm install -g concurrently
+
+# Install frontend dependencies
+RUN cd frontend && npm ci
+
+# Install backend dependencies
+RUN cd backend/packages/Upgrade && npm ci
+
```
danoswaltCL, line 23: I think the `types` package would need to have `npm ci` also?

Makefile
ppratikcr7: Okay, actually removed all things related to nix based setup for docker implementation. Do you want me to keep the Makefile?

dev.Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Use an official Node runtime as a parent image
+FROM node:18-alpine
+
+# Create app directory
+WORKDIR /usr/src/app
+
+COPY ./types ./types
+
+WORKDIR /usr/src/app/types
+RUN ["npm", "ci"]
```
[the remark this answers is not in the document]
ppratikcr7: @danoswaltCL Here is why I removed the `npm ci`.

Makefile
danoswaltCL: shouldn't be anything nix-related in there, i'd like to keep it.

ppratikcr7: Then we would need multiple docker-compose files. One for the docker only implementation and one for docker+nix implementation.

danoswaltCL:
removing my request to keep the Make file, as we discussed it is not essential.

I would still like to bring this change up with @kcalvo-cli for our Thursday meeting to talk about production vs dev builds in this manner and publishing dockerhub images if that is going to be how we are going to go forward recommending launching production-ready UpGrade out-of-the-box.

@amurphy-cl can this be part of the agenda for Thurs SRE mtg?

ppratikcr7:
@bcb37 Can you please check now? Without reverting the `tsconfig` file for the moduleOptions changes. The issue seems to be resolved after the cache-manager null reference PR is merged.

danoswaltCL:
@ppratikcr7 @VivekFitkariwala there's a mtg scheduled on monday w/ CL devops folks that I'll be attending, will get some clarity on where to take this proposed idea

danoswaltCL:
@VivekFitkariwala @ppratikcr7 So after meeting with CL security folks, they've said that they would rather not host images on dockerhub. I do think there is still a benefit to providing an all-in-one production-capable dockerfile that could be used to for both quick-start and quick-hosting, that would be great.

Here's my thoughts:
1. We want to keep the current backend `Dockerfile` because we still need to support containerized single-service backends (such as CL!)
2. We should keep the `dev.Dockerfile`s in frontend and backend for development purposes only (I personally like all the services in their own containers for local development, so I would be sad to see this removed).
3. We can still remove the Make file and Nix stuff 👍 . I'll adapt.
4. This combined image can then be offered as a separate `single_container.Dockerfile` or whatever, instead of overwriting the existing ones.

@bcb37 @amurphy-cl let me know if that sounds like what was talked about.

ppratikcr7:
> @VivekFitkariwala @ppratikcr7 So after meeting with CL security folks, they've said that they would rather not host images on dockerhub. I do think there is still a benefit to providing an all-in-one production-capable dockerfile that could be used to for both quick-start and quick-hosting, that would be great.
> 
> Here's my thoughts:
> 
> 1. We want to keep the current backend `Dockerfile` because we still need to support containerized single-service backends (such as CL!)
> 2. We should keep the `dev.Dockerfile`s in frontend and backend for development purposes only (I personally like all the services in their own containers for local development, so I would be sad to see this removed).
> 3. We can still remove the Make file and Nix stuff 👍 . I'll adapt.
> 4. This combined image can then be offered as a separate `single_container.Dockerfile` or whatever, instead of overwriting the existing ones.
> 
> @kcalvo-cli @bcb37 @amurphy-cl let me know if that sounds like what was talked about.

@kcalvo-cli @bcb37 @amurphy-cl Can you please confirm, so that I can make necessary updates to the PR.

dev.Dockerfile
danoswaltCL, line 1: this looks the same as the `dev.singleContainerApp.Dockerfile`, I don't think this file is needed.

ppratikcr7: You are right, forgot to delete the file. Done.

VivekFitkariwala (approved):

danoswaltCL (approved):

PR #1190, opened 2023-12-15
Enhancement/export segment scv

Yagnik56:
This PR contain all changes to export segment members CSV as feature.


frontend/projects/upgrade/src/assets/i18n/en.json
```diff
@@ -382,6 +382,9 @@
   "segments.duplicate-segment.segmentName-input-placeholder.text": "Name of the duplicate segment",
   "segments.duplicate-segment.segmentDescription-input-placeholder.text": "Description of the duplicate segment",
   "segments.duplicate-segment-duplicate.text": "duplicate",
+  "segments.overview.export-segment.text": "Export Segment",
+  "segments.overview.export-warning.text": "The segment data will be sent to '{{ email }}'.",
```
bcb37, line 386: Does this 'warning' text ever get used? It looks like the segment CSV downloads are not emailed.

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -36,6 +37,11 @@ interface IsSegmentValidWithError {
   isSegmentValid: boolean;
 }
 
+interface SegmentParticipantsRow {
+  UUID: string;
```
bcb37, line 41: This results in the header for the segment (or user) name being 'UUID'. At least in the case of segments, it's not likely to be a UUID. Maybe 'Name' would be a better header text?

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -400,6 +406,43 @@ export class SegmentService {
...
+        throw new Error(SERVER_ERROR.QUERY_FAILED);
+      } else {
+        segmentsDoc.push(segmentDoc);
+      }
+    }
+
+    const segmentExportObj: SegmentFile[] = [];
+    segmentsDoc.forEach((segmentDoc) => {
+      const segmentRows: SegmentParticipantsRow[] = [];
+      // segmentRows.push({ UUID: segmentDoc.context, Type: 'context' });
```
bcb37, line 430: Do we need this commented-out line?

bcb37 (changes requested):
A couple issues.
Also, do we not want to include segment-level metadata (context, description, etc) in the export, like we do with the JSON one?

VivekFitkariwala (changes requested):

Yagnik56:
> segments.overview.export-segment.text

no because we only want members in csv, right?

backend/packages/Upgrade/src/api/services/SegmentService.ts
Yagnik56: true but csv shared by april and zack contains UUID so I kept it that way.

types/src/Experiment/enums.ts
```diff
@@ -181,6 +181,11 @@ export enum EXPORT_METHOD {
   DATA = 'Email Experiment Data (CSV)',
 }
 
+export enum EXPORT_Segment_METHOD {
```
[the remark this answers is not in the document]
Yagnik56: we have experiment export enum in this file and it is also used in frontend only, so can you confirm still wanna move to frontend?

backend/packages/Upgrade/src/api/controllers/SegmentController.ts
```diff
@@ -388,4 +388,24 @@ export class SegmentController {
     }
     return this.segmentService.exportSegments(ids, request.logger);
   }
+
+  @Post('/export/csv')
```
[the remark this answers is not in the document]
Yagnik56: resolved

frontend/projects/upgrade/src/assets/i18n/en.json
Yagnik56: we were but it was unnecessary code, thx for pointing it out

bcb37 (approved):

zackcl:
@amurphy-cl Currently, the format of the expected CSV file looks like the following:

```
UUID,Type
student1,individual
segment1,segment
school1,schoolId
```

Do you think this format is OK, or would you prefer the two columns to be swapped to match the table in the UI?
Screenshot 2024-01-18 at 8 15 21 AM

Currently, the type 'individual' and 'segment' are exported in lowercase, but they appear capitalized in the UI. I believe they should also be capitalized when exported. Additionally, we should check the case sensitivity during import. Let me know your thoughts on this.

Issue #1189, opened 2023-12-14
Remove Dark mode theming in components


danoswaltCL:
If we're officially ending support for dark-mode, we should formally disable dark mode toggle in the UI and remove dark-mode theming from the css and code.

Acceptance criteria: 
1) Dark mode is not accessible from profile in UI
2) Dark mode related code has been verified to have been deleted from the Angular project

danoswaltCL:
QA: @zackcl

zackcl:
QA: Passed

PR #1195, opened 2023-12-21
Feature/angular 16

danoswaltCL:
This will split up the angular migration to 17 into different parts. Updating to 16 was trivial, and this includes most of the dependency updates that will be needed for frontend, I couldn't really update things separately.

Migrating to 17 will be relatively simple also, but there is one breaking change that has to be dealt with, which is to migrate everything off of "legacy" Material components and start using "mdc" components: https://rc.material.angular.io/guide/mdc-migration. We can't go directly to 17 until this is handled.

Luckily, there is a handy script that will help us do this: `ng generate @angular/material:mdc-migration`

I ran the script and it will help a lot, but we will still have to do a number of adjustments manually, as many custom css hacks we've done in the past will definitely no longer work on the new components. Let this be a lesson to limit how often we do that (i.e. things like `::ng-deep .mat-form-field-appearance-legacy.mat-form-field-invalid:not(.mat-focused) .mat-form-field-ripple {
    display: none;
  }`

1. First we update to Angular 16 (this PR)
2. Go ahead and do task to remove dark-theme styling/components to simplify the next step (#1189)
2. Then we update to mdc components, which may be broken up into a few PRs depending on how much funkiness needs fixed.
3. Then we update to Angular 17


frontend/projects/upgrade/src/app/core/auth/auth.guard..ts
bcb37, line 1: Is there a reason for the extra dot in this file's name?

bcb37 (approved):
This looks ok to me: just the one question, and conflict with dev needs to be resolved.

frontend/projects/upgrade/src/app/core/auth/auth.guard..ts
danoswaltCL: bonus dot. i can go ahead and rectify that in this pr.

danoswaltCL: done

danoswaltCL: that dismissed the review if you can re-approve the filename change

bcb37 (approved):
LG

PR #1201, opened 2023-12-29
Fix null reference in CacheService

RidhamShah:


ppratikcr7 (changes requested):
This doesn't fix the issue when caching_enabled is false for a fresh DB. @RidhamShah Can you please check?

danoswaltCL:
just looking at this from quick overview, it seems like we're checking for the existence of cacheManager deep in the internals of these functions that don't need to run at all if there is no cache being used. If we can, I think it would be a simpler pattern to just short-circuit and return early if cache service is not detected.

ppratikcr7 (approved):
Local QA done. `CACHING_ENABLED = false` for clean DB works fine.

ppratikcr7:
> just looking at this from quick overview, it seems like we're checking for the existence of cacheManager deep in the internals of these functions that don't need to run at all if there is no cache being used. If we can, I think it would be a simpler pattern to just short-circuit and return early if cache service is not detected.

Hmm, seems correct. We can track that in a separate ticket. This seems to quickly resolve the issue as of now, so merged it.

PR #1204, opened 2024-01-03
Feature/1189 remove dark theme

danoswaltCL:
This removes dark-theme styling / logic around theming and automatically applies light-theme, i.e. this leaves in place the general theming mechanism, so there is still a "light-theme" even though it's the only theme.

I'm gonna see if it's relatively easy to get rid of the "theme.scss" files altogether as a follow-up, but that be more work than is worth it. This PR should be good enough to clear out dark-theme without major surgery.


bcb37 (approved):
This looks ok to me. One thing: do we want to leave the 
`"profile.action.dark-theme.text": "Dark mode",`
line in the translation text file?

danoswaltCL:
> This looks ok to me. One thing: do we want to leave the `"profile.action.dark-theme.text": "Dark mode",` line in the translation text file?

Ah missed that one

bcb37 (approved):

Issue #1196, opened 2023-12-21
Migrate to "MDC" Material Components


danoswaltCL:
This is a requirement before we can move to Angular 17:

https://rc.material.angular.io/guide/mdc-migration

This will require running a script:
`ng generate @angular/material:mdc-migration`

And then taking care of the "TODOs" that it creates.
This will likely involve some re-styling and potentially some code changes to accommodate stuff that script can't fix for us (hint: there's quite a bit).

Dependencies:
This should wait until Angular 16 is merged into dev ##1189 is completed

AC:
The application will look and feel as close to unchanged as we can.

PR #1205, opened 2024-01-05
mdc migration: Card (use raised instead of outlined)

danoswaltCL:
This just updates the "mat-card" components (and the initial changes in `styles.scss` to prepare migration); this one is fairly straightforward, the migration script introduces `appearance: "outlined" to the html templates, which looked wrong, but switching it to "raised" looked like the way we have styled it previously.

Please make sure everything looks normal!

https://material.angular.io/components/card/api

#1196


zackcl:
Everything looks good as I checked.

zackcl (approved):

PR #1208, opened 2024-01-05
mat-button migration

danoswaltCL:
https://rc.material.angular.io/guide/mdc-migration#button


frontend/projects/upgrade/src/styles.scss
```diff
@@ -236,9 +236,10 @@
...
 }
 
-.mat-button,
-.mat-flat-button,
-.mat-stroked-button {
+.mat-mdc-button,
+.mat-mdc-unelevated-button,
+.mat-mdc-outlined-button {
+  /* TODO(mdc-migration): The following rule targets internal classes of button that may no longer apply for the MDC version.*/
   &.cdk-program-focused .mat-button-focus-overlay {
```
danoswaltCL, line 243: I'm not sure what this was even doing? May be able to remove?

zackcl:
I've made a commit to override the button `letter-spacing` to `normal` based on the [MDC migration guide for button](https://rc.material.angular.io/guide/mdc-migration#button):

> Letter-spacing is 1.25px instead of normal.

To replicate the exact look of our current design, I would also need to include -webkit-font-smoothing: auto; and resize the mat-icon to 24px. However, I found the appearance to be better without applying these changes. Please let me know if you would prefer to adhere more closely to the previous design.

New Design:
Screenshot 2024-01-09 at 4 33 33 PM

Previous Design (`dev` branch):
Screenshot 2024-01-09 at 4 34 44 PM

frontend/projects/upgrade/src/styles.scss
zackcl: I removed them as these classes don't seem to be used.

zackcl (approved):

PR #1210, opened 2024-01-05
mdc checkbox migration

danoswaltCL:
https://material.angular.io/components/checkbox/overview

https://rc.material.angular.io/guide/mdc-migration#checkbox <--- useful!

checkboxes can be found in: 

view-experiment

the enrollment-over-time selector box
image

The weighted checkbox in both the simple and factorial design stepper components.

The start-end date checkboxes


danoswaltCL:
I believe this should be ready now, I had to reapply old styles in a different manner to get labels to look correct. Also the positioning was a little different due to slightly bigger size, so that should be looking right now.

Old classes to force coloring didn't look like they were needed anymore. Removing them looked and functioned same as in staging, so they were removed altogether.

frontend/projects/upgrade/src/styles.scss
```diff
@@ -242,6 +242,19 @@ button[mat-button], button[mat-raised-button], button[mat-flat-button], button[m
   width: 805px;
 }
 
+.mat-mdc-checkbox {
+  .mdc-checkbox {
+    transform: scale(0.889); // transform 18px to 16px
+  }
+
+  span {
```
danoswaltCL, line 250: can we give these spans a class instead of targeting an html element? I've always followed the school of thought that classes are ideal unless you can't avoid it. Unless angular creates class-less spans dynamically at runtime? but I think these spans are the ones you've added in the template.

zackcl: I like that idea. I will add a class called `checkbox-label` to the existing span elements within the `mat-checkbox`.

zackcl:
I think everything looks good now. Here's a brief summary of the changes I made (for future reference):

- Since the text styles are not inherited, setting the font class (e.g., `ft-16-400`) directly to the `mat-checkbox` element no longer worked. I've therefore wrapped the inner text with a `span` element like in the following example
```
    <mat-checkbox color="primary" (change)="setArchivedStatusAgreement()">
      <span class="ft-14-700 checkbox-label">
        {{ 'home.change-experiment-status.archiving-confirmation.text' | translate }}
      </span>
    </mat-checkbox>
```
- These `span` elements with the class `checkbox-label` are now styled using the following overridden CSS in the `styles.scss` file:
```
.mat-mdc-checkbox {
  .mdc-checkbox {
    transform: scale(0.889); // transform 18px to 16px
  }

  .checkbox-label {
    margin-left: -8px;
    color: var(--black-2);
    letter-spacing: normal;
    user-select: none;
  }
}
```
- The remaining changes mostly involve adjusting the position of the checkbox by updating its margins, as shown below:
```
.mat-mdc-checkbox {
    margin-top: -3px;
    margin-left: -12px;
  }
```

Here are the screenshots showing the checkboxes after the implementation of the changes:










Additionally, please note that now, after toggling a checkbox, the ripple effect will remain visible instead of fading out. This behavior is part of the changes introduced in the new MDC-based checkbox.

zackcl (approved):

PR #1212, opened 2024-01-05
mat-chips migration

danoswaltCL:
https://rc.material.angular.io/guide/mdc-migration#chips

These look fine to me, one funny difference is that for some reason now if you create a segment and then click on the "context" mat-chip, it gives a nice little checkmark animation... but it means nothing. The other chips don't do it though 🤷 , and they didn't do it before...

image


danoswaltCL:
the solution above was to use the `mat-chip` element instead of what the migration used, `mat-chip-option`. They don't need to be "selectable".

these `mat-chip` elements can be found as tags in segments list, experiment list, view experiment details, and the first step of the experiment creation stepper.

I think this one should be safe to merge in without further changes, the look and feel seem same as before to me, but double-check me obviously! @zackcl

zackcl:
I think everything looks good now. I've basically restyled the `mat-chip` element's label so that they all look like the previous design:
```
.mat-mdc-chip .mat-mdc-chip-action span.mat-mdc-chip-action-label {
  color: var(--black-2);
  letter-spacing: normal;
}
```
I've also made all `mat-chip` elements look consistent across the app. Here are the screenshots showing the chips after the implementation of the changes:

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.scss
```diff
@@ -49,7 +49,7 @@
     margin-top: -3px;
     margin-left: -12px;
   }
-
+  /* TODO(mdc-migration): The following rule targets internal classes of chips that may no longer apply for the MDC version. */
```
danoswaltCL, line 52: @zackcl do you think this rule is of any importance now?

frontend/projects/upgrade/src/app/features/dashboard/home/pages/view-experiment/view-experiment.component.scss
```diff
@@ -243,9 +243,11 @@ $font-size-small: 15px;
 
           span {
             margin: 5px;
-            border-radius: 14px;
-            background-color: rgba(216, 216, 216, 0.6);
-            padding: 7px 10px;
+            border-radius: 16px;
```
danoswaltCL, line 246: i know it's an existing rule but if these spans are all in the template this could be nicer as a named class also, while we're at it.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.html
```diff
@@ -216,33 +216,33 @@
           </span>
         </th>
         <td class="ft-12-600" mat-cell *matCellDef="let experiment">
-          <mat-chip-list>
+          <mat-chip-listbox>
             <ng-container *ngFor="let context of experiment.context; let contextIndex = index">
-              <mat-chip
+              <mat-chip-option
```
danoswaltCL, line 221: it looks like I missed this in my second commit, these should be `mat-chip` and not `mat-chip-option`, would you mind changing that while you're in there?

frontend/projects/upgrade/src/app/features/dashboard/home/pages/view-experiment/view-experiment.component.scss
zackcl: Thanks! I've wrapped the inner text of `mat-chip` with `span` and added a `chip-label` class to these span elements (just like how I did for styling checkbox labels)

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.html
zackcl: Done!

frontend/projects/upgrade/src/app/features/dashboard/home/pages/view-experiment/view-experiment.component.scss
zackcl: I've also updated the fake chips in the experiment view page to use `mat-chip` so I could get rid of those styles.

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.scss
zackcl: Everything looked fine without those rule, so I removed it.

zackcl:
Just like how we did for `mat-checkbox`, I've wrapped the inner text with a span and added the class `chip-label`:
```
<mat-chip-listbox>
  <mat-chip *ngIf="segment.context != 'ALL'">
      <span class="ft-12-600 chip-label">
          {{ segment.context }}
      </span>
  </mat-chip>
</mat-chip-listbox>
```
And these labels can be styled with the following in the `styles.scss` file:

```
.mat-mdc-chip {
  .chip-label {
    color: var(--black-2);
    letter-spacing: normal;
  }
}
```

zackcl (approved):

PR #1213, opened 2024-01-07
:bug: Free groups

VivekFitkariwala:
On group change free enrollment groups.

It fixes some queries related to removing enrollments.


bcb37:
> On group change free enrollment groups.
> 
> It fixes some queries related to removing enrollments.

I don't see how this removes enrollments on changes to groups. It looks like it removes _exclusions_ on changes to _working_ groups. In particular, entries seem to be being deleted from the `individual_exclusion` and `group_exclusion` tables in the db. How would the application break if those entries were not removed?

VivekFitkariwala:
@bcb37 we don't have any test cases related to these edge cases and you are correct we should have test cases. It is a good idea to add more comments and update the naming. I will add/update this.
Thanks.

RidhamShah:
@bcb37 added test cases for the working group change scenarios. You can re-review this PR.

bcb37:
> @bcb37 we don't have any test cases related to these edge cases and you are correct we should have test cases. It is a good idea to add more comments and update the naming. I will add/update this. Thanks.

The test cases were very helpful to me in understanding this issue. However, shouldn't the group exclusions/enrollments be added back if the group membership is reverted back to the original value?

bcb37 (commented):
Just one more question

bcb37 (approved):

PR #1228, opened 2024-01-10
mdc migrate tooltips

danoswaltCL:
simple one, just class name change, functionality looks the same


zackcl:
I think this looks good without any changes.

zackcl (approved):

PR #1261, opened 2024-01-23
migrate to mdc tabs wip

danoswaltCL:
@zackcl i'm having a hard time figuring out how to style these tabs in a similar way, I've tried on a few different occasions to re-style it and it just is not so easy, or I'm going the wrong way about it. I've committed basically what the migration does, I've opened this as a draft work-in-progress if you wanted to take a stab at it.


zackcl:
@danoswaltCL This one is ready for review.

zackcl (approved):

PR #1265, opened 2024-01-25
Feature/1221 mdc migration form elements

danoswaltCL:
This is a work-in-progress. The migration tool for this one updates multiple form components. Mostly this has impacts for mat-form-field, mat-inputs, and mat-selects, which do no look the same as the previous "legacy" look.

Along with the migration script changes, I've tried to get close to the previous look and feel without much css hacks. Thankfully, functionality of the components is no different, it's mostly just a lot of adjustments to look and feel.

@zackcl there are 2 main things I think to consider about what we need to do get this looking right:

1. Spacing:

The way I found that I could do this without lots of re-styling (yet) is to start with Material theming's very poorly documented but absolutely life-saving "density" property, which allows a kind of magical resizing capability to condense and expand many types of components (think Google Apps settings that you can select as a user to make the layout "default" or "comfortable" or "compact")

https://material.angular.io/guide/theming#customizing-density
https://m2.material.io/design/layout/applying-density.html#usage

The way I did this was to allow us to use a set of classes called `density-1` and `density-2` to that contain the necessary sass mixin, pretty much ripped right out of this:

https://stackblitz.com/edit/angular-kyfetk?file=src%2Fapp%2Fform-field-appearance-example.html

I found that applying this class just to mat-form-fields helped out hugely in the design-stepper, along with a property called `"subscriptSizing"="dynamic"` which tells the mat-form-field not to reserve space underneath, to limit the amount of new css fiddling.

2. Appearance:

There is a property on mat-form-fields called "appearance", which can either be "filled" or "outlined". Both look different than the previous design. Default is filled, and I have removed the dark background for normal state, but check out the "appearance"="outlined" look also, maybe that's the easier one to adjust. I don't know, both sort of look fine to me, just different.

There are other small particular issues with this and that, but overall what do you think of the direction it is going?


zackcl:
This is still work in progress. I've updated the `mat-mdc-form-field` style for the Segments section as well. I still have a few other pages to work on including the Participants, Logs, and Profile page.

danoswaltCL:
@zackcl looks ready to me for approve and merge!

zackcl (approved):

PR #1283, opened 2024-02-01
Bugfix/1270 docker build failure

VivekFitkariwala:
In the upgraded `ts-node` and `tsconfig-paths`, `tsconfig-paths` is not using the typescript configuration file loaded by `ts-node` through `--project` flag. But passing it through the environment variable does the trick.


bcb37 (approved):
Tested and verified that it works locally.

PR #1285, opened 2024-02-01
remaining mdc and final removal of legacy stuff

danoswaltCL:
Last one! Please double-check a few components, the migration script doesn't do a lot with them and they looked fine to me:

`mat-slider-toggle`
`mat-progress-bar`
and the snackbar notifications

Ignore `mat-paginator`, it is only used in a dead component that should itself be removed in a different tech debt story.

The other components like `mat-menu` removed from the "shared.module" have no actual implementations in the html but probably lead to larger build-size.


zackcl:
I think everything still looks fine after the removal!

zackcl (approved):

PR #1287, opened 2024-02-03
angular 17

danoswaltCL:
This is the basic migration to angular 17, nothing very exciting now that the angular material components have gotten swapped over to the new version, ty @zackcl for all of your effort. There may be some other things we want to investigate like @VivekFitkariwala's suggestion of looking at https://angular.io/guide/esbuild but this will let us close the ticket on ng 17 migration.


bcb37 (approved):
This looks good to me after running it a bit. There's the one warning about 'using the disabled attribute with a reactive form directive' which apparently can result in 'changed after checked' errors, and appears to be new. But we'd probably want to address that on another ticket, if at all.

PR #1288, opened 2024-02-05
resolved issue of segment details page shared link not working

Yagnik56:
Resolved issue of sharable links to segment detail pages not working


frontend/projects/upgrade/src/app/core/segments/segments.service.ts
```diff
@@ -23,6 +24,17 @@ export class SegmentsService {
   allExperimentSegmentsInclusion$ = this.store$.pipe(select(selectExperimentSegmentsInclusion));
   allExperimentSegmentsExclusion$ = this.store$.pipe(select(selectExperimentSegmentsExclusion));
 
+  selectSegmentById(segmentId: string) {
```
danoswaltCL, line 27: 2 pipes feel a little funny, I wouldn't be surprised if that's done in other places in our code base but really the regular flat pattern to do this with one pipe should work like this (i think), can you try it?

```
selectSegmentById(segmentId: string) {
  return this.store$.pipe(
    select(selectSegmentById, { segmentId }), // select the segment
    tap((segment) => {
      if (!segment) {
        this.fetchSegmentById(segmentId);
      }
    }),
    map(segment => ({ ...segment }))
  );
}
```

danoswaltCL:
@Yagnik56 just had one the comment for something to try, this looks good to me otherwise

PR #1291, opened 2024-02-05
use array to avoid param limit in log query

bcb37:


backend/packages/Upgrade/src/api/repositories/LogRepository.ts
```diff
@@ -164,7 +164,11 @@ export class LogRepository extends Repository<Log> {
       .innerJoin('queries.metric', 'metric')
       .innerJoin('metric.logs', 'logs')
       .where('experiment.id=:experimentId', { experimentId })
-      .andWhere('logs."userId" IN (:...userIds)', { userIds })
```
VivekFitkariwala, line 167: The where clause here will restrict the list of `userIds` to 50.
Both the codes are doing identical things because after unnesting you are returning 
`'logs."userId" IN ' + subQuery`

bcb37: They were not restricted to 50 when I tested it locally, or when the bug was reported in production. It exceeded the limit of ~64k parameters in postgres. UNNEST fixes it by converting the list of parameters to a single parameter, which is converted by postgres into a native list.

VivekFitkariwala:
@bcb37 Should we merge this PR? 
We also have [this](https://github.com/CarnegieLearningWeb/UpGrade/pull/1296#pullrequestreview-1870304954) PR which will fix this issue.

bcb37:
Bug was addressed in https://github.com/CarnegieLearningWeb/UpGrade/pull/1296

Issue #1286, opened 2024-02-01
csv export fails for > 64k users


bcb37:
Attempting to export an emailed csv of an experiment with > 64k participants results in a postgres error.
We should be able to support experiments with at least that many users.

Dev note: The query that gets log data has a WHERE clause that parameterizes the entire list of users. We may be able to avoid avoid postgres's 64k limit on parameters by using an array.

danoswaltCL:
ah nice find

PR #1296, opened 2024-02-07
Replace skip/take with offset/limit which is relevant for sql based query.

ppratikcr7:
This PR replaces the skip/take with offset/limit in mostly all the typeorm queries. So, basically skip/take is typeorm based and was not reflected back in the converted SQL query and we were not getting 50 user set for log query, which gave the 64k user limit issue #1286. This PR should resolve that as well. I tested the experiment list skip/take query after the update, but seems to not load the next set of experiments in paginated query at few times. Not sure why, might need to dig deeper. So, I have kept that skip/take for now in this query.

@bcb37 Meanwhile you can test the 64k issue by replicating it in your K6 test suite.


bcb37 (changes requested):
This looks ok and seems to fix the bug. There are just some test failures, so we probably need to update the backend tests.

ppratikcr7:
@bcb37 Thanks a lot for updating those test cases.

ppratikcr7:
@danoswaltCL We can merge this now.

PR #1302, opened 2024-02-12
🐛 Bugfix for exclusion's not working as expected

### LINKED ISSUE #1301: Exclusions are not handled as expected in mark call and code needs optimization
e.g "5.1.0"

**Describe the bug**
Currently the various exclusions codes are not stored in the DB for simple and within subject experiments. Mark call code for exclusions needs optimization to not call various code blocks for all experiment state. We are also storing multiple rows for Individual Enrollments for within subject experiments and we are also storing multiple rows for monitored decision point table for experiment level exclusions.

**To Reproduce**
Steps to reproduce the behavior for few exclusions, or refer to the detailed [document](https://docs.google.com/document/d/1e-FPHZ5ofo1u_3MSuhf7VVJdMQmQ2U9vQGf_luyoCHo/edit?usp=sharing) for all exclusion codes:
1. Try to exclude a user who is on exclusion list using a segment or individual exclusion list.
2. Once you hit /mark api for the user, you wont see its Data as the table is not getting populated with the reason for exclusion.

**Expected behavior**
Exclusions should work as expected for the above scenario.

ppratikcr7:
This PR fixes #1301 

Moving the exclusion logic out of the updateExperimentExclusion() function, as the code was never reached for excluded experiment.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -143,7 +143,7 @@ export class ExperimentAssignmentService {
 
     const previewUser: PreviewUser = await this.previewUserService.findOne(userId, logger);
 
-    // search decision points in experiments cahce
+    // search decision points in experiments cache
     const cacheKey = CACHE_PREFIX.MARK_KEY_PREFIX + '-' + site + '-' + target;
     const dpExperiments = await this.cacheService.wrap(cacheKey, () =>
       this.decisionPointRepository.find({
@@ -336,11 +336,6 @@ export class ExperimentAssignmentService {
               groupEnrollment: groupEnrollments,
               groupExclusion: groupExclusions,
             },
-            {
-              user: userExcluded,
-              group: groupExcluded,
-            },
-            exclusionReason,
             status,
             condition
           );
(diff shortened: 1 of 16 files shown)


ppratikcr7:
@danoswaltCL I guess we would need to cherry pick this PR commit into release/v5.1
As I used dev branch initially it has the version bump v5.2 commit included in this PR.

Or else should I create another PR, by taking a new branch off the release/v5.1 ?

danoswaltCL:
> @danoswaltCL I guess we would need to cherry pick this PR commit into release/v5.1 As I used dev branch initially it has the version bump v5.2 commit included in this PR.
> 
> Or else should I create another PR, by taking a new branch off the release/v5.1 ?

Yes please create the branch off of 5.1 and reopen, I'd like to avoid cherry-picking/ moving commits around if we can, merging "down" is just a bit more natural flow for synchronizing branches.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -374,6 +369,56 @@ export class ExperimentAssignmentService {
         condition: monitoredDocument.condition || logDocument.condition,
       };
     } else {
+      // experiment level exclusion
```
bcb37, line 372: Before this change, the enrollment exclusions were only updated when the experiment state was 'enrolling' or 'enrollment complete'. But this could run when the experiment is in any state. Is that a concern?
 
Also, I noticed that almost the identical code runs at the end of each branch of the `if` that test whether there are experiments after experiment-level exclusions are applied (line 256). 
However, in the first there is
``     const logDocument = await this.monitoredDecisionPointLogRepository.save({
        monitoredDecisionPoint: monitoredDocument,
        condition: condition,
        uniquifier: uniquifier,
      });
``
while in the second theres
``      const logDocument = await this.monitoredDecisionPointLogRepository.save({
        monitoredDecisionPoint: monitoredDocument,
        condition: assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS ? condition : null,
        uniquifier: assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS ? uniquifier : null,
      });
``
Do we only want to test for WITHIN_SUBJECTS to set the 'condition' and 'uniquifier' to `null` in the case where there's experiment-level exclusion?

bcb37 (commented):
A couple questions.

PR #1304, opened 2024-02-12
use upsert() on user creation

bcb37:
- replaces `save()` with `upsert()` when creating/editing users from the 'init' endpoint
- fix the response object of 'init' so it always returns the current user data structure (id, group, workingGroup)


backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -88,10 +88,10 @@ export class ExperimentUserService {
     }
   }
 
-  public async create(users: Array<Partial<ExperimentUser>>, logger: UpgradeLogger): Promise<ExperimentUser[]> {
+  public async create(users: Array<Partial<ExperimentUser>>, logger: UpgradeLogger): Promise<InsertResult> {
     logger.info({ message: 'Create a new User. Metadata of the user =>', details: users });
     // insert or update in the database
-    return this.userRepository.save(users);
+    return this.userRepository.upsert(users, ['id']);
```
VivekFitkariwala, line 94: Both the syntax `save` and `upsert` are the same in this case. Right?

VivekFitkariwala (approved):

PR #1309, opened 2024-02-14
storing default condition payload values in db with conditions

Yagnik56:
This PR will address issue of default condition payload not updating in db after changing it by default condition payload values same as condition code getting saved in db.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
+++ b/frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
@@ -323,9 +323,6 @@ export class ExperimentDesignStepperService {
 
     payloadTableData.forEach((payloadRowData: SimpleExperimentPayloadTableRowData) => {
       // if no custom payload, return early, do not add to array to send to backend
-      if (payloadRowData.payload === payloadRowData.condition) {
-        return;
-      }
 
       const parentCondition = conditions.find((condition) => condition.conditionCode === payloadRowData.condition);
 
@@ -356,7 +353,7 @@ export class ExperimentDesignStepperService {
         id: uuidv4(), // TODO: maybe not the right place?
         levels: conditionLevelsData,
         condition: conditions,
-        payload: '',
+        payload: conditions,
         weight: '0.0',
         include: true,
       };


frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
```diff
@@ -323,9 +323,6 @@ export class ExperimentDesignStepperService {
 
     payloadTableData.forEach((payloadRowData: SimpleExperimentPayloadTableRowData) => {
       // if no custom payload, return early, do not add to array to send to backend
```
ppratikcr7, line 325: You need to remove this comment as well. As the code is removed.

ppratikcr7 (commented):
Just one small comment, rest looks good.

ppratikcr7 (dismissed):
@Yagnik56 For Simple experiment, all looks good. But, for factorial experiment, new rows are added for editing of condition payloads instead of overwriting the existing one's.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "UpGrade",
-  "version": "5.1.0",
+  "version": "5.1.1",
```
VivekFitkariwala, line 3: Should this change go in this PR?

package-lock.json
VivekFitkariwala, line 1: Should this change go in this PR?

VivekFitkariwala (changes requested):

package-lock.json
Yagnik56: yes confirmed with dan

VivekFitkariwala (approved):

danoswaltCL:
At UpGrade root dir, run my custom script `npm run version:patch`. It looks like maybe you've run the regular npm command `npm version patch` at the project level, but we'll want all the package.jsons pipped at the same time, so we should see all the package versions changed (and the POM.xml).

Yagnik56:
> npm run version:patch

gettin

> At UpGrade root dir, run my custom script `npm run version:patch`. It looks like maybe you've run the regular npm command `npm version patch` at the project level, but we'll want all the package.jsons pipped at the same time, so we should see all the package versions changed (and the POM.xml).

getting some kind of error with that code, could you try it in branch and see if any changes are there to push.

danoswaltCL:
Ok don't worry about it, I can bump the versions when I merge.

I pulled and don't get any errors but maybe local setup matters, can you share the errors?

danoswaltCL (changes requested):
[deleted comment]

PR #1311, opened 2024-02-16
🐛 Hotfix: Fix for experiment level exclusions and code optimization for mark call

### LINKED ISSUE #1301: Exclusions are not handled as expected in mark call and code needs optimization

ppratikcr7:
This PR resolves #1301 

Here, below changes are made:
1. I have moved the experiment level exclusions logic from `updateEnrollmentExclusion()` which falls under individual/group level exclusions logic. So, basically the experiment level exclusions were not stored in DB.
2. Also, moved a code snippet in the `mark` apis' individual/group level exclusions code logic to be executed only in case of enrolling or enrollmentComplete experiment state.
3. There was duplicate code for storing monitoredDecisionPoint and monitoredDecisionPointLog documents in both individual/group level exclusions and experiment level exclusions logic. The individual/group level exclusions' monitoredDecisionPointLog document didnt had the check for within-subjects. And the experiment level exclusions' monitoredDecisionPoint existing document was not fetched and always a new document was getting stored. So, basically I have used a single code snippet common for both after the if-else branch executes.
4. Added a few comments in the exclusion code enum.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -143,7 +143,7 @@ export class ExperimentAssignmentService {
 
     const previewUser: PreviewUser = await this.previewUserService.findOne(userId, logger);
 
-    // search decision points in experiments cahce
+    // search decision points in experiments cache
     const cacheKey = CACHE_PREFIX.MARK_KEY_PREFIX + '-' + site + '-' + target;
     const dpExperiments = await this.cacheService.wrap(cacheKey, () =>
       this.decisionPointRepository.find({
@@ -254,14 +254,7 @@ export class ExperimentAssignmentService {
     );
 
     if (filteredExperiments.length) {
-      let monitoredDocument: MonitoredDecisionPoint = await this.monitoredDecisionPointRepository.findOne({
-        where: {
-          site: site,
-          target: target,
-          user: userId,
-        },
-        relations: ['user'],
-      });
+      // individual/group level exclusions:
 
       if (experimentId && experiments.length > 0) {
         const selectedExperimentDP = dpExperiments.find((dp) => dp.experiment.id === experimentId);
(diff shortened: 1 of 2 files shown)


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -254,14 +254,7 @@ export class ExperimentAssignmentService {
...
-        where: {
-          site: site,
-          target: target,
-          user: userId,
-        },
-        relations: ['user'],
-      });
+      // individual/group level exclusions:
 
       if (experimentId && experiments.length > 0) {
```
VivekFitkariwala, line 259: We should not use `experimentId` and `experiments` here. We should use `filteredExperiments` to get the eexperiment.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -349,54 +338,96 @@ export class ExperimentAssignmentService {
...
-      const assignmentUnit = experiments
-        ? experiments.find((exp) => exp.id === experimentId)?.assignmentUnit || experiments[0]?.assignmentUnit
-        : null;
-      const monitoredDocument = await this.monitoredDecisionPointRepository.saveRawJson({
-        id: uuid(),
-        experimentId: experimentId,
-        condition: assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS ? null : condition,
-        user: userDoc,
+      // experiment level exclusion
+      const experiment = experiments[0];
```
VivekFitkariwala, line 343: This else case means there is no valid experiment so I am not sure how this code will work here. Ideally, `filteredExperiment` should be an empty array here.

VivekFitkariwala (changes requested):

PR #1324, opened 2024-02-27
🐛 Hotfix: Fix for Exclusions logic(Simple & Within Subject Exps) and code optimization for mark call

ppratikcr7:
This PR resolves issue #1301

Below are the changes made in this PR.

For Simple experiment:
1. So, basically the experiment level exclusions were not getting stored in DB as `filteredExperiment` returned from experimentLevelExclusionInclusion() was empty.
2. Group Level exclusions were not getting stored in DB, as code block to fetch `groupexperiments` was checking for empty/invalid groups and returning empty `experiments` list. This code block was not required as we are already checking for `experimentsWithInvalidGroupAndWorkingGroup()` inside `updateEnrollmentExclusion()`.
3. Added conditions to check if group object has any keys or not for group level exclusions.
4. Also, moved a try catch code snippet in the mark apis' exclusions code logic to be executed only in case of enrolling or enrollmentComplete experiment state, else its not required to be executed for all other states of the experiment.
5. There was duplicate code for storing monitoredDecisionPoint and monitoredDecisionPointLog documents in both individual/group level exclusions and experiment level exclusions logic. The individual/group level exclusions' monitoredDecisionPointLog document didn't had the check for within-subjects. And the experiment level exclusions' monitoredDecisionPoint existing document was not fetched and always a new document was getting stored. So, basically I have used a single code snippet common for both after the if-else branch executes.
6. Added a few comments in the exclusion code enum.
7. Updated wrong variable names using 'Code Spell Checker' VSCode extension (ID: streetsidesoftware.code-spell-checker). Much helpful for oversight typos in variable naming.

For Within subjects experiment:
1. `REACHED_PRIOR` exclusion was not getting set as consistencyRule is null for Within subjects experiments.
2. Individual Enrollment document was not getting overwritten, so added a condition to check if the individualEnrollment is not present.

Please find attached [document](https://docs.google.com/document/d/1e-FPHZ5ofo1u_3MSuhf7VVJdMQmQ2U9vQGf_luyoCHo/edit?usp=sharing) to verify the exclusions and their expected behaviour. Might take some time to verify all, in case of doubts, do let me know.

Note: 
1. Will be adding detailed test cases to check alll exclusion codes in a separate PR using a separate ticket. It's better if these changes go in release branch asap.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -71,16 +71,18 @@ import { UserStratificationFactor } from '../models/UserStratificationFactor';
 @Service()
 export class ExperimentAssignmentService {
   constructor(
-    @InjectRepository() private experimentRepository: ExperimentRepository,
+    @InjectRepository() 
+    private experimentRepository: ExperimentRepository,
     @InjectRepository()
     private decisionPointRepository: DecisionPointRepository,
     @InjectRepository()
     private individualExclusionRepository: IndividualExclusionRepository,
-    @InjectRepository() private groupExclusionRepository: GroupExclusionRepository,
+    @InjectRepository()
+    private groupExclusionRepository: GroupExclusionRepository,
     // @InjectRepository()
     // private groupAssignmentRepository: GroupAssignmentRepository,
-    @InjectRepository() private groupEnrollmentRepository: GroupEnrollmentRepository,
     @InjectRepository()
+    private groupEnrollmentRepository: GroupEnrollmentRepository,
     // @InjectRepository()
     // private individualAssignmentRepository: IndividualAssignmentRepository,
     @InjectRepository()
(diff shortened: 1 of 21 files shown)


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -248,25 +210,38 @@ export class ExperimentAssignmentService {
...
+      }
 
-      if (experimentId && experiments.length > 0) {
-        const selectedExperimentDP = dpExperiments.find((dp) => dp.experiment.id === experimentId);
+      if (
+        (experiment.state === EXPERIMENT_STATE.ENROLLING ||
+          experiment.state === EXPERIMENT_STATE.ENROLLMENT_COMPLETE) &&
+        !previewUser
+      ) {
+        const experiment = await this.experimentService.findOne(experimentId);
```
VivekFitkariwala, line 243: Why are you getting the experiment document from DB? Don't you already have that information?

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
ppratikcr7: Yeah, this code was already existing, and was fetching the complete experiment document, as the earlier experiment object didnt had partitions and conditionPayloads keys. Partitions key was needed for checking enrollmentComplete condition, so added that in the relations for the earlier available experiment document. So, now I have removed this excess fetching of redundant data.

VivekFitkariwala (dismissed):

danoswaltCL (approved):

Issue #1252, opened 2024-01-18
Buttons Disappear When the Page is Refreshed


zackcl:
**Version where bug was found:**
5.1.0

**Describe the bug**
When the Experiments or Segments list page is refreshed, some or all of the buttons at the top (e.g., "Add Experiment") disappear.

**To Reproduce**
1. Go to the Experiments or Segments list page.
2. Refresh the page.
3. Observe that some buttons are not visible on the page.

**Expected behavior**
The buttons should remain visible upon page refresh.

**Screenshots**

https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/6998da28-f5b7-4615-a24d-7d2c7d405c6d

danoswaltCL:
Welp. I know what this probably is, I'll take this. This very likely is related to the fix where I dealt with auth and how pages reroute when refresh happens. These buttons disappear if a user doesn't have certain permissions, something's not working right apparently.

zackcl:
QA: @zackcl

zackcl:
I'm not sure if these are directly related to this issue, but here are a few problems I've noticed:

- When the Segment details page is refreshed, the status changes to `Unused`.
- Signing out from the Experiment or Segment details page triggers a "TypeError: Cannot destructure property 'state' of 'undefined' as it is undefined." error.
- Signing back in directs the user to the Segment (or Experiment) details page, but it might be more intuitive to default to the Experiments list page.
- Occasionally, refreshing the Segment details page (in either the `Global` or `Used` state) causes a "TypeError: Cannot read properties of null (reading 'segments')" error (refer to the screenshot at the bottom)."

https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/0e4a1552-ad37-4997-a090-80b725d1dbe7

<img width="750" alt="error on segment details page refresh" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/1c06bc5d-9084-4781-a08f-1f738c0193ae">

danoswaltCL:
Ok, back to the drawing board, I can take this back into progress and take a look, can create separate issues if they are indeed separate. Does this feel like it should be prioritized as a 5.1 release bugfix @amurphy-cl ? They don't appear like big blockers, but it 

[...]

PR #1327, opened 2024-02-27
Feature/error on details refresh undefined permissions

danoswaltCL:
This addresses another small bug unveiled after fixing segments page refresh in #1252, which was caused by a race condition between fetching permissions and fetching segments where if segments fetch finished first, permissions is null.

The permissions object was not being utilized properly though, that thing should be strictly read-only. The conditions currently used to disable those buttons are being determined by segment type / status, not user permissions.


danoswaltCL:
@zackcl ok NOW it can be reviewed/retested

zackcl (approved):

PR #1333, opened 2024-03-01
Old feature flag Implementation (For partial reuse or reference)

ppratikcr7:
Small Demo For Old Implementation:
https://github.com/CarnegieLearningWeb/UpGrade/assets/33730817/7d1353b5-5faf-4d4d-af2e-c2a980fdff96

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/dashboard-root/dashboard-root.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/dashboard-root/dashboard-root.component.ts
@@ -18,11 +18,11 @@ export class DashboardRootComponent implements OnInit {
       text: 'global.experiment.title',
       iconType: 'assignment',
     },
-    // {
-    //   path: ['/featureFlags'],
-    //   text: 'feature-flags.title.text',
-    //   iconType: 'toggle_on'
-    // },
+    {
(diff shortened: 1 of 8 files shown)


frontend/projects/upgrade/src/app/features/dashboard/dashboard-root/dashboard-root.component.ts
```diff
@@ -18,11 +18,11 @@ export class DashboardRootComponent implements OnInit {
       text: 'global.experiment.title',
       iconType: 'assignment',
     },
-    // {
-    //   path: ['/featureFlags'],
-    //   text: 'feature-flags.title.text',
-    //   iconType: 'toggle_on'
-    // },
+    {
```
danoswaltCL, line 21: can we actually use this PR to just set a "Feature Flag" feature flag boolean in the ENV? Then we'd be able to wrap the nav item / route in flag condition and start working on Feature-Flag in the 'dark', whatever we end up doing with existing functionality.

danoswaltCL: I have no idea what to name it though. FEATURE_FLAG_FEATURE_FEATURE_FLAG 😵‍💫

PR #1338, opened 2024-03-05
Solve segment status got changed on refresh

RidhamShah:
Added new API for getSegmentWithStatusById which return segment data with status


danoswaltCL (approved):
💯

PR #1343, opened 2024-03-06
Storing proper group exclusion and individual exclusion documents

### LINKED ISSUE #1318: Exclusion by group in segment is not working correctly for all the experiments
**Describe the bug**

When the user gets excluded due to groups in the segments the exclusion documents are not getting created properly.

Currently, we are creating a group exclusion document for all the experiment types that don't make sense for Individual and within-subject experiments. For group experiments, this document should only be created if the `experiment.group` is the same as the group type because of which the user is excluded. Else only an individual exclusion document should be created for the user.

**To summarise**
Individual/Within-subject experiment
- Only individual exclusion documents should be created.

Group Experiment
- Individual Exclusion documents should be created.
- A group Exclusion document should be created only if the experiment group is the same as the segment's group because of which the user got excluded

ppratikcr7:
This PR resolves #1318 

The `GROUP_ON_EXCLUSION_LIST` experiment level exclusion wont work due to experiment.group = null for within subject experiments.

Below are the changes done:
1. For a group experiment, a group Exclusion document should be created only if the experiment group is the same as the segment's group because of which the user got excluded
2. If a group (school1) was included earlier and we change the exclusion list and add group (school1), we should delete the group enrollment doc from the db table.
3. We should also store the Individual Exclusion doc as well with the Group Exclusion doc.
4. Code optimization for mark and assign call involving exclusions.

### DIFF
--- a/backend/packages/Upgrade/src/api/repositories/GroupEnrollmentRepository.ts
+++ b/backend/packages/Upgrade/src/api/repositories/GroupEnrollmentRepository.ts
@@ -1,5 +1,7 @@
 import { EntityRepository, In, Repository } from 'typeorm';
 import { GroupEnrollment } from '../models/GroupEnrollment';
+import repositoryError from './utils/repositoryError';
+import { UpgradeLogger } from 'src/lib/logger/UpgradeLogger';
 
 @EntityRepository(GroupEnrollment)
 export class GroupEnrollmentRepository extends Repository<GroupEnrollment> {
@@ -8,5 +10,21 @@ export class GroupEnrollmentRepository extends Repository<GroupEnrollment> {
       where: { experiment: { id: In(experimentIds) }, groupId: In(groupIds) },
       relations: ['experiment', 'condition'],
     });
+  }
+
+  public async deleteGroupEnrollment(id: string, logger: UpgradeLogger): Promise<GroupEnrollment[]> {
+    const result = await this.createQueryBuilder()
+      .delete()
+      .from(GroupEnrollment)
+      .where('groupId=:id', { id })
+      .returning('*')
+      .execute()
+      .catch((errorMsg: any) => {
+        const errorMsgString = repositoryError('GroupEnrollmentRepository', 'deleteGroupEnrollment', { id }, errorMsg);
+        logger.error(errorMsg);
+        throw errorMsgString;
+      });
+
+    return result.raw;
   }
 }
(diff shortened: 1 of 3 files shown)


backend/packages/Upgrade/src/api/repositories/GroupEnrollmentRepository.ts
```diff
@@ -9,4 +11,20 @@ export class GroupEnrollmentRepository extends Repository<GroupEnrollment> {
       relations: ['experiment', 'condition'],
     });
   }
+
+  public async deleteGroupEnrollment(id: string, logger: UpgradeLogger): Promise<GroupEnrollment[]> {
+    const result = await this.createQueryBuilder()
+      .delete()
+      .from(GroupEnrollment)
+      .where('groupId=:id', { id })
+      .returning('*')
```
VivekFitkariwala, line 20: You can remove the returning part

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1311,45 +1302,40 @@ export class ExperimentAssignmentService {
...
           groupId: user?.workingGroup[experiment.group],
           experiment,
           exclusionCode: EXCLUSION_CODE.GROUP_ON_EXCLUSION_LIST,
         };
         await this.groupExclusionRepository.saveRawJson([excludeGroupDoc]);
-      } else {
-        const excludeUserDoc: Pick<IndividualExclusion, 'user' | 'experiment' | 'exclusionCode'> = {
-          user,
+        // check if excluded group was earlier included, if yes - remove them:
+        this.groupEnrollmentRepository.deleteGroupEnrollment(user?.workingGroup[experiment.group], new UpgradeLogger());
```
VivekFitkariwala, line 1323: await here.
For multiple async requests happening here you should use `Promise.all`
Same thing below

VivekFitkariwala (changes requested):

danoswaltCL:
@VivekFitkariwala @ppratikcr7 is this one still waiting on changes? would like to merge this one into the parent branch before merging into that into the release branch so I only have to sync up one release/5.1 update.

ppratikcr7:
> @VivekFitkariwala @ppratikcr7 is this one still waiting on changes? would like to merge this one into the parent branch before merging into that into the release branch so I only have to sync up one release/5.1 update.

The changes are done. Yeah if approved, we can merge this.

VivekFitkariwala (approved):

PR #1348, opened 2024-03-08
Added json editor code to add metrics

Yagnik56:
In this PR we are bring back metrics json editor to add metrics.

PR #1349, opened 2024-03-08
Bugfix/view exp exclude participants header name fixing

Yagnik56:
Header of exclude participant in view experiment were incorrect.
This PR resolve that issue and fix that issue.
image


bcb37:
Commit 0d239d0... Is on another pr (https://github.com/CarnegieLearningWeb/UpGrade/pull/1348), so it should not be on this one.

PR #1355, opened 2024-03-11
Preview Users data not showing on hard refresh

RidhamShah:
On hard refresh in the Preview-Users Tab, the data will not populate.
Updated the code so that on initialization the data call is created.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
@@ -9,13 +9,19 @@ import { StratificationFactorsService } from '../../../../core/stratification-fa
   styleUrls: ['./experiment-users-root.component.scss'],
 })
 export class ExperimentUsersRootComponent {
+  selectedTabIndex = 0;
   constructor(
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
```diff
@@ -9,13 +9,19 @@ import { StratificationFactorsService } from '../../../../core/stratification-fa
...
 export class ExperimentUsersRootComponent {
+  selectedTabIndex = 0;
   constructor(
     private experimentService: ExperimentService,
     private previewUsersService: PreviewUsersService,
     private stratificationFactorsService: StratificationFactorsService
   ) {}
 
+  ngOnInit() {
+    this.selectedTabChange({ index: this.selectedTabIndex });
```
VivekFitkariwala, line 20: This will be called once at the start so it will always called with `this.selectedTabIndex = 0`. So do we need to make this a class variable?

RidhamShah: code updated

VivekFitkariwala (approved):

PR #1356, opened 2024-03-11
debug statements in getExperimentCondition code

danoswaltCL:
Let me know if these logs look like reasonable spots to spy values for the `getExperimentCondition` NPEs. If so I'll publish this branch from my local to Nexus, no need to merge anything in.

### DIFF
--- a/clientlibs/java/pom.xml
+++ b/clientlibs/java/pom.xml
@@ -11,7 +11,7 @@
 
   <!-- 5.0.12 -> 5.0.13: no retries for 4xx, retry other errors once instead of 3 times -->
 
-	<version>5.0.13</version>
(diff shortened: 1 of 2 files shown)


clientlibs/java/pom.xml
```diff
@@ -11,7 +11,7 @@
 
   <!-- 5.0.12 -> 5.0.13: no retries for 4xx, retry other errors once instead of 3 times -->
 
-	<version>5.0.13</version>
+	<version>5.0.13-DEBUG</version>
```
danoswaltCL, line 14: This will work as a version, right?

jerith666:
I think I figured out the NPE without needing this additional logging.  But we can revisit if my theory turns out to be wrong.  :)

PR #1357, opened 2024-03-12
Solve UI bugs of validation loading spinner

RidhamShah:
Solved

For Import Experiment
- Swagger for validation and import now called through POST
- Swagger is updated to contain details of API
- In UI, the import button for experiments will be disabled until validations are completed
- Spinner will be shown while validations are processing for import
- FIXED - Spamming the import button leads to multiple import calls resulting in multiple imports of the same files.

For import Segment
- In UI, the import button for experiments will be disabled until validations are completed
- Created spinner while validations are processing for import
- FIXED - Spamming the import button leads to multiple import calls resulting in multiple imports of the same files.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
@@ -85,8 +85,6 @@ export class ImportExperimentComponent implements OnInit {
   }
 
   async uploadFile(event) {
-    const index = 0;
-    const reader = new FileReader();
     this.uploadedFileCount = event.target.files.length;
     this.importFileErrors = [];
     this.allExperiments = [];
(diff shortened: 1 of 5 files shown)


frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
```diff
@@ -97,11 +96,11 @@ export class ImportExperimentComponent implements OnInit {
     // Set loading to true before processing the files
     this.isLoadingExperiments$ = true;
 
-    const readFile = (fileIndex) => {
+    const readFile = async (fileIndex) => {
```
VivekFitkariwala, line 99: This is loading the files one after another. You can use `map` on `event.target.files` over here to load them together.

VivekFitkariwala (dismissed):
I have added one improvement in the comment.

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/import-experiment/import-experiment.component.ts
RidhamShah: code updated

VivekFitkariwala (approved):

Issue #1382, opened 2024-03-19
Java client lib v5 expects `experimentType` in assign response, but that doesn't exist.


danoswaltCL:
**Version where bug was found:**
5.1

**Describe the bug**
Java Client Library expect `experimentType` in assignment response, but `experimentType` is not a field on the response.

**To Reproduce**
Run Main.java and see `getExperimentType` always return null. Observe that `experimentType` is not a field in the response.

**Fix**
Could update Java lib to figure this out via using the existence of the `assignedFactor` property for now, which is how the TypeScript library does it. A more thorough fix is probably to just send `experimentType` in the assignment response, but that may be a bit more work and potentially an API breaking change.

ppratikcr7:
@danoswaltCL Should we close this, as its moved to Done Column.

PR #1383, opened 2024-03-19
Bugfix/1382 experiment type assign response

danoswaltCL:
#1382

This change adds the `experimentType` string enum to the `IExperimentAssignmentv5` type and the response from `/assign` (as well as the responses created by the within-subjects algorithms). This is an expected field for the Java client, and really should be expected in TS client.

If this looks acceptable and won't cause issues elsewhere to modify the API, I will create a separate ticket to update the v5 TypeScript library to get `experimentType` directly from the server and not use `this._experimentType = assignedFactor ? EXPERIMENT_TYPE.FACTORIAL : EXPERIMENT_TYPE.SIMPLE;`, which is not very future-proof considering that there could be more than these 2 experiment types at some point.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "ab_testing_backend",
-    "version": "5.0.13",
+    "version": "5.0.12",
     "description": "Backend for A/B Testing Project",
     "scripts": {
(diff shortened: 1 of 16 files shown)


danoswaltCL:
oops wrong release

PR #1384, opened 2024-03-19
Bugfix/1382 experiment type in assign response

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/issues/1382

This change adds the `experimentType` string enum to the `IExperimentAssignmentv5` type and the response from `/assign`, as well as the responses created by the within-subjects algorithms. This is an expected field for the Java client, and really should be expected in TS client.

If this looks acceptable and won't cause issues elsewhere to modify the API, I will create a separate ticket to update the v5 TypeScript library to get experimentType directly from the server and not use `this._experimentType = assignedFactor ? EXPERIMENT_TYPE.FACTORIAL : EXPERIMENT_TYPE.SIMPLE;`, which seems bound to break in the future anyhow.

Please let me know what tests should also be updated, I'm not very familiar with that test suite. I was surprised this didn't break anything to add this.

### DIFF
--- a/backend/packages/Upgrade/src/api/Algorithms.ts
+++ b/backend/packages/Upgrade/src/api/Algorithms.ts
@@ -61,11 +61,12 @@ export function randomCondition(
     }
   }
 
-  const randomAssignData = {
+  const randomAssignData: IExperimentAssignmentv5 = {
     site: assignedData.site,
     target: assignedData.target,
     assignedCondition: randomConditionArray,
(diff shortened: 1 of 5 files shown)


VivekFitkariwala:
The PR looks good. You can also include the changes to the JS client in this PR.

bcb37 (dismissed):

bcb37 (approved):

PR #1394, opened 2024-03-26
added the feature of search in segment section

Yagnik56:
In segment section, search feature wasn't working as it wasn't implemented on input text-box. It is fixed now, search can be used now for all or based on particular value like name, context or status.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
@@ -87,7 +87,7 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
-  filterExperimentPredicate(type: EXPERIMENT_SEARCH_KEY) {
+  filterSegmentPredicate(type: EXPERIMENT_SEARCH_KEY) {
     this.allSegments.filterPredicate = (data, filter: string): boolean => {
       switch (type) {
         case EXPERIMENT_SEARCH_KEY.ALL:
@@ -107,15 +107,10 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   applyFilter(filterValue: string) {
-    this.filterExperimentPredicate(this.selectedSegmentFilterOption);
+    this.filterSegmentPredicate(this.selectedSegmentFilterOption);
     if (typeof filterValue === 'string') {
       this.allSegments.filter = filterValue.trim().toLowerCase();
     }
-  }
-
-  filterExperimentByChips(tagValue: string, type: EXPERIMENT_SEARCH_KEY) {
-    this.selectedSegmentFilterOption = type;
-    this.applyFilter(tagValue);
   }
 
   openNewSegmentDialog() {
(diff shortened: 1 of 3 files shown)


bcb37 (commented):
This works, but I wonder if we can encapsulate the filter logic into a reusable component - (I noticed that some of the method names were not changed when copied over to reflect the change from 'experiment' to 'filter')

The primary reason to make a filter component is that we're going to also need the same UI/functionality on the feature flags page, and we'd probably want to make sure all three (or more, possibly, in the future?) filters stay in sync if there are any changes later on.

Yagnik56:
@bcb37  there are few things which aren't same like achieved experiment and all aren't filtered on same properties so let's keep it this was for now, so this functionality start working. We will dig deep in UI feature flags but we are still figuring out how to implement reusable component and it will take a while to implement it so let's not keep this in hold.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
```diff
@@ -69,13 +79,45 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  }
+
+  applyFilter(filterValue: string) {
+    this.filterExperimentPredicate(this.selectedSegmentFilterOption);
+    if (typeof filterValue === 'string') {
+      this.allSegments.filter = filterValue.trim().toLowerCase();
+    }
+  }
+
+  filterExperimentByChips(tagValue: string, type: EXPERIMENT_SEARCH_KEY) {
```
bcb37, line 116: This method is not used. If we want to use it, we should add the actions to the template for 'context' and 'tags', the same as in the experiment template, and also change its name to `filterSegmentByChips()`

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
```diff
@@ -69,13 +79,45 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
...
-      this.allSegments = new CustomMatTableSource();
+      this.allSegments = new MatTableDataSource();
       this.allSegments.data = [...segments];
       this.allSegments.sort = this.sort;
       this.cdr.detectChanges();
+      this.applyFilter(this.searchValue);
     });
   }
 
+  filterExperimentPredicate(type: EXPERIMENT_SEARCH_KEY) {
```
bcb37, line 90: Change this to `filterSegmentPredicate()`

bcb37 (changes requested):
Ok, I'm fine with keeping the duplicated functionality for now. Just a couple more issues to address.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
Yagnik56: done

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
Yagnik56: removed that func for now, done

bcb37: I want to check with product to make sure we don't want to have the filter by chaps functionality. I suspect we probably do.

Yagnik56: I am not sure, if we do and if you think we won't have to redo same in new design can you create separate ticket as this PR target issue of search functionality not working. We had discussed same in our meeting, and there were few other functionality I can add but since we are about start for new design and issue was about search functionality not working so address that only in this PR.

bcb37 (approved):

Issue #1345, opened 2024-03-07
Changes to feature flag admin endpoints


bcb37:
The endpoints used to create, view, and edit feature flags need to be updated to reflect the new functionality.

The existing endpoints support CRUD operations (including a POST endpoint to update a flag's 'status').

- The request and response values should be altered to include participant/segment information, and app-context.
- The DB access methods should be changed to store and retrieve this information from the updated db schema.
- Evaluate whether the existing set of endpoints meets the current needs.

amurphy-cl:
Use the read-replica DB for read methods

bcb37:
QA: I did this so someone else should probably QA it

VivekFitkariwala:
The Lists in the feature flag is using the context of the Feature Flag in which they are created. If the Feature Flag is edited and the context is changed the context value of the lists are not updated.
At present the functionality of the feature flags are not getting affected but we are having an inconsistent state in the database.
An easy way to fix this is to not store the context data for the Private segments.

@bcb37 @RidhamShah

VivekFitkariwala:
Segment status is unused even if the segment is used in the Feature Flag list.

ppratikcr7:
This is now covered in 2 separate tickets (#1959 and #1962). We can close this now.

Issue #1307, opened 2024-02-13
Backend Feature flag API endpoint


amurphy-cl:
Current endpoint `featureflag` only has GET

Need to change to POST, e.g. 
POST /api/v5/featureflags

**Acceptance criteria**

Request should have user and app context

Response should be list of feature flags TRUE for a current user 
[
  "AREA_MOTIVATIONAL_SUPPORT",
  "USE_AREA_PROBLEM"
]

PR #1395, opened 2024-03-27
Feature/feature flag db changes

bcb37:
- add feature_flag_segment_inclusion table
- add feature_flag_segment_exclusion table
- add feature_flag_exposure table
- changes to feature_flag, segment entities


backend/packages/Upgrade/src/api/controllers/validators/FeatureFlagValidator.ts
```diff
@@ -1,5 +1,15 @@
...
+  IsOptional,
+  ValidateNested,
+  IsString,
+  IsBoolean,
+  IsArray,
+  IsEnum,
+} from 'class-validator';
+import { FEATURE_FLAG_STATUS } from 'upgrade_types';
 
 class FlagVariationValidator {
```
RidhamShah, line 14: FeatureFlag Validators should now contain `context` array validations

bcb37: I can make that change, but I thought we could leave the API changes that are not necessary for passing the CI to the API tickets: https://github.com/CarnegieLearningWeb/UpGrade/issues/1345 and https://github.com/CarnegieLearningWeb/UpGrade/issues/1307 

There will be many more changes necessary to complete those.

RidhamShah (approved):

VivekFitkariwala (approved):

PR #1403, opened 2024-03-29
Solved multiple instance of experiments issue

RidhamShah:
Solved the following issue for StratificationFactor Used Tooltip
- Only one instance of the experiment is shown
- Instead of experimentId, experimentName is shown in the tooltip
- Updated spelling errors in stratificationFactor

Before:
image
After
image


frontend/projects/upgrade/src/app/features/dashboard/experiment-users/components/stratification-factor-list/stratification-factor-list.component.ts
```diff
@@ -54,11 +64,14 @@ export class StratificationComponent implements OnInit {
 
         factorSummary = totalUsers.toString() + ' ' + factorSummary + ' (' + tempSummary + ')';
 
+        const experimentNamesToShow = stratificationFactor.experimentIds.map((experimentId) => {
```
VivekFitkariwala, line 67: Use `filter` instead of `map`

VivekFitkariwala (approved):

VivekFitkariwala (approved):

PR #1405, opened 2024-04-01
add env flag for toggling feature flags UI nav

danoswaltCL:
This change will allow us to toggle on feature-flags navigation access in UI.

The boolean flag `featureFlagNavToggle` will be first checked for existence on the `environment.json` object that is fetched by the UI on app load. This will allow us to control this flag without a rebuild/redeploy.

If it is `undefined/null` it will fall back to the env configuration used for that build and look for a `featureFlagNavToggle` there. This will allow us to set a state in the build config as the default for that build.

If both are not defined, fall back to `false`.

Note, this uses the null-coalesce `??` operator instead of OR `||`, because with a boolean `false` is a valid value.

### DIFF
--- a/frontend/projects/upgrade/src/app/app.module.ts
+++ b/frontend/projects/upgrade/src/app/app.module.ts
@@ -29,7 +29,7 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
       .then((config: RuntimeEnvironmentConfig) => {
         env.apiBaseUrl = config.endpointApi || config.apiBaseUrl;
         env.googleClientId = config.gapiClientId || config.googleClientId;
-        env.featureFlagNavToggle = config.featureFlagNavToggle ?? env.featureFlagNavToggle;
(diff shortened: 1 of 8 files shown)


frontend/projects/upgrade/src/app/app.module.ts
```diff
@@ -29,6 +29,7 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
       .then((config: RuntimeEnvironmentConfig) => {
         env.apiBaseUrl = config.endpointApi || config.apiBaseUrl;
         env.googleClientId = config.gapiClientId || config.googleClientId;
+        env.featureFlagNavToggle = config.featureFlagNavToggle ?? env.featureFlagNavToggle ?? false;
```
VivekFitkariwala, line 32: Do we need this line?

danoswaltCL: Yes if we want to be able to toggle this flag at runtime in a particular environment without rebuild/redeploys, this was the only way I can think to do that. That will override the build config if present, and be false if it is not present in build config either.

VivekFitkariwala (approved):

PR #1408, opened 2024-04-02
🐛 Fix for integration test cases with old Decision Point keys & tests for Enrollment Codes 

### LINKED ISSUE #1407: ✨ Integration test case utils function checkExperimentAssignedIsNull() has old decision point keys {expId, expPoint} object check
e.g "5.1"

**Describe the bug**
Integration test case utililty function `checkExperimentAssignedIsNull()` has old decision point keys {expId, expPoint} in object check. We should use {target, site}. This will affect the exclusion checks for various test cases as all were passing this function as this function uses != object [comparison](https://github.com/CarnegieLearningWeb/UpGrade/blob/27367f575ccdc3917df43f893de74084b0868c86/backend/packages/Upgrade/test/integration/utils/index.ts#L17).

ppratikcr7:
This PR resolves #1407 

The test cases are updated mainly due to the old Decision Point keys { expId, expPoint } used in the checkExperimentAssignedIsNull() function instead of { target, site }

There were 3 more test cases related to Enrollment codes that were commented, which are now revived with proper enrollment code checks.  

We would be building the exclusions code test cases on top of these test cases.


backend/packages/Upgrade/test/integration/Experiment/enrollmentCode/ExperimentExperimentEnrollmentCode.ts
```diff
@@ -1,259 +1,269 @@
...
+export default async function testCase(): Promise<void> {
+  const experimentService = Container.get<ExperimentService>(ExperimentService);
+  const userService = Container.get<UserService>(UserService);
+
+  // creating new user
+  const user = await userService.upsertUser(systemUser as any, new UpgradeLogger());
+
+  // experiment object
+  const experimentObject = individualAssignmentExperimentConsistencyRuleExperiment;
+  experimentObject.partitions = [
```
VivekFitkariwala, line 24: This is the same partition coming from the `individualAssignmentExperimentConsistencyRuleExperiment`. Why are you adding it again?

backend/packages/Upgrade/test/integration/ExperimentStats/DetailGroupExperiment.ts
```diff
@@ -185,8 +185,6 @@ export default async function testCase(): Promise<void> {
   // user 2 logs in experiment
   // get all experiment condition for user 2
   experimentConditionAssignments = await getAllExperimentCondition(experimentUsers[1].id, new UpgradeLogger());
-  checkExperimentAssignedIsNull(experimentConditionAssignments, experimentName1, experimentPoint1);
```
VivekFitkariwala, line 188: This function was checking no assignment. How is that getting checked now?

VivekFitkariwala (changes requested):
I have added a few comments.

backend/packages/Upgrade/test/integration/Experiment/enrollmentCode/ExperimentExperimentEnrollmentCode.ts
ppratikcr7: I am setting excludeIfReached as true for these test cases.

backend/packages/Upgrade/test/integration/ExperimentStats/DetailGroupExperiment.ts
ppratikcr7: Yes. So, this `checkExperimentAssignedIsNull` function was not working as expected due to not matching to the correct DP keys. This function was not supposed to be called for the user2 as this user is not excluded. Instead we can update this to check no assignment for user1, as this user came before enrolling state and would be excluded.

backend/packages/Upgrade/test/integration/ExperimentStats/GroupEnrollment.ts
```diff
@@ -93,8 +93,6 @@ export default async function testCase(): Promise<void> {
   // user 2 logs in experiment
   // get all experiment condition for user 2
   experimentConditionAssignments = await getAllExperimentCondition(experimentUsers[1].id, new UpgradeLogger());
-  checkExperimentAssignedIsNull(experimentConditionAssignments, experimentName1, experimentPoint1);
```
VivekFitkariwala, line 96: Check that the assignments are not null

backend/packages/Upgrade/test/integration/ExperimentStats/IndividualEnrollment.ts
```diff
@@ -29,7 +29,28 @@ export default async function testCase(): Promise<void> {
   const user = await userService.upsertUser(systemUser as any, new UpgradeLogger());
   // experiment object
   const experimentObject = individualExperimentStats;
-
+  experimentObject.partitions = [
```
VivekFitkariwala, line 32: Make a util function to update the flag

backend/packages/Upgrade/test/integration/ExperimentStats/IndividualEnrollment.ts
```diff
@@ -115,29 +136,7 @@ export default async function testCase(): Promise<void> {
     ])
   );
 
-  experimentConditionAssignments = await getAllExperimentCondition(experimentUsers[0].id, new UpgradeLogger());
```
VivekFitkariwala, line 118: Why remove this code? It is checking that if the user comes back again it won't get any assignment. By removing this code the test case is no longer checking that scenario. You did this in many test cases please fix them

backend/packages/Upgrade/test/integration/ExperimentStats/WithinSubjectEnrollment.ts
```diff
@@ -29,7 +29,28 @@ export default async function testCase(): Promise<void> {
   const user = await userService.upsertUser(systemUser as any, new UpgradeLogger());
   // experiment object
   const experimentObject = withinSubjectExperiment;
-
+  experimentObject.partitions = [
```
VivekFitkariwala, line 32: Use utils function

backend/packages/Upgrade/test/integration/mockData/experiment/index.ts
```diff
@@ -135,6 +135,62 @@ export const withinSubjectExperiment = {
   consistencyRule: null,
 };
 
+export const individualLevelEnrollmentCodeExperiment = {
```
VivekFitkariwala, line 138: Use util function

VivekFitkariwala (changes requested):

backend/packages/Upgrade/test/integration/ExperimentStats/DetailGroupExperiment.ts
ppratikcr7: Okay, we use the function `checkExperimentAssignedIsNotDefault` to check if the assigned condition is not default condition everywhere in test cases. Will add that.

backend/packages/Upgrade/test/integration/Experiment/conditionalStateChange/GroupUserCount.ts
```diff
@@ -229,13 +208,6 @@ export default async function GroupUserCount(): Promise<void> {
   );
   checkMarkExperimentPointForUser(markedExperimentPoint, experimentUsers[4].id, experimentName, experimentPoint);
 
-  // get the number of group satisfied as 2
```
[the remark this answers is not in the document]
ppratikcr7: This is getting checked at the end, there was a change in number of groups satisfied so this check is removed from here and is checked at the end of the test case.

backend/packages/Upgrade/test/integration/ExperimentStats/IndividualEnrollment.ts
ppratikcr7: Okay, added this.

ppratikcr7:
@VivekFitkariwala Can you please re-review the PR. I have addressed all of your comments in the last commit.

Issue #1431, opened 2024-04-09
BugFix: Java 5.1.4 would send bad request in prod (5.0.13)


danoswaltCL:
Spawned from testing #996, Java 5.1.4 would send bad request in prod (5.0.13).

A bugfix went into the `/mark` controller in 5.1 to do null checking on the `data.assignedCondition.conditionCode`, because the Java 5.1 lib could potentially send in `data.assignedCondtion` as `null`. Unfortunately this still exists in prod.

In retrospect, the proper fix would probably have been what to also do I'll do in the PR here, which is to ensure that the Java lib is always sending a proper empty `Condition` object in the `MarkRequestBody`, no matter if 'null' is sent by client for the assignedCondition, so that the Java lib is unable to send a bad request.

Issue #996, opened 2023-09-08
QA task: test condition assignment for factorial experiment in staging env


amurphy-cl:
Test the factorial experiment Factorial AA test F1 - Java client library for MATHia. Participants should receive appropriate condition assignments

danoswaltCL:
https://github.dev.carnegielearning.com/CarnegieLearning/ctp/pull/891/files <-- PR for Mathia changes

https://github.dev.carnegielearning.com/CarnegieLearning/ctp/blob/mathia-2023.8.61/runtime/services/services-business/services-section-variant-bl/src/main/java/cl/services/ct/sectionvariant/bl/SectionVariantExperimentServiceImpl.java <-- This is where assign and mark are called

danoswaltCL:
Tested manually marking the Factorial assignment, I get an error.

#1001

amurphy-cl:
When trying this, ask Matt to kick off the branch deployment.

PR #1432, opened 2024-04-09
assignedCondition should never be null, always be a valid Condition

danoswaltCL:
This should unblock https://carnegielearning.atlassian.net/browse/STU-2269

See #1431 found during #996, a few things have been related to this but the root of all evil was always that we are allowing the Java client to send in an unproccessable request by allowing 2 different ways for `assignedCondition` to contain `null` instead of a `Condition`, either by using the default `new MarkExperimentRequest(site, target)` or by explicitly setting it, which we allow. `new MarkExperimentRequest(site, target, null)` is valid but would break.

The change is to always have a default 'blank' `new Condition()` object in there when there is no condition, which is the right way for `/mark` to consume 'no assignment' requests. (this is how the TypeScript library works, which is why this issue doesn't happen there).

Earlier in this same thread of tests I fixed this on the API side in 5.1 (#1366) to close the hole there with `assignedCondition` being undefined / null, but the Java lib should have been modified at that time also to just never even be able to do that.

This was sufficient for API v5.1, but API v5.0 does not have that fix. so the 5.1 Java lib is blocked from prod with v5.0 in there.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "ab_testing_backend",
-    "version": "5.1.4",
+    "version": "5.1.5",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "ab_testing_backend",
-            "version": "5.1.4",
+            "version": "5.1.5",
             "license": "ISC",
             "dependencies": {
                 "dayjs": "^1.11.10",
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "ab_testing_backend",
-    "version": "5.1.4",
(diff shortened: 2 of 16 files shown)


jerith666 (commented):
There's a setter for this field that also needs to defend against `null`.  And you need a version bump in the `pom.xml` file.

danoswaltCL:
> There's a setter for this field that also needs to defend against `null`. And you need a version bump in the `pom.xml` file.

done

bcb37 (approved):

bcb37 (approved):

danoswaltCL:
oh i based this PR against the wrong release, 5.1 is the next release, needs to go against 5.0 and be published as 5.0.14. Closing this PR, will reopen.

PR #1438, opened 2024-04-11
use new relic env var

danoswaltCL:
adding a new relic env var, small logical refactor to simplify enabling new relic. Will be able to use new relic in staging. Making it a 'bugfix' into release branch.

Will bump the version in the code once reviewed and approved (so you're gonna need to approve twice 😀 )


kcalvo-cli (approved):
LGTM

PR #1441, opened 2024-04-11
no assignedCondition null in java lib

danoswaltCL:
Same as https://github.com/CarnegieLearningWeb/UpGrade/pull/1432/files but against the correct release, got myself backwards, 5.0 is the target not 5.1, technically we're not ready for 5.1 yet. although we'll need to merge this change down into 5.1 and then down into dev as well.

### DIFF
--- a/clientlibs/java/pom.xml
+++ b/clientlibs/java/pom.xml
@@ -9,9 +9,7 @@
 		at the same time that happen to rev to the same new version will be caught 
 		by a merge conflict. -->
 
-  <!-- 5.0.12 -> 5.0.13: no retries for 4xx, retry other errors once instead of 3 times -->
-
-	<version>5.0.13</version>
+	<version>5.0.14</version>
 	<build>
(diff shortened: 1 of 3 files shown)


danoswaltCL:
btw that automerge job is just confused because it expects to merge into dev. This job was removed in subsequent releases.

bcb37 (approved):

clientlibs/java/pom.xml
```diff
@@ -9,9 +9,7 @@
 		at the same time that happen to rev to the same new version will be caught 
 		by a merge conflict. -->
 
-  <!-- 5.0.12 -> 5.0.13: no retries for 4xx, retry other errors once instead of 3 times -->
-
-	<version>5.0.13</version>
+	<version>5.0.14</version>
```
jerith666, line 12: I think it's worthwhile to keep the comments, it guards against two different PRs trying to claim the same version at the same time.

PR #1443, opened 2024-04-12
Added appContext in Metrics

RidhamShah:
**Completed**

- Updated Metric model to add context as an array
- Updated Migration file
- Update CRUD operations
- Added a new API to get metrics only having the given context.
- Updated stepper to show metrics belonging to the context

**Remaining**

- create a way to add context in the JSON editor of metrics.
- what should be the context in pre-created metrics and metrics from .env file
- Update test cases for the changes


RidhamShah:
The PR is no longer needed as its context is shift to another [PR](https://github.com/CarnegieLearningWeb/UpGrade/pull/1472).

PR #1445, opened 2024-04-12
Segment context chips in list and view comp feature added

Yagnik56:
This PR add feature of chips on segment context functional by clicking on them search filter is set on that context.


frontend/projects/upgrade/src/app/core/segments/segments.service.ts
```diff
@@ -9,21 +9,43 @@ import {
...
+  selectSortKey,
+  selectSortAs,
 } from './store/segments.selectors';
-import { SegmentInput, UpsertSegmentType } from './store/segments.model';
-import { filter, map, tap } from 'rxjs/operators';
+import { SegmentInput, SegmentLocalStorageKeys, UpsertSegmentType } from './store/segments.model';
+import { filter, first, map, tap } from 'rxjs/operators';
 import { Observable, combineLatest } from 'rxjs';
 import { SegmentsDataService } from './segments.data.service';
+import { SEGMENT_SEARCH_KEY, SEGMENT_SORT_AS, SEGMENT_SORT_KEY } from '../../../../../../../types/src/Experiment/enums';
```
bcb37, line 21: Is there a reason we're not importing these from 'upgrade_types', like we do for experiments?

frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
```diff
@@ -1,6 +1,11 @@
 import { AppState } from '../../core.state';
 import { EntityState } from '@ngrx/entity';
 import { SEGMENT_TYPE, SEGMENT_STATUS } from 'upgrade_types';
+import {
+  SEGMENT_SEARCH_KEY,
+  SEGMENT_SORT_AS,
+  SEGMENT_SORT_KEY,
+} from '../../../../../../../../types/src/Experiment/enums';
```
bcb37, line 8: For consistency, I think we can import these three from 'upgrade_types' and then export them, so that everything's coming from the model, like it does with experiments. We assume that the `types/` directory is copied from the top level into frontend/packages/Upgrade/ as part of every build.

bcb37 (dismissed):
This is still not quite the same as the way it's done with experiments, where the constants are re-exported from the model and imported from the model in the services. But it's not clear to me why that was done, so this is probably fine.

Yagnik56:
> This is still not quite the same as the way it's done with experiments, where the constants are re-exported from the model and imported from the model in the services. But it's not clear to me why that was done, so this is probably fine.

@bcb37 can you approve PR again as there was merge conflict I had to resolve so your approval dismissed

frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
```diff
@@ -55,6 +55,35 @@ export class SegmentsEffects {
...
 
+  fetchSegmentOnSearchString$ = createEffect(
+    () =>
+      this.actions$.pipe(
+        ofType(SegmentsActions.actionSetSearchString),
+        map((action) => action.searchString),
+        tap((searchString) => {
+          // Allow empty string as we erasing text from search input
+          if (searchString !== null) {
+            this.store$.dispatch(SegmentsActions.actionFetchSegments({ fromStarting: true }));
```
VivekFitkariwala, line 66: We don't have any pagination for Segment so this fetch should not be triggered on change in the Search string

frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
```diff
@@ -55,6 +55,35 @@ export class SegmentsEffects {
...
+  );
+
+  fetchSegmentOnSearchKeyChange$ = createEffect(
+    () =>
+      this.actions$.pipe(
+        ofType(SegmentsActions.actionSetSearchKey),
+        withLatestFrom(this.store$.pipe(select(selectSearchString))),
+        tap(([_, searchString]) => {
+          if (searchString) {
+            this.store$.dispatch(SegmentsActions.actionFetchSegments({ fromStarting: true }));
```
VivekFitkariwala, line 80: Same as above

frontend/projects/upgrade/src/app/core/segments/segments.service.spec.ts
```diff
@@ -11,6 +11,7 @@ const MockStateStore$ = new BehaviorSubject({});
 describe('SegmentService', () => {
   let mockStore: any;
   let segmentsDataService: any;
+  let localStorageService: any;
```
VivekFitkariwala, line 14: any -> LocalStorageService

frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.ts
```diff
@@ -170,6 +170,7 @@ export class ExperimentListComponent implements OnInit, OnDestroy, AfterViewInit
   }
 
   filterExperimentByChips(tagValue: FLAG_SEARCH_SORT_KEY, type: EXPERIMENT_SEARCH_KEY) {
+    this.searchValue = tagValue;
```
VivekFitkariwala, line 173: You already have a subscription for this change so don't set the value directly here.

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
```diff
@@ -113,6 +126,32 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  }
+
+  changeSorting(event) {
+    this.segmentsService.setSortingType(event.direction ? event.direction.toUpperCase() : null);
+    this.segmentsService.setSortKey(event.direction ? event.active : null);
+    this.segmentsTableContainer.nativeElement.scroll({
+      top: 0,
+      behavior: 'smooth',
+    });
+    this.segmentsService.fetchSegments(true);
```
VivekFitkariwala, line 144: Is this required as we are fetching all the segments upfront?

frontend/projects/upgrade/src/app/features/dashboard/segments/components/segments-list/segments-list.component.ts
```diff
@@ -113,6 +126,32 @@ export class SegmentsListComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  }
+
+  filterSegmentsByChips(tagValue: string, type: SEGMENT_SEARCH_KEY) {
```
VivekFitkariwala, line 147: This should dispatch the action only.

VivekFitkariwala (changes requested):

frontend/projects/upgrade/src/app/core/segments/segments.service.spec.ts
Yagnik56: To mock we use any in all testcases in frontend and backend, can't give specific type like Localstorage as we aren't going to created all properties for `localStorageService` in `segmentService` testcase.

VivekFitkariwala (approved):

PR #1457, opened 2024-04-17
Resolve issue of user coming as undefined in audit logs

Yagnik56:
This PR address issue of user coming as undefined in audit logs.
Solution:
If user going as undefined, keep auth checker always on so if upgrad can't authenticate user it will log out user and audit logs won't get user as undefined.

### DIFF
--- a/backend/packages/Upgrade/test/unit/controllers/ExperimentController.test.ts
+++ b/backend/packages/Upgrade/test/unit/controllers/ExperimentController.test.ts
@@ -89,17 +89,6 @@ describe('Experiment Controller Testing', () => {
     },
   };
 
-  const mockUser: User = {
-    email: 'test@user.com',
-    firstName: 'test',
-    lastName: 'user',
-    role: UserRole.READER,
-    versionNumber: 5,
-    imageUrl: '',
-    createdAt: new Date(),
-    updatedAt: new Date(),
-  };
-
   test('Get request for /api/experiments', () => {
     return request(app).get('/api/experiments').expect('Content-Type', /json/).expect(200);
   });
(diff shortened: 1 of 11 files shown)


VivekFitkariwala (changes requested):
You are removing the environment variable `authCheck` from the code base. So also remove it from the file 
`env.ts`
`env.example`
`env.text`

Check that there is no usage of `authCheck` anywhere.

At a comment in the Ticket describing the change

backend/packages/Upgrade/test/unit/controllers/ExperimentController.test.ts
```diff
@@ -6,11 +6,12 @@ import { Container } from 'typedi';
 import { v4 as uuid } from 'uuid';
 import ExperimentServiceMock from './mocks/ExperimentServiceMock';
 import { ExperimentService } from '../../../src/api/services/ExperimentService';
-
+import { UserRole } from 'upgrade_types';
```
VivekFitkariwala, line 9: I don't see it being used in the file

backend/packages/Upgrade/test/unit/controllers/ExperimentController.test.ts
```diff
@@ -6,11 +6,12 @@ import { Container } from 'typedi';
...
 import { useContainer as classValidatorUseContainer } from 'class-validator';
 import { useContainer as ormUseContainer } from 'typeorm';
 import { ExperimentAssignmentService } from '../../../src/api/services/ExperimentAssignmentService';
 import ExperimentAssignmentServiceMock from './mocks/ExperimentAssignmentServiceMock';
+import { User } from 'src/api/models/User';
```
VivekFitkariwala, line 14: Same with this import

VivekFitkariwala (approved):

VivekFitkariwala (approved):

PR #1463, opened 2024-04-18
experiment list context chip issue is resolved

Yagnik56:
This PR will fix issue in experiment list context chip not working, as search input field wasn't getting context value without refreshing.


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.ts
```diff
@@ -172,6 +172,7 @@ export class ExperimentListComponent implements OnInit, OnDestroy, AfterViewInit
   }
 
   filterExperimentByChips(tagValue: FLAG_SEARCH_SORT_KEY, type: EXPERIMENT_SEARCH_KEY) {
+    this.searchValue = tagValue;
```
VivekFitkariwala, line 175: This is not a proper solution because you already have a subscriber for getting the searchValue but here you are setting the value directly.

VivekFitkariwala (changes requested):

PR #1471, opened 2024-04-22
✨ Toggle for within-subjects experiment type support

### LINKED ISSUE #1436: Add variable to .env file that allows toggle of experiment type
Add variable to .env file that allows us to toggle on/off experiment type. Currently needed for within-subjects experiments

ppratikcr7:
This PR resolves #1436

### DIFF
--- a/frontend/projects/upgrade/src/app/app.module.ts
+++ b/frontend/projects/upgrade/src/app/app.module.ts
@@ -21,7 +21,7 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
   }
 
   // in a prod build, we currently need to fetch environment.json at runtime
-  // to provide apiBaseUrl and googleClientId
+  // to provide apiBaseUr, googleClientId, featureFlagNavToggle and withinSubjectExperimentSupportToggle
   return () =>
     http
       .get('/environment.json')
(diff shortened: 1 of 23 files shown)


zackcl (dismissed):

frontend/projects/upgrade/src/app/app.module.ts
```diff
@@ -21,14 +21,15 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
   }
 
   // in a prod build, we currently need to fetch environment.json at runtime
-  // to provide apiBaseUrl and googleClientId
+  // to provide apiBaseUr, googleClientId, featureFlagNavToggle and withinSubjectExperimentSupportToggle
```
bcb37, line 24: I don't see the feature flag toggle. Are we adding that later?

frontend/projects/upgrade/src/app/app.module.ts
```diff
@@ -21,14 +21,15 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
...
   return () =>
     http
       .get('/environment.json')
       .toPromise()
       .then((config: RuntimeEnvironmentConfig) => {
         env.apiBaseUrl = config.endpointApi || config.apiBaseUrl;
         env.googleClientId = config.gapiClientId || config.googleClientId;
+        env.withinSubjectExperimentSupportToggle = config.withinSubjectExperimentSupportToggle ?? env.withinSubjectExperimentSupportToggle ?? false;
```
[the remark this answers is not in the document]
ppratikcr7: We may get the `withinSubjectExperimentSupportToggle` boolean from either config or env, else if both doesn't exist, then we return false. Your above code snippet only considers `withinSubjectExperimentSupportToggle` boolean from config.

frontend/projects/upgrade/src/app/app.module.ts
ppratikcr7: Actually featureFlagNavToggle is merged. But just checked, its in dev and not in release/v5.1 & it missed this comment. Probably we can add the comment later in dev.

VivekFitkariwala (dismissed):

danoswaltCL (approved):

PR #1473, opened 2024-04-24
Disabled dp and condition table and Allow payload edit

Yagnik56:
Issue: Decision point and condition table aren't disabled properly and save button in disabled so payload wasn't editable
This PR resolve this issue by disabling action button in decision point and conditions table and allowing payload to edit with enabling save button as well.


zackcl:
Thanks! I will check this out.

zackcl:
Sorry, I think I made a mistake. I will revert it back.

zackcl:
@Yagnik56 While your PR functions correctly, I have made a commit to grey out the Decision Points and Conditions/Factors table when the experiment is not editable. Please check if this has been done properly. Thank you!

zackcl (approved):

PR #1474, opened 2024-04-24
Feature flag admin endpoints

bcb37:
Adds (or updates) the following endpoints:

GET /flags
- get all feature flags

POST /flags
- create feature flag

GET /flags/{id}
- get flag by id

DELETE /flags/{id}
- delete flag by id

PUT /flags/{id}
- update flag by id

POST  /flags/paginated
- get all flags paginated

POST  /flags/status
- update flag status

The request (for POST and PUT) and response bodies for these endpoints can be viewed by looking at the swagger page.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
@@ -1,4 +1,4 @@
-import { JsonController, Authorized, Post, Body, CurrentUser, Delete, Param, Put, Req } from 'routing-controllers';
+import { JsonController, Authorized, Post, Body, CurrentUser, Delete, Param, Put, Req, Get } from 'routing-controllers';
 import { FeatureFlagService } from '../services/FeatureFlagService';
 import { FeatureFlag } from '../models/FeatureFlag';
 import { User } from '../models/User';
@@ -21,9 +21,8 @@ interface FeatureFlagsPaginationInfo extends PaginationResponse {
  *       - name
  *       - key
  *       - description
- *       - variationType
  *       - status
- *       - variations
+ *       - context
  *     properties:
  *       id:
  *         type: string
@@ -33,29 +32,101 @@ interface FeatureFlagsPaginationInfo extends PaginationResponse {
  *         type: string
  *       description:
  *         type: string
- *       variationType:
- *         type: string
  *       status:
- *         type: boolean
- *       variations:
- *           type: array
- *           items:
- *             type: object
- *             properties:
- *               value:
- *                type: string
- *               name:
- *                type: string
- *               description:
- *                type: string
- *               defaultVariation:
- *                type: boolean[]
+ *         type: string
+ *         enum: [archived, enabled, disabled]
+ *       context:
+ *         type: array
+ *         items:
+ *           type: string
+ *       tags:
+ *         type: array
+ *         items:
+ *           type: string
+ *       featureFlagSegmentInclusion:
+ *          type: object
+ *          properties:
+ *              segment:
+ *                type: object
+ *                properties:
+ *                  type:
+ *                    type: string
+ *                    example: private
+ *                  individualForSegment:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        userId:
+ *                          type: string
+ *                          example: user1
+ *                  groupForSegment:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        groupId:
+ *                          type: string
+ *                          example: school1
+ *                        type:
+ *                           type: string
+ *                           example: schoolId
+ *                  subSegments:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        id:
+ *                          type: string
+ *                        name:
+ *                          type: string
+ *                        context:
+ *                          type: string
+ *       featureFlagSegmentExclusion:
+ *          type: object
+ *          properties:
+ *              segment:
+ *                type: object
+ *                properties:
+ *                  type:
+ *                    type: string
+ *                    example: private
+ *                  individualForSegment:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        userId:
+ *                          type: string
+ *                          example: user1
+ *                  groupForSegment:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        groupId:
+ *                          type: string
+ *                          example: school1
+ *                        type:
+ *                           type: string
+ *                           example: schoolId
+ *                  subSegments:
+ *                    type: array
+ *                    items:
+ *                      type: object
+ *                      properties:
+ *                        id:
+ *                          type: string
+ *                        name:
+ *                          type: string
+ *                        context:
+ *                          type: string
  */
 
 /**
  * @swagger
  * flags:
- *   - name: Feature flags
+ *   - name: Feature Flags
  *     description: Get Feature flags related data
  */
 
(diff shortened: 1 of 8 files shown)


RidhamShah:
We also need FILTER_MODE (includeAll, excludeAll) in featureFlags for Segments logic too, as it is also present in Experiments.

if we don't want to use FILTER_MODE then we need to discuss how the Segments would work for it.

bcb37:
> We also need FILTER_MODE (includeAll, excludeAll) in featureFlags for Segments logic too, as it is also present in Experiments.
> 
> if we don't want to use FILTER_MODE then we need to discuss how the Segments would work for it.

@RidhamShah  I've created a [pr](https://github.com/CarnegieLearningWeb/UpGrade/pull/1496)  that adds FILTER_MODE to the db. When that's merged we can continue with this one.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -110,36 +140,119 @@ export class FeatureFlagService {
...
-        variations.length > 0 &&
-        variations.map((variation: FlagVariation) => {
-          variation.id = variation.id || uuid();
-          variation.featureFlag = featureFlagDoc;
-          return variation;
-        });
-
-      // saving variations
-      let variationDocs: FlagVariation[];
+      const setSegmentInEx = (inEx: FeatureFlagSegmentExclusion | FeatureFlagSegmentInclusion) => {
```
danoswaltCL, line 143: what does `inEx` mean? inclusion / exclusion?

bcb37: Yeah, I'd meant to eventually give that a better name. Good catch

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -64,6 +135,64 @@ interface FeatureFlagsPaginationInfo extends PaginationResponse {
...
+   *       responses:
+   *          '200':
+   *            description: Get Feature Flag By Id
+   *            schema:
+   *                $ref: '#/definitions/FeatureFlag'
+   *          '401':
+   *            description: AuthorizationRequiredError
+   *          '404':
+   *            description: Feature Flag not found
+   *          '500':
```
danoswaltCL, line 188: i'd expect this to be a 400 bad request if the id type is wrong

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -90,26 +219,27 @@ export class FeatureFlagsController {
...
    *                     type: string
    *                     enum: [ASC, DESC]
    *       tags:
-   *         - Feature flags
+   *         - Feature Flags
    *       produces:
    *         - application/json
    *       responses:
    *          '200':
    *            description: Get Paginated Experiments
```
danoswaltCL, line 240: Get Paginated Feature Flags

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -90,26 +219,27 @@ export class FeatureFlagsController {
...
    *                    key:
    *                     type: string
-   *                     enum: [name, key, status, variationType]
+   *                     enum: [name, key, status, updatedAt]
    *                    sortAs:
    *                     type: string
    *                     enum: [ASC, DESC]
    *       tags:
-   *         - Feature flags
+   *         - Feature Flags
```
danoswaltCL, line 235: Line 240 says Experiments instead of Feature Flags

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
danoswaltCL: I think we that's kind of missing from swagger in a lot of places, not just here (400 = bad request)

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -110,36 +140,119 @@ export class FeatureFlagService {
...
+              groups:
+                segment.groupForSegment?.map((x) => {
+                  return { type: x.type, groupId: x.groupId };
+                }) || [],
+              subSegmentIds: segment.subSegments?.map((x) => x.id) || [],
+            }
+          : inEx;
+      };
+
+      let includeSegmentExists = true;
```
danoswaltCL, line 159: I asked copilot to pull this out 159-223 into own function without the repetition for the segment include and segment exclude. It needs some help with types but I thought something like this looked pretty reasonable, what do you think?

```typescript
async function handleSegment(segmentInclusionExclusion: any, flagDoc: any, flag: any, type: string, logger: any) {

  let segmentExists = true;
  let segmentDoc: Segment;
  let segmentDocToSave: Partial<FeatureFlagSegmentInclusion | FeatureFlagSegmentExclusion> = {};
  if (segmentInclusionExclusion) {
    const segment: any = setSegmentInEx(segmentInclusionExclusion);

    const segmentData: SegmentInputValidator = {
      ...segment,
      id: segment.id || uuid(),
      name: flagDoc.id + ' ' + type + ' Segment',
      description: flagDoc.id + ' ' + type + ' Segment',
      context: flagDoc.context[0],
      type: SEGMENT_TYPE.PRIVATE,
    };
    try {
      segmentDoc = await this.segmentService.upsertSegment(segmentData, logger);
    } catch (err) {
      const error = err as ErrorWithType;
      error.details = 'Error in adding segment in DB';
      error.type = SERVER_ERROR.QUERY_FAILED;
      logger.error(error);
      throw error;
    }
    // creating segment doc
    const tempDoc = type === 'Inclusion' ? new FeatureFlagSegmentInclusion() : new FeatureFlagSegmentExclusion();
    tempDoc.segment = segmentDoc;
    tempDoc.featureFlag = flag;
    segmentDocToSave = this.getSegmentDoc(tempDoc);
  } else {
    segmentExists = false;
  }
  return { segmentExists, segmentDoc, segmentDocToSave };
}
```

Yagnik56:
@bcb37 this PR don't have 'get all feature flags' and 'get flag by id' API endpoints so are we going to add them? Also the interface of `FeatureFlagValidation` isn't updated so can you check that?

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
bcb37: I think it makes sense to reduce duplication. I'll see what I can do about wrangling the types.

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
bcb37: I agree that's probably a more correct error, but it looks like other places where we expect a UUID in a GET ('/experiments/single/:id', '/segments/status/:segmentId), we throw a 500. I'll follow that for now.

danoswaltCL (dismissed):

danoswaltCL:
looks good to me, other than whatever's in the merge conflicts

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -266,7 +294,75 @@ export class FeatureFlagService {
     featureFlag.id = flagDTO.id;
     featureFlag.key = flagDTO.key;
     featureFlag.status = flagDTO.status;
+    featureFlag.context = flagDTO.context;
+    featureFlag.tags = flagDTO.tags;
+    const newExclusion = new FeatureFlagSegmentExclusion();
+    const newInclusion = new FeatureFlagSegmentInclusion();
+    featureFlag.featureFlagSegmentExclusion = { ...flagDTO.featureFlagSegmentExclusion, ...newExclusion };
```
VivekFitkariwala, line 301: Should not this be 
`{ ...newExclusion, ...flagDTO.featureFlagSegmentExclusion }`?

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -110,36 +140,51 @@ export class FeatureFlagService {
...
 
-      const variationDocToReturn = variationDocs.map((variationDoc) => {
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        const { featureFlagId, ...rest } = variationDoc as any;
-        return rest;
-      });
-      return { ...featureFlagDoc, variations: variationDocToReturn as any };
+      const newFeatureFlagObject = {
+        ...featureFlagDoc,
+        featureFlagSegmentInclusion: { ...featureFlagSegmentInclusionDoc, segment: segmentIncludeDoc } as any,
```
VivekFitkariwala, line 183: This will depend on if `includeSegmentExists` and the same for `featureFlagSegmentExclusion`

bcb37: True. Good catch.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
bcb37: I'm not sure it matters because `newExclusion` isn't being instantiated with any attributes, but I'll reverse those spreads.

danoswaltCL:
hey it works. @VivekFitkariwala let us know if this can be merged, this will be helpful for getting the UI moving. I see some outstanding comments but I think Ben has responded.

VivekFitkariwala (approved):

PR #1481, opened 2024-04-29
add google client id in env files

Yagnik56:
This PR is created to add google client id in env files, to fix auth fail which is made mandatory in last audit log PR.
All devs have to add same change in there personal `.env` file to avoid auth fail.

### DIFF
--- a/backend/packages/Upgrade/.env.example
+++ b/backend/packages/Upgrade/.env.example
@@ -63,7 +63,7 @@ MONITOR_PASSWORD=1234
 #
 # Google Client Id
 #
-GOOGLE_CLIENT_ID=google_project_id
+GOOGLE_CLIENT_ID=135765367152-pq4jhd3gra10jda9l6bpnmu9gqt48tup.apps.googleusercontent.com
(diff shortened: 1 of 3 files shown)


danoswaltCL:
so... do we need this because the test suite depends on it? we shouldn't be hard-coding account-specific these things in here, I don't think this can be the solution.

backend/packages/Upgrade/.env.example
```diff
@@ -63,7 +63,7 @@ MONITOR_PASSWORD=1234
 #
 # Google Client Id
 #
-GOOGLE_CLIENT_ID=google_project_id
+GOOGLE_CLIENT_ID=135765367152-pq4jhd3gra10jda9l6bpnmu9gqt48tup.apps.googleusercontent.com
```
danoswaltCL, line 66: .env.example shouldn't need to be changed, it's just a placeholder

Yagnik56: I think we can close this PR as we have diff client ids anyway and just add this step in our setup doc.

PR #1487, opened 2024-05-01
Removed the terms and condition text on login page 

Yagnik56:
In this PR terms and condition text on login page below Google sign in button has been removed as it looked like clickable link but didn't redirect any page and we don't have any term & conditions of our own as we are using google sign in only.


danoswaltCL:
seems like the right thing to do to me, @amurphy-cl thoughts on just removing the link?

zackcl:
Yeah, I also think it's fine to remove it. Or maybe replace the text with something else like "To learn more about UpGrade, please visit www.upgradeplatform.org".

## Before:
Screenshot 2024-05-01 at 3 24 57 PM

## After:
Screenshot 2024-05-01 at 3 24 15 PM


## Demo App Login (For reference):
Screenshot 2024-05-01 at 3 27 18 PM

amurphy-cl:
Linking to the website seems fine.

VivekFitkariwala (approved):

Issue #1453, opened 2024-04-16
"Current Status" drop-down doesn't show the current value on the Change Experiment Status modal


zackcl:
**Version where bug was found:**
5.1.5

**Describe the bug**
The "Current Status" drop-down doesn't show the current value on the Change Experiment Status modal, except when the current status is set to "Inactive" or "Scheduled". When the current status is set to "Enrolling" or "Enrollment Complete", the "Current Status" drop-down shows the placeholder without any value.

**To Reproduce**
1. Go to an experiment with the "Enrolling" status.
2. Open the "Change Experiment Status" modal to change the status.
3. Observe that the "Current Status" drop-down shows the placeholder without any value.

**Expected behavior**
The "Current Status" drop-down on the "Change Experiment Status" modal should display the current status.

**Screenshots**
<img width="1000" alt="Screenshot 2024-04-16 at 2 53 05 PM" src="https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/fd633ee9-fc02-4bc6-acb2-dd06077a5649">

PR #1488, opened 2024-05-01
BugFix #1453. 

KD1712:
Current status value of experiment is now shown in change experiment status modal

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/experiment-status/experiment-status.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/experiment-status/experiment-status.component.html
@@ -13,9 +13,9 @@
         disabled="true"
         [(value)]="experimentInfo.state"
       >
-        <mat-option class="ft-14-400" *ngFor="let status of experimentStatus" [value]="status.value">
-          {{ status.value | experimentState }}
-        </mat-option>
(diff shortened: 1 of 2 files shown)


ppratikcr7 (approved):

frontend/projects/upgrade/src/app/features/dashboard/home/components/modal/experiment-status/experiment-status.component.html
```diff
@@ -13,9 +13,9 @@
         disabled="true"
         [(value)]="experimentInfo.state"
       >
-        <mat-option class="ft-14-400" *ngFor="let status of experimentStatus" [value]="status.value">
-          {{ status.value | experimentState }}
-        </mat-option>
+      <mat-option class="ft-14-400" *ngFor="let status of experimentStatus" [value]="experimentInfo.state">
```
zackcl, line 16: I think we can remove the `*ngFor="let status of experimentStatus"` because we only need to display one option.

zackcl (changes requested):
I think we can remove the `*ngFor="let status of experimentStatus"` because we only need to display one option.

danoswaltCL:
I don't believe `config.json` has any function, I actually believe it was supposed to have been deleted? backend/packages/Upgrade/src/config.json

Yagnik56:
@KD1712 can you remove the `config.json` change from PR?

PR #1490, opened 2024-05-01
Schedule date not visible in experiment details page 

Yagnik56:
This PR fixes the issue of the scheduled date not being visible on the experiment details page.


zackcl:
Review @zackcl

zackcl (approved):

zackcl:
The scheduled date is visible on the details page, but it requires refreshing or revisiting the page. I think this is OK for now.

Yagnik56:
I noticed that. There is an issue with API, it doesn't send a `startOn` when a status update response is sent so the store isn't updating without refresh. I will create a separate Issue for it as it will take some time to resolve.

PR #1353, opened 2024-03-08
Initial standalone components and new feature-flag Root Page

danoswaltCL:
This PR has changed to include the following:

1. Adds a `shared-standalone-component-lib` directory that is meant to mirror the `shared` directory.
2. Renames `feature/feature-flags` to `feature/feature-flags-legacy`.
3. Adds back `feature-flags` dir, now as the new feature flag feature dir
4. Adds a `component playground` app with references to the standalone library so elements can be quickly imported into there and developed / demo'd in isolation.
5. Adds a new [Common Root Page](#1360) presentational layout component.
6. Adds a new [Feature Flag Root Page](#1411) instance of that Common Root Page.
7. Points the feature-flags nav link to this component, which does lazy-loading with `loadComponent` rather than `loadChildren`.

This also contains draft code for Common Section Card List, Common Section Card List, and some demo Section Card usage in the Feature Flag Root Page to demonstrate the way this will work. I've left it in for the next person to pick up and complete.

PR #1491, opened 2024-05-01
Feature/feature flags root page

danoswaltCL:
This is a simplified PR without all the extras of #1353, which I moved into draft for reference.

- Renames all old feature flag files and references to `_LEGACY`
- `core/feature-flags` exists for later, but is just a placeholder and has no ngrx files in it yet.
- Adds Common Root Page
- Adds Common Root Page Header
- Adds Feature Flag Root Page and Header

Component Playground is removed.

Section Card stuff is removed.


frontend/projects/upgrade/src/app/features/dashboard/dashboard-routing.module.ts
```diff
@@ -33,13 +34,17 @@ const routes: Routes = [
...
-      //   path: 'featureFlags',
-      //   loadChildren: () => import('./feature-flags/feature-flags.module').then(m => m.FeatureFlagsModule),
-      //   data: {
-      //     title: 'app-header.title.feature-flag'
-      //   }
-      // },
+      // feature-flags is built with standalone components instead of an ngModule, so we need to lazy load the component directly
+      {
+        path: 'featureflags',
+        loadComponent: () =>
```
Yagnik56, line 40: since we are lazy loading components instead of the module, will this work fine for other components? particularly for loading feature flag root content or the feature flag details page.

Yagnik56 (approved):
@danoswaltCL Everything looks good to me. I just had that one question about lazy loading the component instead of a module won't be a problem for loading other child components.

PR #1502, opened 2024-05-06
Snackbar not visible in production

Yagnik56:
In this PR changes below have been made.
Issue: 
Snackbar was not visible in staging while working fine on the local dev environment.
Solution
- Removed the condition for not creating snackbar when `environment.production = true`.
- Allowed error notifications in production the same way we get in the Dev environment.
- Removed 401 (auth fail error) condition to avoid showing error with `environment.production = true`.
- Removed the test cases which were checking the above conditions.
- removed MatSnackbar module from ng module.


frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
```diff
@@ -17,25 +17,16 @@ export class HttpErrorInterceptor implements HttpInterceptor {
...
-    }
+    this._notifications.create(temp.title, temp.content, temp.type, temp);
   }
 
   intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
     return next.handle(request).pipe(
       catchError((err) => {
-        if (err.status === 401) {
-          // auto logout if 401 response returned from api
-          this.authService.authLogout();
```
VivekFitkariwala, line 37: Is this the code that logs out when the server sends a 401 error?

frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.spec.ts
```diff
@@ -77,25 +77,6 @@ describe('HttpErrorInterceptor', () => {
   });
 
   describe('#intercept', () => {
-    it('should logout if a 401 error is caught and open popup', fakeAsync(() => {
```
VivekFitkariwala, line 80: We should keep this testcase

frontend/projects/upgrade/src/app/core/error-handler/app-error-handler.service.spec.ts
```diff
@@ -15,7 +15,7 @@ describe('AppErrorHandler', () => {
     service = new AppErrorHandler(mockNotificationsService, mockEnvironment);
   });
 
-  it('should call notification service with an error of "An error occured. See console for details." when not in production and not 401', () => {
+  it('should call notification service with an error of "An error occurred. See console for details." when not in production and not 401', () => {
```
VivekFitkariwala, line 18: This test case needs to be modified removing the constraints of the environment. Also update the description of the test case

VivekFitkariwala (changes requested):

frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
Yagnik56: yes

frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.spec.ts
Yagnik56: but it will fail since we removed the code that logs out when a 401 error is received in service, do we want to keep that condition?

zackcl (approved):

zackcl (approved):

zackcl:
I could confirm that the snackbar is displayed properly on QA UpGrade when I test-deployed this branch.

VivekFitkariwala (approved):

PR #1503, opened 2024-05-07
Common component section card list

Yagnik56:
This PR contains common component section card list changes based on requirements mentioned in the issue.
CSS added in here will be fine to fit section-cards properly but I would like to hear what others have in mind.


zackcl:
`common-section-card-list` should be inside the `components` folder.

zackcl (approved):

PR #1525, opened 2024-05-10
Remove unused relation in Experiment Assignment Service

RidhamShah:
Improve Monitored-Decision-Point query by removing unnecessory relations.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -218,7 +218,7 @@ export class ExperimentAssignmentService {
       where: {
         site: site,
         target: target,
-        user: userId,
+        user: userDoc,
       },
     });
     if (experimentId && experiments.length) {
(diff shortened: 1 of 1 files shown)


RidhamShah:
@danoswaltCL @bcb37 you can also test this change and push it in release 5.1

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -220,7 +220,6 @@ export class ExperimentAssignmentService {
         target: target,
         user: userId,
```
VivekFitkariwala, line 221: The user here should be the original user instead of just the `userId`. Can you replace this code with
`user: userDoc`

VivekFitkariwala (changes requested):

danoswaltCL:
I tried load-testing this quickly a couple times and surprisingly didn't see improved performance. Still I think this only makes sense to merge anyway?

danoswaltCL (approved):

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -218,9 +218,8 @@ export class ExperimentAssignmentService {
       where: {
         site: site,
         target: target,
-        user: userId,
+        user: userDoc,
```
danoswaltCL, line 221: It seems like userId should be faster if TypeORM just needs the Primary Key, right? It may not really matter one way or another, just curious.

PR #1536, opened 2024-05-15
Replacing userId to userDoc for client APIs

RidhamShah:
In client APIs, instead of passing userId & userDoc from the Controller level to the Service Level, now we are passing only userDoc and userId is extracted from it.

This will reduce data redundancy and enhance code clarity.


backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -95,16 +95,17 @@ export class ExperimentUserService {
...
-    userId: string,
+    userDoc: RequestedExperimentUser,
     aliases: string[],
-    requestContext: { logger: UpgradeLogger; userDoc: any }
+    logger: UpgradeLogger
   ): Promise<IUserAliases> {
-    const { logger, userDoc } = requestContext;
+    const userId = userDoc.id;
     const userExist = userDoc;
+    console.log('userId', userId, 'userDoc', userDoc);
```
VivekFitkariwala, line 104: Remove console.log

VivekFitkariwala (dismissed):

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
RidhamShah: code updated

VivekFitkariwala (approved):

PR #1541, opened 2024-05-16
fetch feature flag paginated crud and store changes

Yagnik56:
This PR contains Crud operation and store-related changes for the fetch feature flag paginated which will be required for feature flag root page implementation.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.selectors.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.selectors.ts
@@ -1,0 +1,32 @@
+import { createSelector, createFeatureSelector } from '@ngrx/store';
+import { FeatureFlagState } from './feature-flags.model';
+import { selectRouterState } from '../../core.state';
+import { selectAll } from './feature-flags.reducer';
+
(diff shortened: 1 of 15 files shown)


frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.selectors.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const selectFeatureFlagsState = createFeatureSelector<FeatureFlagState>('featureFlags');
+
+export const selectAllFeatureFlags = createSelector(selectFeatureFlagsState, selectAll);
+
+export const selectIsLoadingFeatureFlags = createSelector(
+  selectFeatureFlagsState,
+  (state) => state.isLoadingFeatureFlags
+);
+
```
danoswaltCL, line 14: export const selectSelectedFeatureFlag = createSelector(
  selectRouterState,
  selectFeatureFlagsState,
  ({ state: { params } }, featureFlagState) => featureFlagState.entities[params.flagId]);

danoswaltCL: is exactly the same as what is there now, never a need a ternary for this kind of logic to return undefined

danoswaltCL (approved):

PR #1544, opened 2024-05-16
Feature/remove section card list

danoswaltCL:


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-page/common-page.component.html
```diff
@@ -2,7 +2,7 @@
   <header class="header">
     <ng-content select="[header]"></ng-content>
   </header>
-  <section class="content">
-    <ng-content select="[content]"></ng-content>
+  <section class="content-container">
+    <ng-content select="[section-card]"></ng-content>
   </section>
 </div>
```
zackcl, line 8: Shouldn't the content section be like this?
```
  <section class="content">
    <ng-content select="[content]"></ng-content>
  </section>
  ```

File: frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-page/common-page.component.ts [hunk 2]
```diff
@@ -5,15 +5,32 @@ import { TranslateModule } from '@ngx-translate/core';
...
  * Simply nest the component and add the slot name as an attribute.
  *
  * Example usage:
  *
  * ```
  * <app-common-page>
- *   <div header>Hi I'm in the header slot</div>
- *   <div content>Hi I'm in the content slot</div>
+ *   <div header>header</div>
+ *   <div section-card>section-card 1</div>
```
Yagnik56, line 16: Do you think this is the right way? I feel common page HTML shouldn't be affected by this change

Yagnik56 (commented):
except for changes in `common-page-component` mentioned by zack and me, everything looks good to me.

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-page/common-page.component.ts
danoswaltCL: there didn't seem to be anything that pulling out this structure into a special `section card list` would offer if this is the only kind of "content" that a page should ever expect. Later if we find that's not the case, where a page may not use section-cards, we can pretty easily adapt that and pull this structure into its own component. That was the reasoning.

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-page/common-page.component.html
danoswaltCL: I think class="content" is right for styling, but we want `section-card` here because it'll get used like this:

```
<section class="content>
  <app-common-section-card section-card>Hi I'm a section card 1</app-common-section-card>
  <app-common-section-card section-card>Hi I'm a section card 2</app-common-section-card>
  <app-common-section-card section-card>Hi I'm a section card 3</app-common-section-card>
</section>
```

not like this with `content` to select multiple cards

```
<section class="content>
  <app-common-section-card content>Hi I'm a section card 1</app-common-section-card>
  <app-common-section-card content>Hi I'm a section card 2</app-common-section-card>
  <app-common-section-card content>Hi I'm a section card 3</app-common-section-card>
</section>
```

danoswaltCL:
Pushed up commit removing the feature-flag-details-content-page and feature-flag-root-content-component since we decided that's an unnecessary layer. This removes a layer of nesting in the file structure as well, so we'll just have the header and section cards listed there as children instead of header and content.

@Yagnik56 please let us know if this change is okay with you.

danoswaltCL:
hmm rethinking this, getting rid of section-card-list is fine but maybe we do want feature-flag page-content-components. I think let's talk on monday @Yagnik56 i put this in draft for now

danoswaltCL:
I've flip-flopped back to wanting to keep it like this if we are deleting the section-card-list concept, so I re-opened the pr. I don't think an additional content-component makes sense if there's no content slot, we might as well keep the section-card-list concept if we wanted that. Because it would look like this, which is awkward to me because one element has a `header` attribute, and the other has nothing... 

Details page layout:
```html
<app-feature-flag-detail-page>
  <app-common-details-header header class="header"
    rootName="feature-flags"
    detailsName="Feature Flags"
    rootLink="featureflags"
  ></app-common-details-header>
  <app-feature-flags-detail-content></app-feature-flags-detail-content> <--- the html of `section-card` slots is hidden in this element.
</app-common-page>
```

app-feature-flags-detail-content
```html
<ng-container>
  <app-feature-flags-details-overview-section-card section-card (selectedIndexChange)="selectedIndex = $event">overview-card-with-tabs</app-feature-flags-details-overview-section-card>
    
  <!-- tabbable content below that-->
  <ng-container [ngSwitch]="selectedIndex">
    <ng-container *ngSwitchCase="0">
      <app-feature-flags-inclusions-section-card section-card>inclusions-card</app-feature-flags-inclusions-section-card section-card>
      <app-feature-flags-exclusions-section-card section-card>exclusions-card</app-feature-flags-exclusions-section-card section-card>
    </ng-container>
    <ng-container *ngSwitchCase="1">
      <app-feature-flags-exposures-section-card section-card>exposures-card</app-feature-flags-exposures-section-card section-card>
    </ng-container>
  </ng-container>
</ng-container>
```

Compare to the previous example where the entire page layout is readable with `header` and `section-card` elements. So I think we either merge this PR and go with that plan, or discard this and stick with section-card-list so we can have a content container. @Yagnik56 @zackcl

PR #1549, opened 2024-05-17
common section card search header component

Yagnik56:
This PR contains a common component of the section card search header widget.


danoswaltCL:
I think this looks great! 💯  I'm sure there might be some css adjustments once we see this in the section card, but this is good to go as far as I'm concerned.

danoswaltCL (approved):

PR #1554, opened 2024-05-20
Refactored ExperimentAliasesValidator. Updated the validator to check…

KD1712:
This PR ensures that the aliases property is validated as an array, and throws an error if the value is not an array.


backend/packages/Upgrade/src/api/controllers/validators/ExperimentUserAliasesValidator.ts
```diff
@@ -1,12 +1,13 @@
...
 
 export class ExperimentUserAliasesValidator {
   @IsNotEmpty()
   @IsDefined()
   public userId: string;
 
-  @IsNotEmpty()
+  @IsArray()
+  @IsString({ each: true })
   @Column('text', { array: true })
```
VivekFitkariwala, line 11: Remove @Column because not a validator

KD1712: removed column

ppratikcr7 (approved):

PR #1557, opened 2024-05-21
common section card title subtitle leftside header

Yagnik56:
This PR contains a section card header on the left side, and a title subtitle component.


zackcl:
I think the `chipText` should be optional (`chipText!`).

I don't think `link` and `linkText` inputs are needed as they are fixed to opening a "View Logs" modal when the "View Logs" link text is clicked.

Screenshot 2024-05-21 at 12 01 14 PM


How about we update the inputs like this?
```
  @Input() title!: string;
  @Input() chipText?: string;
  @Input() subtitle!: string;
  @Input() hasViewLogs?: boolean;
  ```
  
  And HTML would like something like this:
  ```
 <div class="title-header">
  <h1 class="title ft-26-700">
    {{ title | translate }}
    <ng-container *ngIf="chipText">
      <mat-chip [color]="getChipColor(chipText)" selected>
        {{ chipText }}
      </mat-chip>
    </ng-container>
  </h1>
  <h3 class="subtitle ft-16-400">
    {{ subtitle | translate }}
    <ng-container *ngIf="hasViewLogs">
      <a href="#" class="view-logs-link" (click)="openModal($event)">View logs</a>
    </ng-container>
  </h3>
</div>
```

zackcl:
OK, we discussed that we'll need a separate component for the Status Indicator and this component would probably expect an optional `status` with an Enum type.

Yagnik56:
@zackcl I think we shouldn't open the modal from the common component instead emit an event when `viewLogs` is clicked so it can be handled by the component which is calling that common component.

zackcl (approved):

PR #1563, opened 2024-05-21
Update Valid Condition Weight Checking

### LINKED ISSUE #1562: The "Next" button doesn't work in the experiment stepper when the first condition weight is set to zero
5.2.0

**Describe the bug**
The "Next" button doesn't work in the experiment stepper when the first condition weight is set to zero.

**To Reproduce**
1. Click "Add Experiment" to create an experiment.
2. In the "Design" step, add 2 or more conditions.
3. Update the first condition weight to zero.
4. Observe that you will not be able to proceed to the next step by clicking the "Next" button.

**Expected behavior**
The "Next" button should still work even if the first condition weight is set to zero.

zackcl:
This PR updates the condition weight checking so that the `experimentDesignForm` is `valid` even when the first condition weight is set to zero.

Closes #1562


Yagnik56 (approved):
looks goods

danoswaltCL (approved):

Issue #150, opened 2021-09-23
Importing the UpGrade client library (JS) without Webpack


zackcl:
As discussed in the meeting, I was able to import the UpGrade client library with the following steps:

1. Run `npm install upgrade_client_lib`.
2. Make the library file accessible from the client by adding `app.use(express.static(path.join(__dirname, "node_modules/upgrade_client_lib/dist")));` in the node express server.
3. Import the bundle.js file by adding `import "/bundle.js";` in the client JavaScript module.
4. Access the UpgradeClient class with `const UpgradeClient = window["upgrade-client-lib"].UpgradeClient;` in JavaScript.

And after learning that the library exists in [JSDelivr](https://www.jsdelivr.com/package/npm/upgrade_client_lib), I could use the library without having to npm install the upgrade_client_lib package, which is like the following:

1. Add `<script src="https://cdn.jsdelivr.net/npm/upgrade_client_lib@1.1.7/dist/bundle.min.js"></script>` into the HTML `<head>` tag.
2. Access the UpgradeClient class with `const UpgradeClient = window["upgrade-client-lib"].UpgradeClient;` in JavaScript.

While this works, accessing the UpgradeClient with `window["upgrade-client-lib"].UpgradeClient;` is not a clean way to write code.

So I think we should produce a separate JS file with Webpack that would allow users to directly access the `UpgradeClient` class by simply adding a script tag to HTML. I think the ideal process would be like the following:

1. Add `<script src="https://cdn.jsdelivr.net/npm/upgrade_client_lib@1.1.7/dist/upgrade_client_lib.min.js"></script>` into the HTML `<head>` tag.
2. In JavaScript, the user can directly construct the `UpgradeClient` class with `const upClient = new UpgradeClient(userId, hostUrl);`

I'm not familiar with producing a JS file with Webpack, so I would like to ask if what I described above is possible and if this makes sense. Let me know if anyone has any concerns or suggestions regarding this issue.

amurphy-cl:
@nirmalpatel Can you review Zack's suggestion here and share thoughts?

mswartzCL:
@zackcl Do you think this is something that we still need to do or shall we close it?

zackcl:
I think this should remain open, we would need to discuss this.

PR #1565, opened 2024-05-22
Refactor webpack configuration to improve library exports

KD1712:
Changed the webpack configuration to specify `library` as 'UpgradeClient' and `libraryExport` as 'default' to improve library exports. This ensures that users can directly access the `UpgradeClient` class without needing to access it through additional objects. Also updated the `globalObject` to 'this' for better compatibility across environments.


KD1712:
@nirmalpatel Here 'output.library' is removed from the webpack config to get direct access to the modules. Should we keep it this way for easier module access, or do we keep the library name for namespacing?

VivekFitkariwala:
@KD1712 can you check the PR? There are some extra change.

clientlibs/js/webpack.config.ts
```diff
@@ -34,8 +34,10 @@ const browser = {
   output: {
     filename: 'index.js',
     path: path.resolve(__dirname, 'dist/browser'),
+    libraryExport: 'default',
```
VivekFitkariwala, line 37: There are four variations where the same change needs to be done. Can you refactor the code and extract the common code?

VivekFitkariwala (changes requested):

VivekFitkariwala (dismissed):

danoswaltCL:
@zackcl can you verify this is the change that was needed in demo app for vanilla js when you get back?

danoswaltCL:
hey @zackcl this is related to a very old issue that you opened to help (#150), is this what you were looking for?

zackcl:
I tried the client library on this branch, and I was able to create the `UpgradeClient` class directly with `const upClient = new UpgradeClient(userId, hostUrl, context);`, which is what I expected.

However, while the `clientlibs/js/dist/browser/index.js` script works fine, the `clientlibs/js/dist/browser-lite/index.js` script does not.

When using the `browser-lite` version, I get the error: `Please provide a valid httpClient, or use the default version of the library with our default httpClient.`

You can see the code I wrote for testing the client library here: [https://jsfiddle.net/my93L5bg/](https://jsfiddle.net/my93L5bg/).

ppratikcr7:
> I tried the client library on this branch, and I was able to create the `UpgradeClient` class directly with `const upClient = new UpgradeClient(userId, hostUrl, context);`, which is what I expected.
> 
> However, while the `clientlibs/js/dist/browser/index.js` script works fine, the `clientlibs/js/dist/browser-lite/index.js` script does not.
> 
> When using the `browser-lite` version, I get the error: `Please provide a valid httpClient, or use the default version of the library with our default httpClient.`
> 
> You can see the code I wrote for testing the client library here: https://jsfiddle.net/my93L5bg/.

Yeah, I do get the same error. For `dist/browser` on `dev` branch build, I had to update the `apiVersion` to `v5` to make it work and get assignment. But doing the same change with `index.js` for `dist/browser-lite` still gives the above error.

ppratikcr7:
@zackcl I have updated the [fiddle](https://jsfiddle.net/tfLgphwa/29/), we need to provide the httpClient object as options in the UpgradeClient so that it returns the assignment for browser-lite.

zackcl:
@ppratikcr7 Thanks, so should we expect users to add this code when using "browser-lite"? I'm not sure if this is worth including in our documentation in case they use "browser-lite". If it's difficult to support "browser-lite", maybe we can just document that the lite version is not supported for JavaScript apps at the moment, and fix this later.

```
let httpClient = {
      doGet: (path, options) => {
        return fetch(path, options).
        then(response => response.json());
      },
      doPost: (path, body, options) => {
        body = JSON.stringify(body);
        return fetch(path, { method: "POST", body, ...options }).
        then(response => response.json());
      },
      doPatch: (path, body, options) => {
        body = JSON.stringify(body);
        return fetch(path, { method: "PATCH", body, ...options }).
        then(response => response.json());
      },
    };
    const options = { httpClient : httpClient }
    const upClient = new UpgradeClient(userId, hostUrl, context, options);
    ```

**Comment by zackcl** [comment]
I will review the updated fiddle soon.

**Comment by zackcl** [comment]
@ppratikcr7 Your fiddle worked fine with the custom HTTP client. I've just created a PR (#1862) suggesting the use of a default HTTP client for the lite version as well. Please feel free to check it out.

By the way, I had to change the version in `clientlibs/js/package.json` from "6.0.0" to "5.2.0" in the current `dev` branch; otherwise, I got an error. I'm not sure if this is because we don't have v6 endpoints yet.

**Review by zackcl — DISMISSED** [review]

File: clientlibs/js/webpack.config.ts [hunk 2]
```diff
@@ -1,6 +1,6 @@
-import path = require('path');
-import webpack = require('webpack');
-import packageJson = require('./package.json');
+const path = require('path');
+const webpack = require('webpack');
+const packageJson = require('./package.json');
```
zackcl, line 3: Could we change these back to` import`? The `eslint` complains "Require statement not part of import statement." if I use `const` instead.

danoswaltCL:
@ppratikcr7 can you take over this PR and look at @zackcl 's comment, then we can get this in for 6.0

ppratikcr7:
> @ppratikcr7 can you take over this PR and look at @zackcl 's comment, then we can get this in for 6.0

@zackcl  Can you  reapprove this and merge?

zackcl (approved):

PR #1572, opened 2024-05-23
Splitting updated export CSV data query for latency optimization and removing excluded user data

### LINKED ISSUE #1567: ✨ Export CSV Data Typeorm Query Optimization for latency improvement
e.g "5.2"

**Describe the bug**
We need to optimize the Typeorm Query for Export CSV Data for latency improvement. For this we need to do the following:

1. We can split the query into two queries. One would deal with MonitoredDecisionPoint & MonitoredDecisionPointLog tables
2. Then later iterate over this MonitoredDecisionPoint query and append the data for the enrolled user which is common for all the marks in order to prepare the CSV.
3. Remove exclused users from this data and remove exclusion code from the new export format.

ppratikcr7:
This PR resolves #1567 

Below are the optimizations done for the experiment export CSV data:
1. We have divided the single typorm query for the new export format into 3 parts. One would gather experiment details and would be kept outside the paginated query call for individualEnrollment table, so this experiment query joins would run once every 50 users paginated data call. The second query would have no joins with individualEnrollment table. The third query would have a single join for MDP(MonitoredDecisionPoint) table with MDPLog table. Later we stitch all these 3 query data together to export CSV.
2. Earlier in the single typeorm query we had 11 joins with the monitoredDecisionPoint table for the new export format, but we reduced that to zero joins with individualEnrollment table. The only join that was required was with UserStratificationFactor table for factorValue, but now we have optimized that by collecting user's list in batch and separately quering their factorValue and later appending it to the result. This will only be done in case of SRS experiment.


backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -422,96 +418,94 @@ export class AnalyticsRepository {
     skip: number,
     take: number
   ): Promise<CSVExportDataRow[]> {
+    const individualEnrollmentRepository = getCustomRepository(IndividualEnrollmentRepository, 'export');
```
VivekFitkariwala, line 421: @ppratikcr7 This query can further be simplified if we don't get the experiment detail in the query. You can query the experiment detail separately and use the paginated query to do the groupBy part and populate the details on the server side.

VivekFitkariwala (changes requested):

ppratikcr7:
Please find the updated export CSV data file using this PR:

Random Algorithm Experiment CSV data:
[Exp-Non-SRS.csv](https://github.com/user-attachments/files/15545657/Exp-Non-SRS.csv)

SRS Algorithm Experiment CSV data:
[Exp-SRS.csv](https://github.com/user-attachments/files/15545659/Exp-SRS.csv)

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
   }
 
   public async getCSVDataForSimpleExport(
+    experimentsData: any,
```
VivekFitkariwala, line 414: User proper typing here

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
...
     experimentId: string,
+    isSRSExperiment: boolean,
```
VivekFitkariwala, line 416: This can be determined from the experiment data itself. You don't need this parameter.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
...
+      .select([
+        '"individualEnrollment"."userId" as "userId"',
+        '"individualEnrollment"."groupId" as "enrollmentGroupId"',
+        '"individualEnrollment"."enrollmentCode" as "enrollmentCode"',
+        '"individualEnrollment"."experimentId" as "experimentId"',
+        '"individualEnrollment"."conditionId" as "conditionId"',
+        '"individualEnrollment"."partitionId" as "partitionId"',
+      ]);
+
+    individualEnrollmentQuery.groupBy('individualEnrollment.userId')
```
VivekFitkariwala, line 433: Why `individualEnrollmentQuery.groupBy`? Just use `.groupBy`

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
...
+      .addGroupBy('individualEnrollment.conditionId')
+      .addGroupBy('individualEnrollment.partitionId')
+      .orderBy('individualEnrollment.userId', 'ASC')
+      .offset(skip)
+      .limit(take)
+      .where('individualEnrollment.experimentId = :experimentId', { experimentId });
+    
     const monitoredDecisionPointRepository = getCustomRepository(MonitoredDecisionPointRepository, 'export');
-    return monitoredDecisionPointRepository
+    const monitoredDecisionPointQuery = monitoredDecisionPointRepository
```
VivekFitkariwala, line 445: This query needs to be run for users which you are getting from the above query otherwise, it will run the join for all the Monitored experiment tables which is not what we want.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
...
+    const [individualEnrollmentQueryResults, monitoredDecisionPointQueryResults] = await Promise.all([
+      individualEnrollmentQuery.getRawMany(),
+      monitoredDecisionPointQuery.getRawMany(),
+    ]);
+
+    const userStratificationFactorUserList = [];
+
+    const individualEnrollmentExperimentData = [];
+    individualEnrollmentQueryResults.forEach(individualEnrollmentQueryResult => {
+      experimentsData.forEach((experiment) => {
```
VivekFitkariwala, line 468: Why this `experimentData` is an array? This should be a single experiment

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -188,7 +193,51 @@ export class AnalyticsService {
         if (experiment.assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS) {
           csvExportData = await this.analyticsRepository.getCSVDataForWithInSubExport(experimentId, skip, take);
         } else {
-          csvExportData = await this.analyticsRepository.getCSVDataForSimpleExport(experimentId, skip, take);
+          // Get the experiment details
+          const experimentRepository = getCustomRepository(ExperimentRepository, 'export');
+          const experimentQuery = experimentRepository
```
VivekFitkariwala, line 198: This is a copy of the function in `ExperimentService` -> `findOne`. Can you refactor this code and move it to the Experiment Repository?

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -418,100 +411,110 @@ export class AnalyticsRepository {
...
+          'userStratificationFactor.stratificationFactorValue as "stratificationFactorValue"'
+        ])
+        .where('userStratificationFactor.user IN (:...userIds)', { userIds: userStratificationFactorUserList.map((data) => data) });
+      
+      userStratificationFactorQueryResult = await userStratificationFactorQuery.getRawMany();
+    }
+    // Combined results
+    const combinedCSVExportData = [];
+
+    individualEnrollmentExperimentData.forEach(individualEnrollmentExperiment => {
```
VivekFitkariwala, line 504: It would be best if you used `lodash` groupBy or something similar on the `monitoredDecisionPointQuery` result to groupBy userId to avoid nested loops.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
ppratikcr7: I am preparing an array of experimentData to have information about each condition and decisionpoint, so that I can get the conditionName and conditionPayload from this array. Because from the individualEnrollment table query we just get the conditionId assigned for the user, again we either need to perform a join with this table or have a separate api call to get the condition and payload data. We get the site and target information from the monitoredDP query. Below is the array for 1 DP and 1 conditions to get an idea:

experimentsData:  [
  {
    experimentId: 'be3ae74f-370a-4015-93f3-7761d16f8b17',
    experimentName: 'Test Experiment',
    context: [ 'home' ],
    assignmentUnit: 'group',
    group: 'teacher',
    consistencyRule: 'group',
    designType: 'Simple',
    algorithmType: 'random',
    stratification: null,
    postRule: 'continue',
    revertTo: null,
    enrollmentStartDate: 2024-06-03T12:05:27.132Z,
    enrollmentCompleteDate: null,
    payload:  'payload1',
    excludeIfReached: true,
    expDecisionPointId: '2dfa7f7f-47c6-4b81-bf49-d0f41e718687',
    expConditionId: 'c22467b1-f0e9-4444-9517-cc03037bc079',
    conditionName: 'ConditionA'
  },
  {
    experimentId: 'be3ae74f-370a-4015-93f3-7761d16f8b17',
    experimentName: 'Test Experiment',
    context: [ 'home' ],
    assignmentUnit: 'group',
    group: 'teacher',
    consistencyRule: 'group',
    designType: 'Simple',
    algorithmType: 'random',
    stratification: null,
    postRule: 'continue',
    revertTo: null,
    enrollmentStartDate: 2024-06-03T12:05:27.132Z,
    enrollmentCompleteDate: null,
    payload: 'payload2',
    excludeIfReached: true,
    expDecisionPointId: '42ffaf10-6da1-4dc4-a6fe-6b54bb8cc95e',
    expConditionId: 'd22467b1-f0e9-4444-9517-cc03037bc079',
    conditionName: 'ConditionB'
  }
  ]

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
ppratikcr7: This is subset of the `findOne`. In `findOne` function we have a lot of extra joins and selections which we dont require here. If you say, I can move this query in a separate function `findExperimentForCSVExport` in the `ExperimentService ` and call it from there.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
ppratikcr7: Earlier did that, but thought of providing a separate parameter, else would have to get from the first array element of experimentData. Ok, will remove this.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
VivekFitkariwala: Please include the types.

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
VivekFitkariwala: Yup move it to the repository and reuse the common code.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -188,7 +193,51 @@ export class AnalyticsService {
...
+            .createQueryBuilder('experiment')
+            .select([
```
bcb37, line 200: This same query will now potentially be run multiple times to get the same data, since it is inside the 'do...while' loop. However, we're considering getting rid of that loop altogether anyway. But if we keep the query, it must be moved outside the loop.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -485,22 +456,60 @@ export class AnalyticsRepository {
       .limit(take)
       .where('monitoredDecisionPoint.experimentId = :experimentId', { experimentId });
     
-    const [individualEnrollmentQueryResult, monitoredDecisionPointQueryResult] = await Promise.all([
+    const [individualEnrollmentQueryResults, monitoredDecisionPointQueryResults] = await Promise.all([
```
bcb37, line 459: I'm not sure we can be guaranteed that the data from these two queries will match up for every iteration of the loop. If there are multiple monitored decision point entries for a given experiment and student, and only one enrollment, for instance, we could read 50 records from `montiored_decision_point` before we get the the entries for that student, even though we'll have the entry in `individual_enrollment`. That said, we're considering removing the loop anyway.

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
ppratikcr7: Yeah, it was a mistake. I had made the change for moving it outside the do while loop and was waiting for other changes and have now pushed that in my last commit. But, yeah you told we want to remove the loop for skip take.

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
ppratikcr7: Removed skip take from `montiored_decision_point` query in the last commit so that we guarantee that we take all MDPlogs for the users from `individual_enrollment` query. If you want, you can carry forward from here to remove skip take throughout.

bcb37 (dismissed):
@ppratikcr7   I've added commits to remove the skip/takes from the export queries and indexes on the 'monitored' tables. 
We want to add the non-merge commits from this pr to the release/5.1 branch for next week.
Please cherry-pick them and create a pr based on that branch.

Actually, that's not necessary. We'll remove the pagination from the 5.1 branch for release, that we can reconcile that with these optimizations for the next release.

backend/packages/Upgrade/src/api/repositories/ExperimentRepository.ts
```diff
@@ -427,4 +432,84 @@ export class ExperimentRepository extends Repository<Experiment> {
       throw error;
     }
   }
+
+  public createBaseQueryBuilder() {
```
VivekFitkariwala, line 436: Should be private

VivekFitkariwala (approved):

VivekFitkariwala (approved):

PR #1581, opened 2024-05-24
common status indicator chip component

Yagnik56:
This PR contains a common status indicator chip component.

### DIFF
--- a/frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-status-indicator-chip/common-status-indicator-chip.component.html
+++ b/frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-status-indicator-chip/common-status-indicator-chip.component.html
@@ -1,3 +1,3 @@
 <mat-chip class="status-indicator" [ngClass]="chipClass">
-  {{ convertKebabToTitleCase(chipClass) }}
+  {{ chipText }}
 </mat-chip>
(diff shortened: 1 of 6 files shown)


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-status-indicator-chip/common-status-indicator-chip.component.html
```diff
@@ -0,0 +1,3 @@
+<mat-chip class="status-indicator" [ngClass]="chipClass">
+  {{ convertKebabToTitleCase(chipClass) }}
```
danoswaltCL, line 2: functions in templates are generally not best practice because they will run each time change-detection happens. that really doesn't matter here, but still would be best to just run this function in the constructor or ngOnInit instead of in the template.

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-status-indicator-chip/common-status-indicator-chip.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  selector: 'app-common-status-indicator-chip',
+  templateUrl: './common-status-indicator-chip.component.html',
+  styleUrls: ['./common-status-indicator-chip.component.scss'],
+  imports: [CommonModule, TranslateModule, MatChipsModule],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CommonStatusIndicatorChipComponent {
+  @Input() chipClass!: STATUS_INDICATOR_CHIP_TYPE;
+
+  convertKebabToTitleCase(kebabCaseStr: STATUS_INDICATOR_CHIP_TYPE): string {
```
danoswaltCL, line 28: run this in the constructor (see other comment)

danoswaltCL (changes requested):

danoswaltCL (approved):

PR #1585, opened 2024-05-27
ClientLib changes for featureFlags to getKeys

RidhamShah:
In this Pull Request, the ClientLib `getFeatureFlags` method invokes `getAllFeatureFlags`, which returns an array of feature flags associated with the key user. The `getFeatureFlags` method then returns true or false based on whether the specified feature-flag key is present in the array.


clientlibs/js/src/ApiService/ApiService.ts
```diff
@@ -302,23 +302,18 @@ export default class ApiService {
     });
   }
 
-  public async getAllFeatureFlags(): Promise<IFeatureFlag[]> {
-    const response = await this.sendRequest<IFeatureFlag[], never>({
+  public async getAllFeatureFlags(): Promise<string[]> {
+    const requestBody: UpGradeClientRequests.IGetAllExperimentConditionsRequestBody = {
```
VivekFitkariwala, line 306: Should have `IGetAllFeatureFlagRequestBody` which internally could be same

VivekFitkariwala (approved):

clientlibs/js/src/ApiService/ApiService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
```diff
@@ -846,9 +862,14 @@ export class ExperimentClientController {
...
-  public getAllFlags(@Req() request: AppRequest): Promise<FeatureFlag[]> {
-    return this.featureFlagService.find(request.logger);
+  @Post('featureflag')
+  public async getAllFlags(
+    @Req() request: AppRequest,
+    @Body({ validate: true })
+    experiment: ExperimentAssignmentValidator
+  ): Promise<string[]> {
+    const experimentUserDoc = await this.experimentUserService.getUserDoc(experiment.userId, request.logger);
+    return this.featureFlagService.getKeys(experimentUserDoc, experiment.context, request.logger);
```
bcb37, line 863: We should return 404 when the user is not found, like with experiment assignment (and the removed 'keys' endpoint).

danoswaltCL (approved):
will spawn task to keep the discussion going on Ben's comment

Issue #1105, opened 2023-11-08
Create env variable to toggle error logs feature in the UI


danoswaltCL:
#1080 We disabled error-logging in UI due to it being a huge query to run, too much data in the table.

This was disabled via commenting out but in the future it would be easier to toggle on/off via environment variable.

amurphy-cl:
Need to have a larger discussion on how to make this tab and its messages useful to the front-end user (or whoever might be reading these logs).

danoswaltCL:
https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1712780394950259

PR #1595, opened 2024-05-30
added env variable to toggle for errorLogs feature

JayMehta11:
#1105 

Created env variable to toggle error logs feature in the UI.

For testing you can change errorLogsToggle in environment.ts files to true and see error logs in your UI.


danoswaltCL (approved):
looks good 💯

PR #1599, opened 2024-06-03
Common Details Overview Section Card Content

KD1712:
Created a reusable component for content in common section card


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.ts
```diff
@@ -0,0 +1,44 @@
...
+//   { description: 'something' },
+//   { tags: ['Tag1', 'Tag2'] },
+//   { Appcontext: 'Context1' },
+// ];
+//
+// Simply pass the data to the component as shown below:
+// <app-common-section-card-content [data]="contentDetails"></app-common-section-card-content>
+export class CommonSectionCardContentComponent implements OnChanges {
+  @Input() data: { key: string; value: string }[] = [];
+  keys: string[] = [];
```
danoswaltCL, line 31: `ngOnChanges` isn't going to be a thing we'd normally use directly in angular unless we're doing something really tricky, this shouldn't be necessary.

I don't think we'll want to be doing any text manipulation here either, this component should just reflect the values that are input. The parent component will be in charge sending in the text in the right display format.

I think we just want to remove lines 31-43 and that'll be all that's really needed.

danoswaltCL (changes requested):

danoswaltCL (approved):

PR #1601, opened 2024-06-03
feature flag details page overview header left code changes

Yagnik56:
This PR contains changes related to the details page overview header left integration.
Screenshot from 2024-06-03 19-41-25


frontend/projects/upgrade/src/app/core/feature-flags/feature-flags.service.ts
```diff
@@ -20,6 +20,27 @@ export class FeatureFlagsService {
   isAllFlagsFetched$ = this.store$.pipe(select(selectIsAllFlagsFetched));
   activeDetailsTabIndex$ = this.store$.pipe(select(selectActiveDetailsTabIndex));
 
+  formatDateString(dateString, dateTime) {
```
danoswaltCL, line 23: we have `dayjs` for date formatting:
`const date = dayjs("2024-05-31T21:54:19.692Z");
date.format('MMM DD, YYYY, hh:mm A');`

But we also already have a `FormatDate` pipe. We should probably just use that to be consistent with how this and other dates are done. I wouldn't worry too much about getting it exactly like in the design, but if you want to create an additional format, we should use the pipe to do that as well.

```
<span class="ft-14-700 flag-create-date">
        {{ ('home.view-experiment.created-on.text' | translate) + (flag.createdAt | formatDate: 'medium date') }}
      </span>
```

danoswaltCL (dismissed):
comment on the date format function is only thing, otherwise looks good

frontend/projects/upgrade/src/app/core/feature-flags/feature-flags.service.ts
Yagnik56: addressed this cmt

Yagnik56:
> Comment on the date format function is the only thing, it looks good

I made the change but it's not the same as shown in the design as it has only 2 options, we can add more options in the pipe later to provide more variety as per the needs.

frontend/projects/upgrade/src/app/core/feature-flags/feature-flags.service.ts
danoswaltCL: cool, this function can be deleted then.

Yagnik56: done

danoswaltCL (approved):

PR #1621, opened 2024-06-05
Feature/1094 stratification UI multiple csv uploads

### LINKED ISSUE #1094: Support multiple stratification CSV uploads
Currently, we only support uploading one stratification CSV file at a time, but ideally we want to support uploading multiple CSV files at once for ease of use.

In the Participants -> Stratifications tab, the user should be able to select and import multiple CSV files at once to the Stratifications table.

**Additional context**
* An example showing how it should work:

https://github.com/CarnegieLearningWeb/UpGrade/assets/90279765/8ce0933e-c008-4953-8750-ee25b7859f18

* The previous PR (closed due to needing more time to review and test)
https://github.com/CarnegieLearningWeb/UpGrade/pull/1088

* The `multer` library should be removed since the files are not being sent as actual file uploads. Instead, they're sent as plain text within the request body.

JayMehta11:
This PR resolves #1094 

It allows multiple CSV uploads in stratifiacation-UI and removed multer in the code and updated the testcases.


ppratikcr7 (changes requested):
You need to uninstall the multer library and that change in `package.json` and `package.lock.json` should be present in this PR

PR #1627, opened 2024-06-06
Bugfix/re throw error in http intercepter

danoswaltCL:
as discussed in some mtg, we should allow this error keep moving through so the error can be caught in angular components


frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
```diff
@@ -32,7 +32,9 @@ export class HttpErrorInterceptor implements HttpInterceptor {
           this.authService.authLogout();
         }
         this.openPopup(err);
-        return EMPTY; // returning EMPTY instead of throwError as Error is handled using snacker here itself
+
+        // re-throw to allow the error to be caught by the calling code
+        return throwError(() => new Error(JSON.stringify(err)));
```
VivekFitkariwala, line 37: Why not just `throw error`?

VivekFitkariwala (changes requested):

frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
danoswaltCL: I'm sorry, what is the code suggestion to try?

danoswaltCL: ok I see just `throw err` does work, let me fix the tests.

danoswaltCL: fixed @VivekFitkariwala

danoswaltCL:
@VivekFitkariwala is this looking right to you now? I'll need this fixed for the demo.

VivekFitkariwala (approved):

VivekFitkariwala:
@danoswaltCL yup approved

PR #1628, opened 2024-06-06
add both include and exclude headers with common header tableRowCount…

danoswaltCL:
Added the header-left components for both inclusions and exclusions, common section card title header modified to add a counter to the title

Note* this is mocking the count, this will need to get the length of table data in an observable stream when that is hooked up

image


Yagnik56 (approved):
Looks good to me,

PR #1629, opened 2024-06-06
Content for FF deatils page overview section card

Yagnik56:
This PR contains the content part for the feature flag details page overview card and rewriting the common section card content component to make it more usable and added proper styling to it. 
image


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.html
```diff
@@ -1,14 +1,14 @@
 <div class="content-section">
-  <div *ngFor="let obj of data">
-    <div *ngFor="let item of obj | keyvalue">
-      <span style="font-weight: bold">{{ item.key }}:</span>
-      <ng-container *ngIf="item.key !== 'tags'; else chipSection">
-        <span>{{ item.value }}</span>
+  <div *ngFor="let key of objectKeys(data)">
```
danoswaltCL, line 2: Please use `keyvalue` pipe, then we don't need extra `objectKeys` function.

<div *ngFor="let item of data | keyvalue">
  <div class='item-key-value'>
    <span class="ft-16-600 item-key">{{ item.key }}:</span>
    <ng-container *ngIf="item.key !== 'Tags'; else chipSection">
      <span class="ft-16-500 item-value">{{ item.value }}</span>
    </ng-container>

    <ng-template #chipSection>
      <mat-chip-list>
        <mat-chip *ngFor="let tag of item.value" class="chip">{{ tag }}</mat-chip>
      </mat-chip-list>
    </ng-template>
  </div>
</div>

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.ts
```diff
@@ -17,15 +17,19 @@ import { SharedModule } from '../../../shared/shared.module';
...
+//   Description: 'something',
+//   Tags: ['Tag1', 'Tag2'],
+//   app_Context: 'Context1',
+// };
 //
 // Simply pass the data to the component as shown below:
 // <app-common-section-card-content [data]="contentDetails"></app-common-section-card-content>
 export class CommonSectionCardContentComponent {
-  @Input() data: { key: string; value: string }[] = [];
+  @Input() data!: KeyValueFormat;
```
danoswaltCL, line 30: I think we want something more specific than a general "KeyValueFormat", because it's important that Tags is an array.

Can we try this as an interface that is specific to these overview components?

```
export interface OverviewDetails {
  [key: string]: string;
  ['Tags']?: string[];
}
```

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.ts
```diff
@@ -17,15 +17,19 @@ import { SharedModule } from '../../../shared/shared.module';
...
+// contentDetails = {
+//   Key: 'name',
+//   Description: 'something',
+//   Tags: ['Tag1', 'Tag2'],
+//   app_Context: 'Context1',
+// };
 //
 // Simply pass the data to the component as shown below:
 // <app-common-section-card-content [data]="contentDetails"></app-common-section-card-content>
 export class CommonSectionCardContentComponent {
```
danoswaltCL, line 29: I just noticed this, I think we want to call this component a `CommonSectionCardOverviewDetailsComponent` instead of `CommonSectionCardContentComponent`, this is specific to overview details, there will be several different components that can be "Section Card Content".

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.ts
Yagnik56: Giving this error Property '['Tags']' of type 'string[]' is not assignable to 'string' index type 'string'.
can't we keep it this way?
```
export interface KeyValueFormat {
  [key: string]: string | string[];
}
```

danoswaltCL: please try this instead. We're describing a specific type for the needs of this specific component only, not a general format that needs to be on the `types` file.

export type OverviewDetails = {
  [key: string]: string;
} & {
  Tags?: string[];
};

Yagnik56: this is done, little diff but done

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-content/common-section-card-content.component.html
Yagnik56: can you provide more info here? didn't get it what's this is for `{{ item.key }}: {{ item.value }}`

danoswaltCL: weird, I edited it, I must have pasted it into the formatting in a bizarre way

Yagnik56: ok, that worked but it is showing all keys in an alphabetically sorted manner, not as given in the object according to UI. Is that fine?

danoswaltCL: ah okay, let me reverse then, because maybe then we do want an array. So I think it made sense for Khanjan to use an array for the Input, but we can just leave it with object.keys if it's working, I don't want to spend more time on it

Yagnik56: I found a solution if I do it like this
`<div *ngFor="let item of data | keyvalue: noSort">`
where noSort is
`noSort() {
    return 0;
  }`
  then it works fine but can pass only func there can't direct put 0 next to keyValue

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-section-card-overview-details/common-section-card-overview-details.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="content-section">
+  <div *ngFor="let item of data | keyvalue: noSort">
```
danoswaltCL, line 2: :100 noSort. cool, didn't know that was a thing.

danoswaltCL (approved):

PR #1634, opened 2024-06-07
Get flag details on details page 

KD1712:
Get details on navigation. Issue to load details on page reload or url hit

### DIFF
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
@@ -2,8 +2,8 @@ import { FeatureFlagsDataService } from '../feature-flags.data.service';
 import { Actions, createEffect, ofType } from '@ngrx/effects';
 import { Injectable } from '@angular/core';
 import * as FeatureFlagsActions from './feature-flags.actions';
-import { catchError, switchMap, map, filter, withLatestFrom, tap, first } from 'rxjs/operators';
-import { FeatureFlagsPaginationParams, NUMBER_OF_FLAGS } from './feature-flags.model';
+import { catchError, switchMap, map, filter, withLatestFrom, tap, first, mergeMap } from 'rxjs/operators';
+import { FeatureFlag, FeatureFlagsPaginationParams, NUMBER_OF_FLAGS } from './feature-flags.model';
 import { Router } from '@angular/router';
 import { Store, select } from '@ngrx/store';
 import { AppState } from '../../core.module';
@@ -15,6 +15,7 @@ import {
   selectSortAs,
   selectSearchString,
 } from './feature-flags.selectors';
+import { of } from 'rxjs';
 
 @Injectable()
 export class FeatureFlagsEffects {
@@ -99,7 +100,6 @@ export class FeatureFlagsEffects {
     )
   );
 
-
   fetchFeatureFlagsOnSearchString$ = createEffect(
     () =>
       this.actions$.pipe(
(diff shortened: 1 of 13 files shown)


danoswaltCL:
So I'm guessing the issue you're seeing is this: what's happening is that the data you see when you nav from root page is from the pre-existing flag data, not actually from the fetch-by-id response. It's a trick! Your set-up is working up to that point. You can spy on the `featureFlags.entities` data-store value in the dev tools, when you navigate from root, the selector takes the id from url and finds it in there already because it exists from in the paginated call. When you refresh, that call is absent, so it will be empty, and it won't work that way because you're missing this:

Please add this in the reducer to pop the flag into the data store on success:

```
on(FeatureFlagsActions.actionFetchFeatureFlagByIdSuccess, (state, { flag }) => {
    return adapter.addOne(flag, {
      ...state,
      isLoadingFeatureFlags: false,
    });
  }),
```

2. If you put that in there, couple other things: 
- you'll want to make sure to wait until "data" is defined because this is an async call and it can be undefined at first
`<app-common-section-card-list *ngIf="featureFlag">`. It would be good to show a load-spinner if else.

- I found that I needed to help angular realize there was a change with `this.changeDetectorRef.detectChanges();` after setting this.featureFlag 

Now when the data store is updated on success, the page will render it when ready.

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-details-page/feature-flag-details-page-content/feature-flag-details-page-content.component.html
```diff
@@ -1,12 +1,18 @@
...
-    >inclusions-card</app-feature-flag-inclusions-section-card
-  >
-  <app-feature-flag-exclusions-section-card *ngIf="(activeTabIndex$ | async) === 0" section-card
-    >exclusions-card</app-feature-flag-exclusions-section-card
-  >
-  <app-feature-flag-exposures-section-card *ngIf="(activeTabIndex$ | async) === 1" section-card
-    >exposures-card</app-feature-flag-exposures-section-card
-  >
-</app-common-section-card-list>
+<div class="content-container">
```
danoswaltCL, line 1: if we load the sub to the feature flag directly with an async pipe, we can do this and eliminate half of the component code: 

```
<ng-container *ngIf="(featureFlag$ | async) as featureFlag; else loading">
  <app-common-section-card-list class="common-section-card-list">
    <app-feature-flag-overview-details-section-card [data]="featureFlag" section-card></app-feature-flag-overview-details-section-card>
    <app-feature-flag-inclusions-section-card [data]="featureFlag" *ngIf="(activeTabIndex$ | async) === 0" section-card>inclusions-card</app-feature-flag-inclusions-section-card>
    <app-feature-flag-exclusions-section-card [data]="featureFlag" *ngIf="(activeTabIndex$ | async) === 0" section-card>exclusions-card</app-feature-flag-exclusions-section-card>
    <app-feature-flag-exposures-section-card [data]="featureFlag" *ngIf="(activeTabIndex$ | async) === 1" section-card>exposures-card</app-feature-flag-exposures-section-card>
  </app-common-section-card-list>
</ng-container>

<ng-template #loading>
  <mat-spinner></mat-spinner>
</ng-template>
```

danoswaltCL: then in component code, you just need this:

```typescript
featureFlag$ = this.featureFlagsService.selectedFeatureFlag$
```

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
```diff
@@ -15,6 +15,7 @@ import {
   selectSortAs,
   selectSearchString,
 } from './feature-flags.selectors';
+import { of } from 'rxjs';
```
danoswaltCL, line 18: i think this is unused?

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-details-page/feature-flag-details-page-content/feature-flag-details-page-content.component.html
danoswaltCL: yeah i'll need this as an observable like suggested above in order to complete some work I'm currently doing, let me know if this suggestion works for you, I tried it in my local and it works fine.

KD1712: Yes works for me too

danoswaltCL (approved):

PR #1647, opened 2024-06-12
fetch context metadata when add feature flag modal load

Yagnik56:
This PR fixes the issue of app context metadata staying empty in the add feature flag modal by fetching context metadata when the component is initialized.


frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/add-feature-flag-modal/add-feature-flag-modal.component.ts
```diff
@@ -63,11 +64,13 @@ export class AddFeatureFlagModalComponent {
     public dialog: MatDialog,
     private formBuilder: FormBuilder,
     private featureFlagsService: FeatureFlagsService,
+    private experimentService: ExperimentService,
     private formHelpersService: CommonFormHelpersService,
     public dialogRef: MatDialogRef<AddFeatureFlagModalComponent>
   ) {}
 
   ngOnInit(): void {
+    this.experimentService.fetchContextMetaData();
```
danoswaltCL, line 73: yeah I created an issue for the same, https://github.com/orgs/CarnegieLearningWeb/projects/1/views/12?pane=issue&itemId=67041859. I know we can just do this and it will work for now but architecturally it sort of doesn't make sense for "contextmetadata" to still be tied to the experiment service. Plus this cal is a little redundant if we already have that data.

But we can talk about it in meeting, I think we can merge this since we need it and keep my ticket open in case we feel like refactoring a little later.

danoswaltCL (approved):

PR #1652, opened 2024-06-14
Changed response to only the ff object

KD1712:
Also changed its test case for the same


danoswaltCL:
@bcb37 can you take a peak at this one, looks fine to me but i didn't work on this

danoswaltCL:
@KD1712 @bcb37 I'm going to add a commit to this to correct the frontend at the same time so it doesn't break.

danoswaltCL (approved):

PR #1669, opened 2024-06-18
remove pagination from export queries

bcb37:


backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -179,164 +179,146 @@ export class AnalyticsService {
...
-      let skip = 0;
-      const take = 50;
-      do {
-        let csvExportData: CSVExportDataRow[];
-        if (experiment.assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS) {
-          csvExportData = await this.analyticsRepository.getCSVDataForWithInSubExport(experimentId, skip, take);
-        } else {
-          csvExportData = await this.analyticsRepository.getCSVDataForSimpleExport(experimentId, skip, take);
-        }
-        const userIds = csvExportData.map(({ userId }) => userId);
```
ppratikcr7, line 193: why is this part of code removed? We will still execute further code if no data.

bcb37: This part was removed because 'break' doesn't make sense outside of a loop. 
If the `experimentId` parameter is undefined then we probably should return immediately.
If there's an experiment with no enrolled students we'll return then

ppratikcr7 (approved):

PR #1691, opened 2024-06-25
Updated auto refresh while importing or deleting stratification files

### LINKED ISSUE #1689: Stratifications Page Auto Refresh Not working as expected
e.g "5.2"

**Describe the bug**
After deleting strata, it doesn't remove the entry without refreshing the page or moving to another page. Also, Multiple imports also show 1st one at first, then on refresh we see others.

**Expected behavior**
The delete/import of stratas should reflect instantly on the UI.

JayMehta11:
This PR resolves #1689 .

There was issue in Stratifications Page that sometimes auto refresh was not working as expected.


ppratikcr7 (changes requested):
- Disable import button for stratification modal when no file is selected.
- Increase size of table to view more rows.

Yagnik56 (commented):
tested and working fine

backend/packages/Upgrade/src/api/services/StratificationService.ts
```diff
@@ -220,4 +224,18 @@ export class StratificationService {
 
     return createdStratificationData;
   }
+
+  public async insertStratificationFiles(
+    files: UploadedFilesValidator[],
+    logger: UpgradeLogger
+  ): Promise<UserStratificationFactor[]> {
+    const stratificationResults: UserStratificationFactor[] = [];
```
VivekFitkariwala, line 232: This could be refactored with Array.map
return files.map((file) => this.insertStratification(fileObj.file, logger))

ppratikcr7: I can handle this ticket now.

VivekFitkariwala (approved):

ppratikcr7 (approved):

PR #1606, opened 2024-06-04
Typeorm updation

RidhamShah:
This PR covers below changes:

1. Update typeORM and all related depreciated code to new code.
2. Now, the export replica hostname would also be an array as a separate env variable `TYPEORM_HOSTNAME_EXPORT_REPLICA` and rest replicas (if any) would be in the `TYPEORM_HOSTNAME_REPLICAS`. For now you need to move the value in `TYPEORM_HOSTNAME_REPLICAS` to `TYPEORM_HOSTNAME_EXPORT_REPLICA`.
3. We updated the `typeormLoader` file to have the mainDB connection use the 2nd read replica as slaves which will be automatically used for any read operation that uses the main DB connection by default. Only the export queries would use the export connection which will hit the 1st read replica. In case we dont have a 2nd read replica, all the reads apart from export queries would hit the main RDS instance.

PR #1692, opened 2024-06-25
Throw error when user not found in ExperimentController

KD1712:


RidhamShah:
This changes should be on all version of ExperimentClientController.

backend/packages/Upgrade/src/api/middlewares/UserCheckMiddleware.ts
```diff
@@ -0,0 +1,36 @@
...
+export class UserCheckMiddleware {
+  constructor(public settingService: SettingService, public experimentUserService: ExperimentUserService) {}
+
+  public async use(req: AppRequest, res: AppRequest, next: express.NextFunction): Promise<any> {
+    try {
+      const user_id = req.get('User-Id');
+      req.logger.child({ user_id });
+      req.logger.debug({ message: 'User Id is:', user_id });
+      console.log(user_id);
+      if (req.url !== '/api/init') {
```
RidhamShah, line 18: There is no need to check userId from the header for the init call, so bring this if statement on the top.

RidhamShah (changes requested):

VivekFitkariwala:
@bcb37 @danoswaltCL, At present we are fetching the user doc in each HTTP request of Experiment Client and that is redundant. In this PR is moving that in the middleware and the error will be thrown at the place where the document is fetched instead of the service which is incorrect.
We can have this middleware in the latest v5 version and not touch the service for backward compatibility.

danoswaltCL:
Another consideration is that this will break current non-client-library usage of the API, such as Livehint AI. They will need to begin sending this in the header for all calls.

danoswaltCL:
Let's wait on merging this in, see my comments in https://github.com/CarnegieLearningWeb/UpGrade/pull/1606

VivekFitkariwala (approved):

danoswaltCL:
holding off on this one, going with existing 404 errors in the individual endpoints still for now

danoswaltCL:
Needs to be only a part of the v6 controller. see: https://github.com/CarnegieLearningWeb/UpGrade/issues/1851 @ppratikcr7 @VivekFitkariwala

ppratikcr7:
Closing this as its handled in a new [PR](https://github.com/CarnegieLearningWeb/UpGrade/pull/1955) with v6 client controller.

PR #1710, opened 2024-06-28
Not able to create feature flag without adding tags

Yagnik56:
This PR fixed the issue of tags sending null instead of array which wasn't allowing to create a feature flag.


danoswaltCL:
i see the `tags` array is being initialized with `[]` , that looks like why null is coming, should be an empty array `[[]]`.

danoswaltCL (approved):

PR #1711, opened 2024-07-01
sort functionality for feature flag root table

Yagnik56:
This PR adds sort functionality for the Feature flag root page table


danoswaltCL (approved):
😎

PR #1712, opened 2024-07-02
Added pagination call on scroll in FF

KD1712:
I have used 'wheel' event in this PR because 'scroll' event is not working


frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
```diff
@@ -29,15 +31,37 @@ import { CommonStatusIndicatorChipComponent } from '../../../../../../../../shar
...
   @Input() isLoading$: Observable<boolean>;
 
+  isAllFeaturFlagFetched = false;
+  isAllFeatureFlagsFetchedSub = new Subscription();
+
+  constructor(private featureFlagsService: FeatureFlagsService) {}
+
+  ngOnInit(): void {
+    this.isAllFeatureFlagsFetchedSub = this.featureFlagsService.isAllFlagsFetched$.subscribe(
+      (value) => (this.isAllFeaturFlagFetched = value)
```
danoswaltCL, line 51: I know this is how we have done it elsewhere, but this is a pattern we should make us pause. Any time we're subscribing to something from the store$ directly in the component just to "pull out" the latest value and save local copies of it, we are creating places for state-management bugs to creep in. Usually there's a good way to make these subscriptions go away.

Here it should be easy to refactor I think: the component doesn't need to know anything about what happens after it dispatches the fetchFlagsOnScroll action. It can always just dispatch `fetchFlagsOnScroll` and let the pre-existing filters in the `fetchFeatureFlags$` effect be where we check `isAllFlagsFetched$`. That effect itself is looking a little crazy, but that's a different story, can you try checking if allFlagsFetched in the effect instead of the component?

KD1712:
So now i have removed the subscription for  `isAllFeaturFlagFetched` from the component and updated effect to check isAllFlagsFetched state to prevent unnecessary calls. Ensured proper parameter indexing in the filter logic for accurate state checks and set loader to false when all flags are fetched

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
```diff
@@ -47,6 +50,9 @@ export class FeatureFlagRootSectionCardTableComponent implements OnInit {
     }
   }
 
+  isAllFeaturFlagFetched = false;
```
danoswaltCL, line 53: these properties can be removed now

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
```diff
@@ -36,9 +37,16 @@ export class FeatureFlagsEffects {
         this.store$.pipe(select(selectTotalFlags)),
         this.store$.pipe(select(selectSearchKey)),
         this.store$.pipe(select(selectSortKey)),
-        this.store$.pipe(select(selectSortAs))
+        this.store$.pipe(select(selectSortAs)),
+        this.store$.pipe(select(selectIsAllFlagsFetched))
       ),
-      filter(([fromStarting, skip, total]) => skip < total || total === null || fromStarting),
+      filter(([fromStarting, skip, total, searchKey, sortKey, sortAs, isAllFlagsFetched]) => {
+        if (isAllFlagsFetched) {
```
danoswaltCL, line 44: it appears like we don't set the `isLoadingFeatureFlags` to true until after this, it seems like this should suffice like the other filters that cancel this action though, do this work?

```javascript
filter(([fromStarting, skip, total, searchKey, sortKey, sortAs, isAllFlagsFetched]) => {
  return !isAllFlagsFetched || skip < total || total === null || fromStarting;
})
```

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="flags-list-container">
-  <div scroll (scrolled)="fetchFlagsOnScroll()" class="flags-list-table-container" #tableContainer>
-    <mat-progress-bar class="spinner" mode="indeterminate" *ngIf="isLoading$ | async"></mat-progress-bar>
+   <div (wheel)="fetchFlagsOnScroll()" class="flags-list-table-container" #tableContainer>
+    <mat-spinner class="spinner" mode="indeterminate" *ngIf="isLoading$ | async"></mat-spinner>
```
danoswaltCL, line 3: @zackcl @Yagnik56 now that the mat-progress-bar is showing up correctly in our table elements (the progress bar at top of the table border that animates while loading), do we feel like this should remain `mat-progress-bar`?

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
KD1712: No, actually we do set `isLoadingFeatureFlags` to **true** in the reducer on `actionFetchFeatureFlags`, which is why I explicitly made `isLoadingFeatureFlags` to **false** in the effect when all flags are fetched.

So either we can :
- set `isLoadingFeatureFlags` to **false** explicitly in Effects
or
-  remove state handling on `actionFetchFeatureFlags` in Reducer

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
```diff
@@ -50,7 +50,6 @@ export class FeatureFlagRootSectionCardTableComponent implements OnInit {
     }
   }
 
-  isAllFeaturFlagFetched = false;
   isAllFeatureFlagsFetchedSub = new Subscription();
```
danoswaltCL, line 53: `isAllFeatureFlagsFetchedSub` this isn't used either right? please remove

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
danoswaltCL: @KD1712 per the discussion in mtg yesterday, we want to keep it as mat-progress-bar, i think this PR is ready otherwise

PR #1719, opened 2024-07-02
feature flag lists backend changes

bcb37:
With these changes, the feature flag endpoints will not be used to perform CRUD operations on the inclusion and exclusion lists (with the exception of GET by :id, which will return the lists along with the other feature flag data)

Operations on the lists will be performed using the segments endpoints, which have three new optional attributes:
- enabled: boolean
- includedInFeatureFlag: FeatureFlag
- excludedFromFeatureFlag: FeatureFlag

Example body for flag creation (POST to /api/flags endpoint):
```
{
    "name": "Test Flag",
    "key": "TEST_FLAG",
    "description": "Test Description",
    "status": "enabled",
    "context": [
        "assign-prog"
    ],
    "tags": [
        "tag99",
        "tag96"
    ],
    "filterMode": "includeAll"   
}
```

Example body to add or edit an inclusion list for the above (POST to /api/flags/addInclusionList endpoint):
```
{
    "flagId": "78f3293e-c797-44e2-81f3-8c6a6b830003",
    "enabled": true,
    "listType": "segment",
    "list": {
        "id": "2f86e720-d8b9-47b4-88f3-fa9784784902", // if editing
        "name": "seg to seg",
        "context": "assign-prog",
        "userIds": [],
        "groups":[],
        "subSegmentIds":["fe826649-cacf-4c3d-9f79-e96e05307892"]
    }
    }
```

Deleting lists can be done by hitting the `/api/segments/:id` DELETE endpoint

This would use the ID of the feature flag; it's for a 'segments' list, which can only have one element.

assuming the above UUIDs refer to entities that exist in the db.


danoswaltCL:
Ok, I'm working on the modals for adding/modifying a flag...we have slightly different constraints for add and modify requests that we can enforce with types, so would you say these are accurate representations of our types (FeatureFlag being the full document from the db that we should expect as a response?)

```
export interface AddFeatureFlagRequest {
  name: string;
  key: string;
  description?: string;
  context: string[];
  tags: string[];
}

export interface ModifyFeatureFlagRequest {
  name?: string;
  key?: string;
  description?: string;
  status?: FEATURE_FLAG_STATUS;
  filterMode?: FILTER_MODE;
  context?: string[];
  tags?: string[];
  enabled?: boolean;
}

export interface FeatureFlag {
  createdAt: string;
  updatedAt: string;
  versionNumber: number;
  id: string;
  name: string;
  key: string;
  description: string;
  status: FEATURE_FLAG_STATUS;
  filterMode: FILTER_MODE;
  context: string[];
  tags: string[];
  enabled: boolean;
}
```

bcb37:
> Ok, I'm working on the modals for adding/modifying a flag...we have slightly different constraints for add and modify requests that we can enforce with types, so would you say these are accurate representations of our types (FeatureFlag being the full document from the db that we should expect as a response?)
> 
> ```
> export interface AddFeatureFlagRequest {
>   name: string;
>   key: string;
>   description?: string;
>   context: string[];
>   tags: string[];
> }
> 
> export interface ModifyFeatureFlagRequest {
>   name?: string;
>   key?: string;
>   description?: string;
>   status?: FEATURE_FLAG_STATUS;
>   filterMode?: FILTER_MODE;
>   context?: string[];
>   tags?: string[];
>   enabled?: boolean;
> }
> 
> export interface FeatureFlag {
>   createdAt: string;
>   updatedAt: string;
>   versionNumber: number;
>   id: string;
>   name: string;
>   key: string;
>   description: string;
>   status: FEATURE_FLAG_STATUS;
>   filterMode: FILTER_MODE;
>   context: string[];
>   tags: string[];
>   enabled: boolean;
> }
> ```

This looks correct, with the following exceptions:

- Add and Modify should be the same: they both should have 'context' and 'filterMode'
- Neither Add nor Modify should have 'enabled', which is not a thing (that's handled by 'status)
- The response from the GET endpoint for a single Feature Flag would also contain two arrays: 'featureFlagSegmentInclusion' and 'featureFlagSegmentExclusion', which would be arrays of private segments.

VivekFitkariwala:
@bcb37 I am not sure why you have deleted the junction table of `Feature Flag Segment Exclusion` and `Feature Flag Segment Exclusion`. Because of this, the relationship data is now moved into the Segment table. If you had added the enabled property in these junction tables and changed the OneToOne relationship to OneToMany in the Feature Flag table you could have achieved the same thing.

bcb37:
> @bcb37 I am not sure why you have deleted the junction table of `Feature Flag Segment Exclusion` and `Feature Flag Segment Exclusion`. Because of this, the relationship data is now moved into the Segment table. If you had added the enabled property in these junction tables and changed the OneToOne relationship to OneToMany in the Feature Flag table you could have achieved the same thing.

I was thinking it's better to simplify the database and data models. We can now store the information we need with fewer moving parts, and query with fewer joins. And now we can manage the inclusions and exclusions from the front end just using the segment endpoints that already exist. Is there a compelling reason to use those junction tables for the relationships between feature flags and segments?

VivekFitkariwala:
@bcb37 this change is adding property which doesn't belong to Segment and that is the concern. I don't know if we are facing any issues because of this specific join.
Thinking ahead when we have to refactor the Experiment we will end up adding those specific data also in the Segment table. The present design of table is better I think.

bcb37:
> @bcb37 this change is adding property which doesn't belong to Segment and that is the concern. I don't know if we are facing any issues because of this specific join. Thinking ahead when we have to refactor the Experiment we will end up adding those specific data also in the Segment table. The present design of table is better I think.

Ok, that makes sense to restore the junction tables. I think we'll still need new endpoints for editing the lists. They should be relatively similar to the /segment endpoints, with the addition of managing the junction table entries. Since we don't create feature flags initially with any exclusion or inclusion lists, I still think it makes sense to edit them separately.

danoswaltCL:
@VivekFitkariwala please consider this the highest priority review...

putting comments from dev sync in writing: it was suggested that `PUT api/flags/addInclusionList/:id` for edit may be preferred for the edit flow over `POST /api/flags/addInclusionList`.

also enums for the `listType` were discussed, but I think we landed on leaving it as is because group `listType` is not constant, comes from the context metadata.

Zack mentioned that it might be best API practice to only need to send changed fields when updating a known entity, but that is a larger architectural design change, currently we tend to always pass the entire document (correct me if this what was basically discussed)

bcb37:
> @VivekFitkariwala please consider this the highest priority review...
> 
> putting comments from dev sync in writing: it was suggested that `PUT api/flags/addInclusionList/:id` for edit may be preferred for the edit flow over `POST /api/flags/addInclusionList`.
> 
> also enums for the `listType` were discussed, but I think we landed on leaving it as is because group `listType` is not constant, comes from the context metadata.
> 
> Zack mentioned that it might be best API practice to only need to send changed fields when updating a known entity, but that is a larger architectural design change, currently we tend to always pass the entire document (as was my understanding of that concern)

I spent all day making changes and updating the tests, but it doesn't look like those commits ended up on the branch. I'll try again tomorrow.

backend/packages/Upgrade/src/api/controllers/SegmentController.ts
```diff
@@ -53,6 +53,12 @@ export interface getSegmentData {
  *         type: array
  *         items:
  *           type: string
+ *       enabled:
```
VivekFitkariwala, line 56: Do we need this?

backend/packages/Upgrade/src/api/models/FeatureFlagSegmentExclusion.ts
```diff
@@ -9,10 +9,16 @@ export class FeatureFlagSegmentExclusion extends BaseModel {
...
     primary: true,
   })
   @JoinColumn()
   public featureFlag: FeatureFlag;
+
+  @Column()
+  public enabled: boolean;
+
+  @Column()
+  public listType: string;
```
VivekFitkariwala, line 23: We can use enum in the database but if we don't want that we should use string literals in the type.

backend/packages/Upgrade/src/api/models/FeatureFlagSegmentExclusion.ts
```diff
@@ -9,10 +9,16 @@ export class FeatureFlagSegmentExclusion extends BaseModel {
...
 
-  @OneToOne(() => FeatureFlag, (featureFlag) => featureFlag.featureFlagSegmentExclusion, {
+  @ManyToOne(() => FeatureFlag, (featureFlag) => featureFlag.featureFlagSegmentExclusion, {
     onDelete: 'CASCADE',
     primary: true,
   })
   @JoinColumn()
   public featureFlag: FeatureFlag;
+
+  @Column()
```
VivekFitkariwala, line 19: We can add a default value as false.

backend/packages/Upgrade/src/api/models/FeatureFlagSegmentInclusion.ts
```diff
@@ -9,10 +9,16 @@ export class FeatureFlagSegmentInclusion extends BaseModel {
...
   public segment: Segment;
 
-  @OneToOne(() => FeatureFlag, (featureFlag) => featureFlag.featureFlagSegmentInclusion, {
+  @ManyToOne(() => FeatureFlag, (featureFlag) => featureFlag.featureFlagSegmentInclusion, {
     onDelete: 'CASCADE',
     primary: true,
   })
   @JoinColumn()
   public featureFlag: FeatureFlag;
+
```
VivekFitkariwala, line 18: Same as in Segment Exclusion

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -110,17 +108,39 @@ export class FeatureFlagService {
...
     logger.info({ message: `Delete Feature Flag => ${featureFlagId}` });
-    const featureFlag = await this.featureFlagRepository.find({
-      where: { id: featureFlagId },
+    return getConnection().transaction(async (transactionalEntityManager) => {
+      const featureFlag = await this.findOne(featureFlagId, logger);
+
+      if (featureFlag) {
+        const deletedFlag = await this.featureFlagRepository.deleteById(featureFlagId, transactionalEntityManager);
+
+        featureFlag.featureFlagSegmentInclusion.forEach(async (segmentInclusion) => {
```
VivekFitkariwala, line 117: Do we need this? The cascade delete should automatically delete this.

VivekFitkariwala: In that case the transaction is not required

VivekFitkariwala (changes requested):

VivekFitkariwala:
@danoswaltCL Sorry for the delay. I have added my review comments.

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -475,13 +417,241 @@ export class FeatureFlagsController {
...
+   *         - Feature Flags
+   *       produces:
+   *         - application/json
+   *       responses:
+   *          '200':
+   *            description: Delete Feature Flag Inclusion List by segment Id
+   */
+  @Delete('/inclusionList/:id')
+  public async deleteInclusionList(
+    @Param('id') segmentId: string,
```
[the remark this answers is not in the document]
bcb37: This seems to be how single UUID parameters are verified throughout the codebase. If we create a validator that checks whether something's a UUID, we should probably start using it everywhere. Maybe a separate task for that.

VivekFitkariwala (approved):

PR #1728, opened 2024-07-03
replaced add include/exclude list button to Add

Yagnik56:
This PR replaced the add include/exclude list button with a simple Add as we discussed.


zackcl:
@Yagnik56 @danoswaltCL Are we sure about this? If we simply make the button labels "Add", shouldn't we also update the "Add Feature Flag" and "Add Segment" button labels for consistency?

danoswaltCL:
🤷‍♂️ I have no opinion, just making a ticket to remember Steve's suggestion.

zackcl:
@danoswaltCL How about we ask one more time before we merge this? Or maybe I can create and share two separate design images so that we can compare.

zackcl:
Closing this as we decided to keep the button labels for now.

PR #1730, opened 2024-07-03
Import-Export buttons in common tag component

KD1712:
This PR adds Action buttons which toggle between Import/Export icons w.r.t Tag value


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-tag-input/common-tag-input.component.ts
```diff
@@ -18,6 +18,17 @@ import { TranslateModule } from '@ngx-translate/core';
 // Example Usage:
 //   <app-common-tags-input formControlName="tags"></app-common-tags-input>
 
+// To add Import/Export button while using component
+// you can add the property 'showUploadIcon' or 'showDownloadIcon'
+// with its repective binding 'uploadClicked' or 'downloadClicked'
```
danoswaltCL, line 23: The tsdoc looks like it needs some updates, the wording is a little different and the usage should show import also. Could even do two examples to demonstrate that the typical usage is simpler than when dealing with buttons:

```
// Typical usage
// <app-common-tags-input
//         formControlName="tags"
//       ></app-common-tags-input>
//
// Manage built-in optional action buttons
// <app-common-tags-input
//         [showExportIcon]="true"
//         (exportClicked)="onExportClick()"
//         [showExportIcon]="false"
//         (importClicked)="onImportClick()"
//         formControlName="tags"
//       ></app-common-tags-input>
```

danoswaltCL:
can you put a screenshot on here?

are we okay with allowing the state to potentially be "true" for both buttons, or should this strictly be a toggle (can't have a state where both are true)? Potentially "action buttons" like these that will live in an input deserve to be their own widget and have a slot like we do for section card header-right action buttons. @zackcl @Yagnik56

Yagnik56:
> can you put a screenshot on here?
> 
> are we okay with allowing the state to potentially be "true" for both buttons, or should this strictly be a toggle (can't have a state where both are true)? Potentially "action buttons" like these that will live in an input deserve to be their widget and have a slot like we do for section card header-right action buttons. @zackcl @Yagnik56

it should be a toggle, when tags are empty it should show import otherwise export. @khanjan, please refer to design and make the change accordingly

KD1712:
Screenshot 2024-07-04 at 3 10 35 PM
Screenshot 2024-07-04 at 12 27 18 PM

The icons will toggle accordingly if we edit the tags. Just need to add functionality for icon click in parent component

danoswaltCL:
@zackcl the icon used looks different from your design, what do you think?

KD1712:
Screenshot 2024-07-15 at 12 11 56 PM
Screenshot 2024-07-15 at 12 12 08 PM

Previously i used mat-icons that were available in the list. But now changed to google icon as there is none similar to Figma design in Angular material

KD1712:
@zackcl yes the '**edit/add**' modals don't have these buttons. Just for the display check i've used it in Edit modal and shared. **No changes are made in Add/Edit modal**

danoswaltCL (approved):

PR #1741, opened 2024-07-09
Collapse all section cards in FF details page

KD1712:
Added collapsible functionality to toggle all section cards by giving input to 'Inclusion,Exclusion and Exposure' section cards and emitting output from 'overview details'


danoswaltCL:
hold up, we need to add something to the store to handle individual details page ui state so we can handle things like this with redux patterns to avoid making data traverse up and down the component tree in this way. i will get back to you on this, it is end of day...

danoswaltCL:
well this is perfectly fine I'll approve, I would rather that we handle kind of thing with the ngrx store / redux patterns but we may need some additional stuff in the reducer to support that better.

danoswaltCL (approved):

PR #1748, opened 2024-07-10
Update demo-app.js for the demo app's tour to work with UpGrade v5.2.0

### LINKED ISSUE #1538: Update Demo App to use the latest version of UpGrade
Update the demo app to use the latest version of UpGrade for Roster integration.

zackcl:
This PR updates the `frontend/projects/upgrade/src/assets/js/demo-app.js` file used by the [demo app](https://github.com/CarnegieLearningWeb/upgrade-demo-app) to interact with UpGrade v5.2.0's (mdc-migrated) frontend for guided tours.

Closes #1538

### DIFF
--- a/frontend/projects/upgrade/src/assets/js/demo-app.js
+++ b/frontend/projects/upgrade/src/assets/js/demo-app.js
@@ -5,80 +5,88 @@ if (window.self !== window.top) {
   const getElementById = (id) => {
     switch (id) {
       // Home
-      case 'login-with-google-button':
-        return document.querySelector('div.login-container button.google-sign-in-btn');
+      case 'upgrade-logo-link':
+        return document.querySelector('div.logo a.logo-link');
       case 'experiments-tab':
         return document.querySelectorAll('div.list-item-container a.nav-item')[0];
       case 'signout-button':
-        return document.querySelector('span.mat-list-item-content a.logout-link');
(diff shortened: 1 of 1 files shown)


danoswaltCL:
not sure who or what is to be reviewed, I have to assume this is in working condition if this is the standard operating procedure for selecting elements.

can you explain the reason for this switch list though? is it not possible for you to put unique selectors on your elements in the html itself that can be selected withs `id` or `data-` selectors? like:

`<div data-upgrade-logo-link>`

```javascript
function getElementById(id) {
  const selector = `[data-${id}]`;
  return document.querySelector(selector);
}
```

danoswaltCL:
o wait duh, i forgot that these selectors are picking up on elements from the upgrade app. still, that switch list makes my stomach hurt, my code review brain wants desperately want to turn that into an object map, but that's just me looking at this one file at a glance.

danoswaltCL (approved):

zackcl:
Yes, those selectors need to be in that file to access the UI elements from UpGrade. I agree that some parts of the code can be moved to the demo app, but it's difficult to generalize the element selecting pattern since new tours might require different ways to access UpGrade elements. I plan to clean up the code in the `demo-app.js` file once we are sure no new tours need to be added.

Issue #1749, opened 2024-07-10
FF: General Add List Modal


danoswaltCL:
Implement a modal that can be utilized for adding or editing lists for both Inclusion and Exclusion flows.

We should start with just the initial `type` selector, then other work can follow. Depending on the type selected, form will act differently, so we can start here and then develop the rest.

Screenshot (Updated):

<img width="688" alt="Feature Flag - Add Include List" src="https://github.com/user-attachments/assets/d9ebfce3-42ac-42b1-9ae0-a233369072c1">

Considerations:

- This modal will need to be templated for use in both the Include and Exclude List functionality
- This modal will need to be able to handle both Add and Edit functions

When the user opens this modal, we will only show a type selector at first. They can select:
- Segment
- Individual
- Group (listed by each group type supported in that context)

danoswaltCL:
hola, the design shows (`All` won't be used now, ignore that):

`Individual`
`Segment`
`groupId1 (group)`
`groupId2 (group)`
`groupId3 (group)`

![Image](https://github.com/CarnegieLearningWeb/UpGrade/assets/97542869/b3137d89-57db-49e9-92a2-6290319880c4)

Couple thoughts:

1. Is this the preferred order from user perspective? Segment will probably be the most used thing right, can it be topmost?

![Image](https://github.com/CarnegieLearningWeb/UpGrade/assets/97542869/c536f419-dc14-45a2-a8f9-eedc94a43e2e)


2. While Zack was out I suggested a different format for how we show these groups [here] so they look more like the other options (https://carnegielearning.slack.com/archives/C02QA168MNW/p1718981331228129). That would be an app-wide change that can be implemented outside this later, but that would give us a list that looks like this for `assign-prog`. What do you think about that @zackcl 

That would look like this:
![Image](https://github.com/CarnegieLearningWeb/UpGrade/assets/97542869/b2a7bbf1-711b-46c1-8f91-9c61b39ce15a)

@amurphy-cl @bcb37

zackcl:
Yeah, I like that! I can update the options to the following in Figma if @amurphy-cl confirms:

```
Segment
Individual
Group: schoolId
Group: classId
Group: instructorId
```

What I'm not totally sure about is how the group types are ordered. It looks like it's not alphabetical. Do you know how the order is determined?
[...]

PR #1756, opened 2024-07-11
Feature/add list modal

danoswaltCL:
Initial add-list-modal #1749 

This creates the base modal and the type field to get this general modal started.
It also implements it for the `Add Include List` button.

Eventually this modal is planned to be used for Add and Edit list flows in both the include / exclude flows.

You'll notice a few placeholder things, i tried to comment most places that are expected to need to change as we go.

image

Feature Flag - Add Include List

### DIFF
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
@@ -43,14 +43,14 @@ export interface UpdateFeatureFlagStatusRequest {
   status: FEATURE_FLAG_STATUS;
 }
 
-export enum UpsertFeatureFlagListAction {
+export enum UPSERT_FEATURE_FLAG_LIST_ACTION {
   ADD = 'add',
   EDIT = 'edit',
 }
 
 export interface UpsertFeatureFlagListParams {
   sourceList: any; // TODO define me
-  action: UpsertFeatureFlagListAction;
+  action: UPSERT_FEATURE_FLAG_LIST_ACTION;
 }
 
 export enum LIST_OPTION_TYPE {
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
@@ -43,6 +43,21 @@ export interface UpdateFeatureFlagStatusRequest {
   status: FEATURE_FLAG_STATUS;
 }
 
+export enum UpsertFeatureFlagListAction {
+  ADD = 'add',
+  EDIT = 'edit',
+}
+
+export interface UpsertFeatureFlagListParams {
+  sourceList: any; // TODO define me
+  action: UpsertFeatureFlagListAction;
+}
+
+export enum LIST_OPTION_TYPE {
+  INDIVIDUAL = 'Individual',
+  SEGMENT = 'Segment',
+}
+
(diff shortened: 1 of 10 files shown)


frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
```diff
@@ -43,6 +43,21 @@ export interface UpdateFeatureFlagStatusRequest {
...
+
+export interface UpsertFeatureFlagListParams {
+  sourceList: any; // TODO define me
+  action: UpsertFeatureFlagListAction;
+}
+
+export enum LIST_OPTION_TYPE {
+  INDIVIDUAL = 'Individual',
+  SEGMENT = 'Segment',
+}
```
zackcl, line 59: Do we use any naming conventions for the `enum` types? `ALL_CAPS` or `UpperCamelCase`? Also, do we need to define these in a separate file (`types/src/Experiment/enums.ts`)? I have the same question for the `interface` type.

frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -87,6 +92,32 @@ export class DialogService {
     return this.dialog.open(UpdateFlagStatusConfirmationModalComponent, config);
   }
 
+  openAddIncludeListModal() {
+    const commonModalConfig: CommonModalConfig<UpsertFeatureFlagListParams> = {
+      title: 'Add Include List',
+      primaryActionBtnLabel: 'Add',
+      primaryActionBtnColor: 'primary',
+      cancelBtnLabel: 'Cancel',
```
zackcl, line 100: The `primaryActionBtnLabel` should be 'Create' instead of 'Add'.

frontend/projects/upgrade/src/styles.scss
```diff
@@ -249,6 +249,10 @@ button[mat-icon-button] {
   width: 805px;
 }
 
+.full-width {
+  width: 100%;
+}
+
```
zackcl, line 255: Could we remove the `full-width` class for now until we determine it's widely used and needed? I think it might be better to add this later.

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
danoswaltCL: our various type enums and constants such are definitely not standardized / organized very well. I tend to use all caps for any kind of enum or constant that produces a value, but upper-camel-case for type interfaces

frontend/projects/upgrade/src/styles.scss
danoswaltCL: i added it precisely because it was becoming a common pattern, search for `class="full-width"` in the feature-flags project files.

that being said, I'm kind of an idiot when it comes to organizing css styles, so I'm all ears for better ideas. I can localize it to the component if that's what you'd prefer for now.

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.model.ts
Yagnik56: you have used `UpsertFeatureFlagListAction` enum can you change that...

danoswaltCL:
Comments resolved on this one. This one is not dependent on any backend changes, so it should be good to go now.

Yagnik56 (approved):

PR #1782, opened 2024-07-25
⬆️  Release v5.2 : Update Typeorm version to 0.3.0

ppratikcr7:
This PR covers below changes:

1. Update typeORM and all related depreciated code to new code based off `release/v5.1` branch.

A few issues which exists in v5.1 branch will be observed in this PR as well. They are resolved in dev branch. So, the below issues would be addressed once we merge dev.

- Preview participants editing gives error

- Segments search feature not working

- Names not visible in audit logs


bcb37:
I think we should create a new branch, which would be a clone of 'release/v5.1' and call it 'release/v5.2'. Then we can merge these commits onto that to prepare to release it.

backend/packages/Upgrade/src/loaders/typeormLoader.ts
```diff
@@ -54,7 +53,7 @@ export const typeormLoader: MicroframeworkLoader = async (settings: Microframewo
     name: CONNECTION_NAME.REPLICA,
     type: env.db.type as 'postgres',
     replication: {
-      master: masterHost,
+      master: exportReplicaHost,
```
bcb37, line 192: I don't think we want to add this as part of the typeorm update release. We're still using the read replica only for exports, not all db reads.

ppratikcr7: Just to clarify once more, there is a separate postgres connection string called "export" that would call this particular connection options. And we use this connection string explicitly for only export related queries. All the other DB reads are by default calling the main DB connection option above. Still, will update this back to old where all the export related write queries are hit on `masterHost` and all export related read queries will hit the `exportReplicaHost` under the slaves mode. Any change to the `exportReplicaDBConnectionOptions` will never affect any other DB read queries apart from where its specifically targeting this 'export' connection string in the queries.

ppratikcr7: But, yes due to your comment, I found that the mainDBConnection has slaves set to `replicaHost` which would evenly redirect reads to replica. I will update this as well. Thanks :)

ppratikcr7: You can refer typeorm [documetation](https://orkhan.gitbook.io/typeorm/docs/multiple-data-sources#replication) for more details. Adding a small snippet out of it for a better understanding for all.

Screenshot 2024-08-02 at 6 04 53 PM

bcb37 (commented):
This looks good as far as I can tell. We'll deploy this branch to staging for testing if nobody else has any objections.

VivekFitkariwala:
The PR looks good to me.

backend/packages/Upgrade/src/api/repositories/LogRepository.ts
```diff
@@ -685,7 +687,7 @@ export class LogRepository extends Repository<Log> {
     isFactorialExperiment: boolean,
     unitOfAssignment: string
   ): SelectQueryBuilder<Experiment> {
-    const experimentRepo = getCustomRepository(ExperimentRepository, 'export');
+    const experimentRepo = Container.getCustomRepository(ExperimentRepository);
```
bcb37, line 690: For now, we need to continue to use the 'export' custom repository so that the log queries associated with exports are still performed on the replica.

bcb37 (changes requested):
I found one more thing that should be changed while testing. Also, when we do the release we should update all the version numbers to 5.2.

backend/packages/Upgrade/src/api/repositories/LogRepository.ts
ppratikcr7: Okay, I thought we don't want that. I Will push the update.

ppratikcr7:
> I found one more thing that should be changed while testing. Also, when we do the release we should update all the version numbers to 5.2.

Should I push this commit as well with the change to add replica for analytics query?

danoswaltCL:
yeah Pratik, there's that command you can run on the root of the project, npm run version:minor, I'd just add that with the commit.

ppratikcr7:
> yeah Pratik, there's that command you can run on the root of the project, npm run version:minor, I'd just add that with the commit.

Yeah, I know that command. Thanks. Are you going to push the version update and export connection string change? Or should I do it?

danoswaltCL:
oops, missed that comment, wish I'd seen it earlier. you can as I'm not 100% sure what the right change Ben is asking for, he's just saying we want `getCustomRepository(ExperimentRepository);` without `export` named, right?

ppratikcr7:
> oops, missed that comment, wish I'd seen it earlier. you can as I'm not 100% sure what the right change Ben is asking for, he's just saying we want `getCustomRepository(ExperimentRepository);` without `export` named, right?

Okay, I will push the changes. Actually I removed that 'export' connection string for analytics query, but ben wants to have that for now so that they are triggered on read replica.

danoswaltCL:
I tried to update it because I didn't see any changes pushed to this branch from the time Ben requested the change, so I tried changing the logRepository line back passing in `export`. but there were test errors. so I backed out the change I made. I then went ahead pushed the version update part. @ppratikcr7 I must be confused, can you make sure the right change is in for Friday so we can release it?

VivekFitkariwala (dismissed):

danoswaltCL (approved):

VivekFitkariwala (commented):
Looks good

PR #1784, opened 2024-07-25
Feature/add list modal segments

danoswaltCL:
- Adds a configurable upsert-private-segment-list modal.
- Implements the modal for Add Feature Flag Inclusions List.
- Adds a close-modal interceptor that other modals can use to simplify close-modal logic
- Adds updated details to store on success

Other notes:
- Will need to spawn a story to fix the `common-tag-input` so that it doesn't hard-code certain things like the label, currently it says "Tags (optional)" instead of "Values".
- The upload/download action button for tags has not been wired up either.
- Was mindful to create this in a way that it should be reusable for the following actions with minimal additional code:

```
export enum UPSERT_PRIVATE_SEGMENT_LIST_ACTION {
  ADD_FLAG_INCLUDE_LIST = 'add_flag_include', <--- one down, 5 to go!
  EDIT_FLAG_INCLUDE_LIST = 'edit_flag_include',
  ADD_FLAG_EXCLUDE_LIST = 'add_flag_exclude',
  EDIT_FLAG_EXCLUDE_LIST = 'edit_flag_exclude',
  ADD_SEGMENT_LIST = 'add_segment_list',
  EDIT_SEGMENT_LIST = 'edit_segment_list',
}
```

Please be savage in reviews, this piece in particular will suck later if we don't get it the way we like.


danoswaltCL:
i have no idea why a couple tests are failing, things work normally in the UI for the cases that are supposedly failing, I will look into that tomorrow.

danoswaltCL:
FYI the upsert on success seems not to be working and the primary button doesn't disable properly while the response is pending. will look into these tomorrow also, please review anyway, these are minor fixes...

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.scss
```diff
@@ -0,0 +1,14 @@
...
+  width: 100%;
+
+  .mat-mdc-form-field {
+    width: 100%;
+  }
+
+  .mat-mdc-form-field:not(:last-child) {
+    margin-bottom: 20px;
+  }
+}
```
zackcl, line 14: I think it's OK to remove the whole style from this file. The `upsert-private-segment-list-container` selector doesn't exist anyway.

frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
```diff
@@ -97,6 +101,24 @@ export class SegmentsEffects {
...
+          ...list,
+        };
+        return this.featureFlagDataService.addInclusionList(request).pipe(
+          map((listResponse) => FeatureFlagsActions.actionUpsertFeatureFlagInclusionListSuccess({ listResponse })),
+          catchError((error) => of(FeatureFlagsActions.actionUpsertFeatureFlagInclusionListFailure({ error })))
+        );
+      })
+    )
+  );
+
```
zackcl, line 121: Are we sure this code belongs in `segments.effects` instead of `feature-flags.effects?` It seems like all the `FeatureFlagsActions` are in that file.

danoswaltCL:
woah please do not commit things to my branch without asking, make a PR comment please.

frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
danoswaltCL: yeah i went back and forth, originally i had everything in feature flags, then you pointed out it was also a segments thing, so then it seemed more natural in the segments stuff, and then I ended up moving things mostly back over to feature flags... anyway I think you're right though I'll move it

zackcl:
Maybe we should discuss how to handle situations where the user types a non-existent segment or doesn't select one from the options. Also, if there are no defined segments, it might be better to disable the "Segment" type or field to avoid confusion.

Screenshot 2024-07-26 at 10 20 30 PM

danoswaltCL:
I'm surprised it doesn't flag a typed selection as invalid, seems like something else might be off if it's letting you submit an invalid value... but yeah, if there's nothing in the list to show maybe some kind of message to say how to add a segment, with a link to the segments page? I think if the option is disabled or doesn't show up it might not occur to the user that they can even add public segments.

zackcl:
@danoswaltCL Here is a summary of our discussion:

- If there are no defined segments, the "Segment" option in the "Type" drop-down should be disabled and greyed out.
- If the user types something into the "Segment" field but doesn't select one from the options, when the field loses focus, it should turn red and display a message like "A segment should be selected from the options" below it. I think this approach is better than validating when the "Create" button is clicked, although we should still call the validation endpoint finally.
- As the user types into the "Segment" field, the options should be filtered.

danoswaltCL:
Ok see the video, I think I addressed the things we talked about. There is a property in the MatAutoComplete called `requireSelection` that will trigger validation error on this field and clear the field if a vaild value is not selected. That pretty much takes care of issues with not-selecting things and preventing submitting and getting an error.

I put a mat-hint underneath with a message about needing to select. We said do it on validation error but I wanted to show just having a normal mat-hint seems pretty natural there to help the user understand what that field is about, let me know what you think. @zackcl @amurphy-cl 

https://github.com/user-attachments/assets/559f8087-7798-42f3-95cc-f47cbb8ea1ed

danoswaltCL:
anything needed for this? it's still sitting here? anyone reviewing it?

zackcl (approved):

frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
```diff
@@ -114,7 +117,27 @@ const reducer = createReducer(
...
+    return adapter.updateOne(
+      {
+        id: featureFlag?.id,
+        changes: { featureFlagSegmentInclusion: [listResponse, ...existingFlag.featureFlagSegmentInclusion] },
+      },
+      { ...state }
+    );
+    return adapter.upsertOne(featureFlag, { ...state, isLoadingUpsertPrivateSegmentList: false });
+  }),
+  on(FeatureFlagsActions.actionUpsertFeatureFlagInclusionListFailure, (state) => {
```
zackcl, line 138: @danoswaltCL It's OK to remove this line right? This is not called because of the `return adapter.updateOne()` above it.

danoswaltCL: oops yes. I'm surprised some kind of linting error doesn't happen here since it's unreachable code.

PR #1786, opened 2024-07-26
Adding validation and 409 error to flags

RidhamShah:
### Updated Final Approach
Upon submission to the feature-flag functionality, the create/update API will be called. This API will return a 409 error if the key is not unique within the specified context (the combination of key and context must be unique). This error will be displayed in the UI as an error message beneath the key text box.

We will not take the snack-bar approach as the modal will always come above the snack-bar, and to solve that, many things need to be changed.
image


zackcl:
I think it's better to not show the snackbar error while the modal is open.

Screenshot 2024-07-26 at 4 51 34 PM

danoswaltCL:
I think there is somehow a lot of confusion about what was needed. This is the story:

```
When submitting a feature flag, either from POST or PUT, the key that is submitted should be unique per the requested context. If it is not, the user should see an error and be given the chance to correct it.

For implementation, 409 HTTP status should be sent back with a message to display in the snackbar.

"A flag with this key already exists for this app-context, please enter a unique key."
```

The ask was to create a uniqueness check for the existing endpoint, and to show a snackbar message on error.

danoswaltCL:
There is a different story to add a validation endpoint for MVP2 if we wanted to, maybe that was what was what was being thought of, but this should essentially just be a simpler backend change to the existing endpoint.

danoswaltCL (changes requested):
The pr doesn't match the requirements of the story.

RidhamShah:
@danoswaltCL updated this PR as needed.
I also have commented out the form-validation endpoint (not removed) if we want to use Infuture or other versions.

Yagnik56:
@zackcl @danoswaltCL I tried to move snackbar above the modal but I can't find the class we can target to do that, maybe we can rethink the design on mvp2 for now we can go with this.

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -16,6 +16,14 @@ interface FeatureFlagsPaginationInfo extends PaginationResponse {
   nodes: FeatureFlag[];
 }
 
+export interface FeatureFlagFormData {
```
danoswaltCL, line 19: this doesn't look like it belongs here

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -300,6 +308,39 @@ export class FeatureFlagsController {
     return this.featureFlagService.create(flag, request.logger);
   }
 
+  /**
```
danoswaltCL, line 311: this is no longer needed

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -204,6 +206,23 @@ export class FeatureFlagService {
     return this.segmentService.deleteSegment(segmentId, logger);
   }
 
+  public async validate(flagDTO: FeatureFlagValidation, logger: UpgradeLogger) {
```
danoswaltCL, line 209: `validate` is vague, we are specifically validating key + context uniqueness.

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.html
```diff
@@ -19,8 +19,17 @@
       <mat-label class="ft-14-400">Key</mat-label>
       <input matInput formControlName="key" placeholder="Key for this feature flag" class="ft-14-400"/>
       <mat-hint class="form-hint ft-12-400">
-        {{'feature-flags.upsert-flag-modal.key-hint.text' | translate}}
-        <a class="learn-more-link" href="https://www.upgradeplatform.org/" target="_blank" rel="noopener noreferrer">Learn more</a>
+        <ng-container *ngIf="!validationError; else duplicateError">
```
danoswaltCL, line 22: please remove code that isn't being used now since we are saying snackbar is place for message for now

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.scss
```diff
@@ -0,0 +1,3 @@
+.duplicate-error-message {
```
danoswaltCL, line 1: please remove

backend/packages/Upgrade/src/api/middlewares/ErrorHandlerMiddleware.ts
```diff
@@ -97,6 +97,10 @@ export class ErrorHandlerMiddleware implements ExpressErrorMiddlewareInterface {
             message = error.message;
             type = SERVER_ERROR.QUERY_FAILED;
             break;
+          case 409:
+            message = error.message;
+            type = SERVER_ERROR.DUPLICATE_KEY;
```
danoswaltCL, line 102: 409 is a general status, duplicate "key" is pretty specific. i know this is the only planned usage of 409 but this is what I'm talking about avoiding making specific custom code to just handle one specific case. The type should be something more general right?

danoswaltCL:
ok. let's talk in refinement, I want to hit pause on this story and talk as a team.

danoswaltCL:
@RidhamShah please see #1797 .

Once that is available, you'd add a new SERVER_ERROR enum for the duplicate key error, send that error from backend, and do the same steps as "EXAMPLE_ERROR" to create the observable. The form component can subscribe to that observable to show the error, and the errorEvent itself will have a method called `clear()` that you have to call when the error has been handled.

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.ts
```diff
@@ -154,16 +158,32 @@ export class UpsertFeatureFlagModalComponent {
     this.subscriptions.add(this.isSelectedFeatureFlagUpdated$.subscribe(() => this.closeModal()));
   }
 
-  onPrimaryActionBtnClicked(): void {
+  onPrimaryActionBtnClicked() {
     if (this.featureFlagForm.valid) {
-      // Handle extra frontend form validation logic here?
       this.sendRequest(this.config.params.action, this.config.params.sourceFlag);
+      // this.sendValidateRequest().then((validationError) => {
```
danoswaltCL, line 164: remove the comments please

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
danoswaltCL: can this be changed to `validateUniqueAppContextKey` or some such thing?

danoswaltCL:
We ditched the global error handling idea, so @zackcl did we want to revert back to no snackbar message, yes form hint warning?

zackcl:
I'm good with either option, whichever is easier to implement for now.

zackcl:
@RidhamShah Could you please resolve the comments by @danoswaltCL?

Also, I think we should close the modal when feature flag creation fails with a 409 error so that users can clearly see the snackbar message. I believe this is sufficient for now, but later we can consider showing the validation error message in the modal instead of the snackbar.

Screenshot 2024-09-05 at 9 32 49 PM

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: name refractored

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
RidhamShah: code removed

zackcl:
> which is to inform the user that the key is already in use and allow them to modify it. 

Can we inform the user through the Snackbar message?

> we recommend exploring the possibility of displaying a form error message beneath the key input field

If we display the form error message below the key input field, can we make the Snackbar notification to not appear in this case? I thought we couldn't find a straightforward solution to do this.

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.ts
RidhamShah: code removed

backend/packages/Upgrade/src/api/middlewares/ErrorHandlerMiddleware.ts
VivekFitkariwala: This code should not be here. We don't want all 409 status codes to be treated as a `DUPLICATE_KEY` error.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -407,6 +409,23 @@ export class FeatureFlagService {
     return this.segmentService.deleteSegment(segmentId, logger);
   }
 
+  public async validateForm(flagDTO: FeatureFlagValidation, logger: UpgradeLogger) {
+    logger.info({ message: `Validating featureFlags` });
+    const result = await this.featureFlagRepository
```
VivekFitkariwala, line 414: This should go in the Repository.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -407,6 +409,23 @@ export class FeatureFlagService {
     return this.segmentService.deleteSegment(segmentId, logger);
   }
 
+  public async validateForm(flagDTO: FeatureFlagValidation, logger: UpgradeLogger) {
```
VivekFitkariwala: This validation should also happen during the `ImportFeatureFlag`

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah:
```
if (keyExists) {
        compatibilityType = FF_COMPATIBILITY_TYPE.INCOMPATIBLE;
```

In import validation, we are sending as `Incompatible` if the key already exists

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code moved

backend/packages/Upgrade/src/api/repositories/FeatureFlagRepository.ts
```diff
@@ -111,4 +112,19 @@ export class FeatureFlagRepository extends Repository<FeatureFlag> {
...
     return result;
   }
+
+  public async validateUniqueKey(flagDTO: FeatureFlagValidation) {
+    const result = await this.createQueryBuilder('feature_flag')
+      .where('feature_flag.key = :key', { key: flagDTO.key })
+      .andWhere('feature_flag.context = :context', { context: flagDTO.context })
+      .getOne();
+
+    if (result) {
```
VivekFitkariwala, line 122: This part of the code should be in the service as it is not releated to data fetching

RidhamShah: code moved

VivekFitkariwala (dismissed):

zackcl:
@zackcl Test/review this

backend/packages/Upgrade/src/api/middlewares/ErrorHandlerMiddleware.ts
zackcl: It seems like this code is still here? @RidhamShah Shouldn't we delete or update this code as @danoswaltCL or @VivekFitkariwala suggested?

zackcl:
@RidhamShah Could we also make the Key field red when the key is duplicated to emphasize that it is invalid?

Screenshot 2024-09-24 at 6 26 34 PM

This is also expected behavior for other fields when they're invalid:

Screenshot 2024-09-24 at 6 44 09 PM

VivekFitkariwala (approved):

zackcl:
@RidhamShah I can confirm that the Key field's outline turns red when the field is invalid. However, I found a bug where if the user tries to update the Feature Flag's Name without updating the Key, we get a validation error in the modal.

Screenshot 2024-09-25 at 1 01 47 AM

This also happens when updating the Description. In this case, the Key field's outline is not red, which is strange.

Screenshot 2024-09-25 at 1 02 51 AM

Could you please make it possible to update a Feature Flag through the "Edit Feature Flag" modal while leaving the Key unchanged?

zackcl (dismissed):
Please fix the bug I mentioned in the comment.

RidhamShah:
> Please fix the bug I mentioned in the comment.

Can you check now, updated the code.

zackcl:
@RidhamShah I still can't update a feature flag while leaving the Key unchanged. Please see this video:

https://github.com/user-attachments/assets/3ed1a26a-fa11-44e1-a40d-57095449e7a7

We should allow updating a feature flag while leaving the Key unchanged (e.g., only update Name or Description).

Yagnik56:
@zackcl can you try again? Issue is fixed!

zackcl:
@Yagnik56 I can now update a feature flag while leaving the Key unchanged. However, the duplicated key error no longer appears on the modal, and the error snackbar is visible again. Can we fix this?

zackcl (approved):

PR #1787, opened 2024-07-26
Hide scrollbar for modal content, section card list, and tag input when not in use on Linux

zackcl:
This PR fixes the issue of the scrollbar always being displayed, even when not in use, on Linux.

@Yagnik56 @RidhamShah @ppratikcr7 I would appreciate it if anyone using Linux could confirm that this issue is resolved with this PR.


Yagnik56 (approved):
This looks good, the issue is resolved now.

Issue #1777, opened 2024-07-23
FF: Implement the Enable/Disable Slide Toggle for the "Include Lists" Table Rows


zackcl:
Clicking the slide toggle on the "Include/Exclude Lists" table rows should open the "Enable Include/Exclude List" modal to confirm the action. The message in the modal should include the corresponding list's name. Upon clicking the "Enable/Disable" button in the modal, the modal should close, and the corresponding row's slide toggle should be updated accordingly.

<img width="1536" alt="Feature Flag Details" src="https://github.com/user-attachments/assets/9dab3c47-57bc-4888-a380-946042fd89d1">

<img width="512" alt="Enable Include List" src="https://github.com/user-attachments/assets/82fef078-643a-45c7-8a5f-26a35828c2a4">

<img width="1536" alt="Feature Flag Details (1)" src="https://github.com/user-attachments/assets/f56fc077-a0df-48d1-839f-a91a6aa06b93">

<img width="512" alt="Disable Include List" src="https://github.com/user-attachments/assets/dfd3fb8a-cd2a-4002-ae71-97d261e93924">

Issue #1776, opened 2024-07-23
FF: Implement the Delete Button for the "Include/Exclude Lists" Table Rows


zackcl:
Clicking the "Delete" button on the "Include/Exclude Lists" table rows should open the "Delete Include/Exclude List" modal to confirm the deletion. The message in the modal should include the corresponding list's name. Upon clicking the "Delete" button in the modal, the modal should close, and the corresponding row should be deleted from the table.

<img width="1536" alt="Feature Flag Details" src="https://github.com/user-attachments/assets/3939fad1-3156-47e9-b15e-7049ec18c127">

<img width="512" alt="Delete Include List" src="https://github.com/user-attachments/assets/cfcfb97b-9374-40f3-a094-6d66211d0a6e">

Yagnik56:
QA passed

Issue #1775, opened 2024-07-23
FF: Implement the Edit Button for the "Include/Exclude Lists" Table Rows


zackcl:
Clicking the "Edit" button on the "Include/Exclude Lists" table rows should open the "Edit Include/Exclude List" modal, pre-filled with the row's field values. The "Save" button should remain disabled until the form is modified. Upon clicking the "Save" button after making edits, the modal should close, and the updated values should be reflected in the corresponding table row.

<img width="1536" alt="Feature Flag Details" src="https://github.com/user-attachments/assets/b918d4ba-29e3-4faa-ae84-5e45d7700ee5">

<img width="688" alt="Feature Flag - Edit Include List" src="https://github.com/user-attachments/assets/83351b82-4dbb-4da5-92bc-1f08eb77f750">

Yagnik56:
@zackcl @bcb37 In the update list modal for individual lists, I cannot remove the value tags. There are no errors, but the value is not removed from the list. This issue occurs with both the include and exclude lists for individual lists, while the group list functions correctly.

https://github.com/user-attachments/assets/925694a0-7808-4f04-986f-2b2db0ac05ae

Yagnik56:
QA passed

Issue #1757, opened 2024-07-12
FF Map backend response to Include/Exclude Tables


danoswaltCL:
The Include and Exclude table data design has changed a little and will need to be reimplemented, as well as the endpoint to gather data from.

The inclusion of the 'include all' will be in a separate story.

Each row will have an enable toggle and a couple buttons, but for this particular story we just want to make sure the row data for of each list type is showing up correctly and ready to be wired up.

@bcb37 can you attach an example of the data fetch response that we should be building this table(s) from now?

Latest design:

![Image](https://github.com/user-attachments/assets/ac8f86c0-59a1-4437-b7c8-40e67fcdab6b)

![Image](https://github.com/user-attachments/assets/b29163bd-b157-4418-84be-d985e25762d4)

PR #1793, opened 2024-07-30
Implement Include List Row Actions

zackcl:
This PR implements Enable/Disable/Edit/Delete actions for rows in the Include List table on the Feature Flag details page. Additionally, it fixes an issue where include/exclude lists were not being updated on PUT requests by defining a separate `updateList()` function for use by the PUT endpoints.

Specifically, this PR:
- Adds Enable/Disable functionality for Include List rows
- Implements Edit action for Include List rows
- Implements Delete action for Include List rows
- Fixes list update functionality for both include and exclude lists

This resolves the "Include List" part of the following issues:
- #1757
- #1775
- #1776
- #1777


danoswaltCL:
well the work is largely done, but if this _can_ be broken into smaller PRs that map to the stories on the board it would be easier to review for sure.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.ts
```diff
@@ -80,6 +82,13 @@ export class UpsertPrivateSegmentListModalComponent {
     this.fetchData();
     this.createPrivateSegmentListForm();
     this.initializeListeners();
+    this.populateFormForEdit();
+  }
+
+  ngAfterViewInit() {
+    if (this.config.params.action === UPSERT_PRIVATE_SEGMENT_LIST_ACTION.ADD_FLAG_INCLUDE_LIST) {
+      setTimeout(() => this.typeSelectRef.open(), 150);
```
danoswaltCL, line 90: so what's going on here?

zackcl: That's to make the "Type" dropdown is focused and opened as soon as the modal opens (like in the Figma design prototype I shared). It was the only solution that worked based on my attempts. The `setTimeout()` wasn't necessary, but I thought it would be better to add a slight delay.

frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -156,7 +159,7 @@ export class DialogService {
       data: commonModalConfig,
       width: '656px',
       height: 'auto',
-      autoFocus: 'input',
+      autoFocus: false,
```
danoswaltCL, line 162: what element does it focus on if you make this false? I think we want at least 'heading', but as long as it makes sense for keyboard tabbing.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.ts
danoswaltCL: what does the delay do though?

frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
zackcl: I don't fully understand how the `autoFocus` parameter works, but setting it to `false` disabled the auto-focus on any fields by default, which is what I intended. If we set it to 'input', it automatically focuses on the "Segment" or "Values" field, which I find a bit unnatural.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.ts
zackcl: It just delays displaying the "Type" dropdown options by 150ms. You might not notice much difference with or without the delay, though I find it a bit more natural to add a slight delay.

frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
danoswaltCL: yeah 'input' was probably wrong, I guess false is fine as long as the focus is in the modal on the container or the header and tabbable to the form, but 'first-heading' would be a good default option also.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.ts
danoswaltCL: Ah ok, when I see setTimeout my first thought is always that it's a band-aid to fix some other issue about rendering or change-detection, but this is for an intended UX effect, I got it.

setTimeout/setInterval can be a little dicey, especially in Angular. Some of the nastiest bugs I've had to unravel have dealt with setTimeout/setInterval. It schedules a task outside of the execution context that Angular knows about, so it can create subtle bugs in change-detection that are really hard to find, among other things like memory leaks.

Anyway, it's easy to replace, you can do a thing like this, I tested it, works the same. Internally this thing probably uses setTimeout also, but the side-effects are managed by rxjs and plays nicer with Angular.

Can you try this, and maybe just a comment in the code to explain that this is for a specific UX effect?

```
this.subscriptions.add(timer(150).subscribe(() => this.typeSelectRef.open())
 ```

File: frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component.ts [hunk 3]
```diff
@@ -118,6 +127,37 @@ export class UpsertPrivateSegmentListModalComponent {
...
+    }
+
+    const values = this.determineValues(sourceList.listType, sourceList.segment);
+    this.privateSegmentListForm.patchValue({
+      listType: sourceList.listType,
+      segment: sourceList.segment,
+      values,
+      name: sourceList.segment.name,
+      description: sourceList.segment.description,
+    });
```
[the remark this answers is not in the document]
zackcl: Thank you for the insightful explanation about reactive forms in Angular. I appreciate the detailed advice. I'll update the code to implement your suggestions tomorrow, focusing on minimizing unintended side effects in our form handling.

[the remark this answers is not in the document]
danoswaltCL: No other changes needed, all else looks good to me, just in this one spot to be safe, I don't see any other spots that should be an issue. Was a long way of saying we probably want to use `{ emitEvent: false })`.

danoswaltCL:
I don't really have any other suggestions, as long as backend `updateList` stuff is good with @bcb37 I approve

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -500,12 +500,12 @@ export class FeatureFlagsController {
     @Body({ validate: true }) exclusionList: FeatureFlagListValidator,
     @Req() request: AppRequest
   ): Promise<FeatureFlagSegmentExclusion> {
-    return this.featureFlagService.addList(exclusionList, 'exclusion', request.logger);
+    return this.featureFlagService.updateList(exclusionList, 'exclusion', request.logger);
```
bcb37, line 503: Should we add the test to verify the ids in the body and URL match, like you have for inclusion lists?

bcb37 (commented):
Just one question, otherwise this looks good.

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
zackcl: I was thinking that could be done in another PR for implementing the Exclude List table, but I can also apply that change now. Thanks!

zackcl:
I resolved all your comments. Please feel free to approve this PR.

bcb37 (approved):

Issue #1794, opened 2024-07-30
Add non-http error code enum strings - placeholder


amurphy-cl:
Placeholder for adding error codes to show in front end. 

Related issue: https://github.com/CarnegieLearningWeb/UpGrade/issues/1762

danoswaltCL:
I have discovered that `AppErrorHandler` service is a mirage that doesn't appear be doing anything, and the errors are actually all intercepted. I commented out the `AppErrorHandler` service in the app module, everything still seems to work fine.

That which complicates my plan... but I have a similar different potential solution that I think is also good will share tomorrow.

PR #1797, opened 2024-08-01
Error Event Bus (UI)

danoswaltCL:
This adds a service called Error Event bus. #1794

This is a new ngrx feature store that will have the sole purpose of setting known errors to a store that can be easily subbed to in components.

I can go over in a demo how this works, but the gist when you have an new error you want to flag for the UI to act upon:

1. Create 2 actions, one for `set` to dispatch the `NewErrorEvent`, and one for `clear` to dispatch its clearing action.
2. Add those actions to the `errorTypeToActionsMap` with the SERVER_ERROR enum type as the key.
3. Add those actions to the reducer and add the error by name to the error-event-bus state (also in the reducer).
4. Create a selector for your error. The value emitted will be falsey (`null`) or a truthy (`NewErrorEvent`), so it can be used with simple `if` statements as true or false (the error exists or it doesn't).
5. Set up a listener for your error in the component that will flag error logic if it is true.
6. To clear the error when it is handled, you will call the `clear` function on the `NewErrorEvent` that is emitted by the observable, i.e. `exampleError.clear()`

Technically this is for the Duplicate Key error.


danoswaltCL:
I can remove the "EXAMPLE_ERROR" part of this, was planning on demo'ing this but if anyone plans on reviewing this it is there for ease of seeing how an implementation would work. You can play around by forcing an http error of that type, and look in the store in the dev tools to see that error value on the `ErrorEventBus` change.

frontend/projects/upgrade/src/app/core/error-event-bus/store/error-event-bus.reducer.ts
```diff
@@ -0,0 +1,28 @@
+import { createReducer, on } from '@ngrx/store';
+import * as ErrorActions from './error-event-bus.actions';
+import { SERVER_ERROR } from 'upgrade_types';
+
+export interface ErrorState {
```
VivekFitkariwala, line 5: For each error, we will have a new attribute?

frontend/projects/upgrade/src/app/core/error-event-bus/store/error-event-bus.model.ts
```diff
@@ -0,0 +1,7 @@
+import { SERVER_ERROR } from 'upgrade_types';
+
+export interface NewErrorEvent {
+  type: SERVER_ERROR | string;
```
VivekFitkariwala, line 4: Do you need `string` here? Because `NewErrorEvent` is getting created from `errorTypeToActionsMap` where you will add only `SERVER_ERROR`.

frontend/projects/upgrade/src/app/core/error-event-bus/error-event-bus.service.ts
```diff
@@ -0,0 +1,61 @@
...
+import { NewErrorEvent } from './store/error-event-bus.model';
+import { errorTypeToActionsMap } from './store/error-event-bus.actions';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ErrorEventBus {
+  constructor(private store$: Store<AppState>, private notificationService: NotificationService) {}
+
+  handleError(error: any) {
```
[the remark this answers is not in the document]
danoswaltCL: I observed this during testing at some point, else I wouldn't have thought to check it. I would need to try and reproduce again to be sure though where that came from, or if it was just a funky test error.

frontend/projects/upgrade/src/app/core/error-event-bus/store/error-event-bus.model.ts
danoswaltCL: probably not, I'm not sure what my thought was there.

frontend/projects/upgrade/src/app/core/error-event-bus/store/error-event-bus.reducer.ts
danoswaltCL: Yeah, that's the idea. I had a different idea at first, which was basically to have just one error stream that all components could listen to on their own for errors of that type.

I think I ended up liking this better just so all the logic for creating and cleaning up errors could be kinda managed by this feature itself, minimal code for the component, which just needs to subscribe to that error type to catch it and use it like a boolean flag, and `clear` to mark that it was handled to remove it.

danoswaltCL:
@VivekFitkariwala @zackcl So, do we not want to do this? The reason for doing this this way was to avoid a lot of the kind of thing where we handle some http responses outside of ngrx and directly handle everything in the component, with promises because observables were deemed to complicated or something. I thought that was a wrong approach and that's what made me do this and hold up approving https://github.com/CarnegieLearningWeb/UpGrade/pull/1786.

But it seems that others don't really mind and actually prefer it like that such as we have talked about in https://github.com/CarnegieLearningWeb/UpGrade/pull/1798... maybe we have some later tech debt to handle local component state different with signals and or some thing like ngrx-component store....so if we're approving that PR, we might was well also approve https://github.com/CarnegieLearningWeb/UpGrade/pull/1786 and ditch this idea, I'm tired of fighting it and I don't have the capacity now to keep up the feature flag code reviews right now.

VivekFitkariwala:
I have not approved this PR because we are centralising the error capture and removing the context information from where it was generated. So in cases where multiple errors of the same kind occur on the front end, they will replace each other. Another problem is the reset which needs to be done by the consuming component itself. What if the user forgets to handle it or navigates away from that component itself? 
For these reasons, I would prefer that instead of accumulating errors in the central store, that should be consumed immediately.

PR #1799, opened 2024-08-05
Prevent App Freezing in Metrics Component with Large Datasets

### LINKED ISSUE #1796: Rendering of nested profile metrics runs out of memory
Discovered while researching [this issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/1703):

On the profile page, the rendering of the nested tree of metrics data runs out of memory when there is a top-level metric with thousands of children. This happens in all three deployed environments in the case of the 'masteryWorkspace' and 'conceptBuilderWorkspace' metrics.

This affects all deployed versions:

To reproduce on dev (v5.2): log into UpGrade > navigate to the profile page.
To reproduce on other environments (v5.1): log into UpGrade > navigate to the Metrics tab of the profile page > click on the '>' to expand the row for the 'masteryWorkspace' and 'conceptBuilderWorkspace' metrics.

Deliverable: 
Implement a fix that handles the volume of metrics data somehow without running out of memory (possibly by truncating the metric children data at some cutoff, or disabling the recursive tree display in some or all cases).

If there's no solution the allows the current design to handle the volume of data, we should limit the data being rendered for the current release (6.0), and then re-think the design. In that case, spawn an issue.

zackcl:
This PR (Resolves #1796) addresses a critical issue where the Metrics component causes the app to freeze when rendering large metrics datasets up front.

The PR implements a lazy loading mechanism for the metrics tree structure, which:

- Prepares metrics data for on-demand loading
- Supports asynchronous child node loading
- Renders child nodes only when a parent node is expanded
- Improves rendering performance by removing `tree-invisible` class usage ([reference](https://stackoverflow.com/questions/55965653/angular-mat-tree-re-render-performance-issue-takes-too-much-time))


frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
```diff
@@ -52,10 +52,10 @@
           <td class="ft-12-600" mat-cell *matCellDef="let element; let index = index">
             <ng-container *ngIf="element">
               <mat-tree [dataSource]="[element]" [treeControl]="nestedTreeControl" class="mat-nested-tree">
-                <!-- This is the tree node template for leaf nodes -->
+                <!-- Leaf node template -->
                 <mat-tree-node *matTreeNodeDef="let node">
-                  <button mat-icon-button>
-                    <mat-icon class="leaf-node-button-icon"> fiber_manual_record</mat-icon>
+                  <button mat-icon-button disabled>
```
bcb37, line 57: Does this have to be an always-disabled button? If it were just the plain icon it wouldn't be greyed-out, I think.

bcb37 (commented):
This looks good Just one question.
Also, search by 'All' or 'Context' throws an error on lines 168 and 174 of `metrics.component.ts` in cases where there's no context. It could be fixed by coalescing like : `!!data.context?.filter(...`

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
zackcl: The "disabled" state is added mainly to remove the hover effect (ripple) as it indicates the bullet is interactive when it's not. However, I think it makes sense to make them all black-colored for consistency as you suggested. I will apply this change.

zackcl:
@bcb37 I resolved all your comments. Please feel free to approve this PR.

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
bcb37: Just to be clear: I was asking why it needed to be a button at all. Can't it just be the icon, without having the <button> tags around it?

bcb37 (commented):
Just a clarification.

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
zackcl: Understood, thanks for clarifying. The button tags were already in place before I made any changes, so I'm not entirely sure of the original reason for using them. I'll go ahead and remove the button tags and keep just the icon.

zackcl:
@bcb37 I have applied the suggested change. Please feel free to approve the PR.

bcb37 (dismissed):

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
```diff
@@ -21,6 +21,11 @@ import { METRIC_SEARCH_KEY } from '../../../../../../../../../../types/src/Exper
 import { IMetricUnit } from '../../../../../../../../../../types/src';
 import { DeleteComponent } from '../../../../../shared/components/delete/delete.component';
 
+// Extend IMetricUnit to include a function for lazy loading children
+type LazyLoadingMetric = IMetricUnit & {
```
danoswaltCL, line 25: interesting! should go in the metrics.model.ts

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
```diff
@@ -36,60 +41,83 @@ export class MetricsComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  // Process a single metric node, setting up lazy loading for its children
+  private insertNode(metric: IMetricUnit): LazyLoadingMetric {
+    const processedMetric: LazyLoadingMetric = {
+      ...metric,
+      children: [],
     };
-    return metrics;
+
+    // If the node has children, create a loadChildren function instead of processing them immediately
+    if (metric.children && metric.children.length > 0) {
```
danoswaltCL, line 92: We can do this a little easier if we can stick to observables instead of promises, if I'm not mistaken this should work....

if (metric.children && metric.children.length > 0) {
      processedMetric.loadChildren = () => of(metric.children?.map((child) => this.insertNode(child)));
    }

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
```diff
@@ -36,60 +41,83 @@ export class MetricsComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      processedMetric.loadChildren = () => Promise.resolve(metric.children?.map((child) => this.insertNode(child)));
+    }
+
+    return processedMetric;
+  }
+
+  // Custom method to get children for a node, supporting lazy loading
+  private getChildren = (node: LazyLoadingMetric): Observable<LazyLoadingMetric[]> => {
```
danoswaltCL, line 100: then this can be smooshed down to:

```
private getChildren = (node: LazyLoadingMetric): Observable<LazyLoadingMetric[]> => {
    if (node.loadChildren) {
      return node.loadChildren().pipe(tap((children) => (node.children = children)));
    }
    return of(node.children || []);
  };
  ```

frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
danoswaltCL: (and change the loadChildren type to `Observable` instead of `Promise`

danoswaltCL:
looks good to me too, just one request to avoid the awkward conversion from promise to an observable. Please try that out, I tested it quickly and it worked on the smaller metric dataset that I have loaded in, or seemed to.

zackcl:
@danoswaltCL I have resolved all your comments and updated the `deleteNode` function, which was not working previously. I've tested it with large metrics datasets, and everything seems to work fine. Please feel free to approve this PR.

zackcl:
I will review this one more time.

danoswaltCL:
I'd approve, looks to be back to draft though, did you find an issue?

zackcl:
I realized I didn’t need to update the `deleteNode` and `findParents` functions to make the deletion work. Instead, I needed to update the `insertNode` function to include the `id` again, which I had previously removed. I will discard some of my later commits and redo them properly.

zackcl:
@danoswaltCL Sorry for the delay, the PR is ready now. After resolving your comments (https://github.com/CarnegieLearningWeb/UpGrade/pull/1799/commits/e6e82fb3f95cbef912c0d78d72a04101c7d6a8f7), I made one more commit (https://github.com/CarnegieLearningWeb/UpGrade/pull/1799/commits/0db0de5f770a1f98ebb6104c45438493a9bd0e42) to make the node deletion work properly. Please feel free to approve this PR.

danoswaltCL (approved):

PR #1803, opened 2024-08-06
Update getAllExperimentConditions to respond 404 when /assign is requested without /init

### LINKED ISSUE #1801: Requesting api/v5/assign without api/v5/init should return a proper 404 response
5.2.0

**Describe the bug**
When `api/v5/assign` is requested without `api/v5/init` first, it throws `TypeError: Cannot read properties of null (reading 'requestedUserId')` in the backend console. However, the proper response should be "Experiment user not defined (404)" as written in the [clientlib documentation](https://github.com/CarnegieLearningWeb/UpGrade/blob/dev/clientlibs/js/README.md).

**To Reproduce**
1. Make a request to `api/v5/assign` without `api/v5/init` first.
2. Check the backend console.
3. You will see the `TypeError: Cannot read properties of null (reading 'requestedUserId')` in the console.

**Expected behavior**
Requesting api/v5/assign without api/v5/init should return a proper 404 response.

zackcl:
Resolves #1801


bcb37 (dismissed):
This works, but we've got the same problem with /log, /mark, and /alias. I'll create issues.

danoswaltCL (approved):

PR #1805, opened 2024-08-06
Update getKeys to respond 404 when /featureflag is requested without /init

### LINKED ISSUE #1802: Requesting /featureflag without /init should return a proper 404 response
5.2.0

**Describe the bug**
When `/featureflag` is requested without `/init` first, it throws `TypeError: Cannot read properties of null (reading 'id')` in the backend console. However, the proper response should be "Experiment user not defined (404)" to be consistent with how `/assign` works ([clientlib documentation](https://github.com/CarnegieLearningWeb/UpGrade/blob/dev/clientlibs/js/README.md)).

**To Reproduce**
1. Make a request to `api/v5/featureflag` without `api/v5/init` first.
2. Check the backend console.
3. You will see the `TypeError: Cannot read properties of null (reading 'id')` in the console.

**Expected behavior**
Requesting `/featureflag` without `/init` should return a proper 404 response.

zackcl:
Resolves #1802


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -39,7 +39,21 @@ export class FeatureFlagService {
   }
 
   public async getKeys(experimentUserDoc: ExperimentUser, context: string, logger: UpgradeLogger): Promise<string[]> {
-    logger.info({ message: 'Get all feature flags' });
+    logger.info({ message: `Get all feature flag keys: User: ${experimentUserDoc?.id}` });
+
+    // throw error if user not defined
+    if (!experimentUserDoc || !experimentUserDoc.id) {
+      logger.error({ message: `User not defined in getAllExperimentConditions: ${experimentUserDoc?.id}` });
```
bcb37, line 46: `experimentUserDoc?.id` in that interpolated string will never be anything but undefined. We should probably just throw a more generic message. (same on line 50 below)

bcb37 (changes requested):
One suggestion.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
zackcl: Thanks, I've updated the code so that it logs with `experimentUserDoc?.requestedUserId` instead of `experimentUserDoc?.id` which will always be undefined. This is exactly the same as how we log and return 404 in the `getAllExperimentConditions` function.

However, in the current code, the `experimentUserDoc?.requestedUserId` is also `undefined` when the user document is not found. This is due to how the `getUserDoc` function is currently implemented:
```
public async getUserDoc(experimentUserId, logger): Promise<RequestedExperimentUser> {
    try {
      const experimentUserDoc = await this.getOriginalUserDoc(experimentUserId, logger);
      if (experimentUserDoc) {
        const userDoc = { ...experimentUserDoc, requestedUserId: experimentUserId };
        logger.info({ message: 'Got the user doc', details: userDoc });
        return userDoc;
      } else {
        return null;
      }
    } catch (error) {
      logger.error({ message: `Error in getting user doc for user => ${experimentUserId}`, error });
      return null;
    }
  }
```
I believe this should be updated to the following to allow logging `experimentUserDoc?.requestedUserId` when the user document is not found:
```
public async getUserDoc(experimentUserId: string, logger: UpgradeLogger): Promise<RequestedExperimentUser | null> {
    try {
      const experimentUserDoc = await this.getOriginalUserDoc(experimentUserId, logger);
      const userDoc = { ...experimentUserDoc, requestedUserId: experimentUserId };
      logger.info({ message: 'Got the user doc', details: userDoc });
      return userDoc;
    } catch (error) {
      logger.error({ message: `Error in getting user doc for user => ${experimentUserId}`, error });
      return null;
    }
  }
```

I can create a separate PR to apply this change. Please feel free to approve this one if this makes sense.

zackcl: I created a PR to update the `getUserDoc` function : https://github.com/CarnegieLearningWeb/UpGrade/pull/1809

danoswaltCL (approved):

PR #1806, opened 2024-08-06
major version bump for feature flags

danoswaltCL:
Now that we have a release v5.2 branch, we should make sure dev is on the real next version `6.0`.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "ab_testing_backend",
-    "version": "5.2.0",
+    "version": "6.0.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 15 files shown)


bcb37:
> Now that we have a release v5.2 branch, we should make sure dev is on the real next version `6.0`.

It's going to be tricky to merge version 5.2 down to dev once it's tested and deployed. One set of conflicts we can avoid when that time comes is the version numbers (which will be overwritten on merge and have to be reverted anyway). It might be a good idea to hold off on this until after 5.2 is done.

danoswaltCL:
Okay I'll sit on my hands. It's the normal 'git-flow' thing to do after cutting a release branch, those package.json merge conflicts are just a ritual of the process, but as long as someone has a plan that's all I care.

PR #1808, opened 2024-08-06
filter out disabled lists from consideration

bcb37:
When applying the logic to return a list of feature flags for a given user, remove inclusion and exclusion lists that have `enabled` set to 'false' from the data.


zackcl:
The change looks good. I tested this change on top of the [feature/feature-flag-include-all-toggle](https://github.com/CarnegieLearningWeb/UpGrade/tree/feature/feature-flag-include-all-toggle) branch and everything seems to work as expected.

zackcl (approved):

PR #1106, opened 2023-11-08
Optimized init call

RidhamShah:

PR #1809, opened 2024-08-07
Update getUserDoc to return requestedUserId even when userDoc is not found

zackcl:
This PR resolves the issue mentioned in the PR: https://github.com/CarnegieLearningWeb/UpGrade/pull/1805

Currently, we use `requestedUserId` to log the requested user ID in the functions like `getAllExperimentConditions` and `markExperimentPoint` when the user document was not found. However, currently, the `getUserDoc` function returns `null` when the user is not found, making the `requestedUserId` to be undefined. This PR updates the `getUserDoc` function to still return the `requestedUserId` so that the requested user ID can be logged when the user was not found.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
@@ -295,16 +295,12 @@ export class ExperimentUserService {
     return this.userRepository.save(newDocument);
   }
 
-  public async getUserDoc(experimentUserId, logger): Promise<RequestedExperimentUser> {
+  public async getUserDoc(experimentUserId: string, logger: UpgradeLogger): Promise<RequestedExperimentUser | null> {
     try {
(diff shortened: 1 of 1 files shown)


bcb37 (commented):
Now that `getUserDoc()` cannot return null, the returned value will always be truthy. Because of this, tests like the one on line 149 of `FeatureFlagController.ts` will always be false. That test may now be redundant, but there are 9 other such tests in `ExperimentClientController.v5.ts` and a few others elsewhere as well, I think. It might work to change all those tests to look for the `id` but we'd want to be exhaustive.

zackcl:
@bcb37 That's a good observation. Maybe we should keep the `getUserDoc` function unchanged to avoid unexpected behavior? (even though the `requestedUserId` will be `undefined` in some logs) 

I can still try to fix the tests you mentioned though. I'm curious to hear what @RidhamShah thinks about this, as he seems to have worked on some relevant code (PR #1106, #1536).

PR #1823, opened 2024-08-09
parse failed message added in common import container

Yagnik56:
This PR contains the code change of adding a parse fail error message in a common import container.
(JSON is used to example only)
image


danoswaltCL (approved):

zackcl:
@Yagnik56 Just to clarify, can the "Invalid JSON" error message show up when importing an invalid JSON file? Shouldn't it be immediately converted to the compatibility table (with a single "Incompatible" row) if the JSON file is invalid (when importing a feature flag)?

I'm asking because my PR (https://github.com/CarnegieLearningWeb/UpGrade/pull/1824) will remove the code changes you made with this PR. In my PR, it only shows the invalid CSV error message when the user uploads an invalid CSV file.

Yagnik56:
@zackcl we have to pass observable (optional) as true to show that invalid message and it is dynamic so it will read the file type and show there. I just used JSON as an example, it works the same way for CSV. For JSON we won't pass observable value so it won't show that error

zackcl:
I'm not sure what you mean by "pass observable (optional) as true" but my PR basically does the same thing as what you did to show the invalid error message (I've updated to "Import failed. Please check your file and try again.") when the input variable `importFailed` is set to `true`.

```
    <span *ngIf="importFailed" class="ft-12-400 import-failed-msg">
      {{ 'feature-flags.upsert-list-modal.import-csv.error.message.text' | translate }}
    </span>
    ```
   I just wanted to clarify if removing your code changes won't affect your other code for importing JSON file, and it seems it doesn't affect it since we never show it for JSON. Thanks!

**Comment by Yagnik56** [comment]
Rather than putting hard coded messages keep it dynamic for fileType, it will be better in my view. @zackcl

**Comment by zackcl** [comment]
I think it's fine to keep it static for now since it's never used for JSON. Also, the message is actually fine to be used for both types:
```
"feature-flags.upsert-list-modal.import-csv.error.message.text": "Import failed. Please check your file and try again.",
```

Also, I think we want to maintain such messages in the `en.json` file for best practice. If you have any specific concerns about the changes that will be made with my PR regarding this, please feel free to leave comments there (https://github.com/CarnegieLearningWeb/UpGrade/pull/1824). Thanks!

PR #1825, opened 2024-08-12
FeatureFlag import API

RidhamShah:
In the I implemented import API for FeatureFlags, the API is designed to create a FeatureFlag using its provided details and then add list details for both inclusion and exclusion by creating their respective private segments.

The backend implementation of this API is complete. However, during testing, several UI issues were identified:

- The FeatureFlag only refreshes upon a hard refresh; it does not update when switching tabs and returning to the FeatureFlag. As a result, flags imported through Postman are not displayed.
- When importing a FeatureFlag, the import list is created and sent to the frontend UI. However, due to the importAll toggle being enabled, it does not appear in the view until the toggle is turned off.
- Upon importing a FeatureFlag, the export list is created and sent to the frontend UI, but it is not displayed in the view.
- Necessary frontend modifications are still pending to ensure that only non-INCOMPATIBLE FeatureFlags are sent to the import API.


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -424,6 +430,68 @@ export class FeatureFlagService {
...
+      .filter((fileStatus) => fileStatus.error === null)
+      .map((fileStatus) => {
+        const featureFlagFile = featureFlagFiles.find((file) => file.fileName === fileStatus.fileName);
+        return JSON.parse(featureFlagFile.fileContent as string);
+      });
+
+    const createdFlags = await Promise.all(
+      validFiles.map(async (featureFlag) => {
+        const { featureFlagSegmentInclusion, featureFlagSegmentExclusion, ...rest } = featureFlag;
+        const newFlag = await this.addFeatureFlagInDB(rest as any, logger);
```
VivekFitkariwala, line 460: Use transaction.

VivekFitkariwala, line 460: Remove `as any` from here and make `featureFlagSegmentInclusion` and `featureFlagSegmentExclusion` optional in the types.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -424,6 +430,68 @@ export class FeatureFlagService {
...
+
+        const inclusionDoc = await Promise.all(
+          featureFlagSegmentInclusion.map(async (segmentInclusion) => {
```
VivekFitkariwala, line 463: The transaction should be used here.
Instead of calling `this.addList` for each inclusion list, we should use a single call to add to the database.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -490,31 +558,31 @@ export class FeatureFlagService {
...
-        ),
-      ];
-
-      const segments = await this.segmentService.getSegmentByIds(segmentIds);
-      segments.forEach((segment) => {
-        if (segment == undefined) {
-          compatibilityType = FF_COMPATIBILITY_TYPE.WARNING;
-        }
-      });
+      const keyExists = await this.featureFlagRepository.findOne({ key: flag.key });
```
VivekFitkariwala, line 565: This change is from another PR. Right?

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code added

RidhamShah: code updated

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -445,6 +482,74 @@ export class FeatureFlagService {
...
+    });
+
+    const validFiles: FeatureFlagValidation[] = fileStatusArray
+      .filter((fileStatus) => fileStatus.error === null)
+      .map((fileStatus) => {
+        const featureFlagFile = featureFlagFiles.find((file) => file.fileName === fileStatus.fileName);
+        return JSON.parse(featureFlagFile.fileContent as string);
+      });
+
+    const createdFlags = await getConnection().transaction(async (transactionalEntityManager) => {
```
VivekFitkariwala, line 509: Using it like this means the import will happen in a single transaction. If we encounter a problem with any single import everything will fail. I am not sure if this is the behaviour we want.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -445,6 +482,74 @@ export class FeatureFlagService {
...
+          );
+
+          const featureFlagSegmentExclusionList = featureFlag.featureFlagSegmentExclusion.map(
+            (segmentExclusionList) => {
+              segmentExclusionList.list.id = uuid();
+              segmentExclusionList.flagId = newFlag.id;
+              return segmentExclusionList;
+            }
+          );
+          const exclusionDoc = await await this.addList(
```
VivekFitkariwala, line 539: 2 await

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -445,6 +482,74 @@ export class FeatureFlagService {
...
+          );
+
+          const featureFlagSegmentInclusionList = featureFlag.featureFlagSegmentInclusion.map(
+            (segmentInclusionList) => {
+              segmentInclusionList.list.id = uuid();
+              segmentInclusionList.flagId = newFlag.id;
+              return segmentInclusionList;
+            }
+          );
+          const inclusionDoc = await this.addList(
```
VivekFitkariwala, line 525: inclusionDoc and exclusionDoc be added in a `Promise.all`

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
VivekFitkariwala: Instead of doing `findOne` request for each flag, can we move the fetching in the calling function (find In array query) and send the result as the argument of `validateImportFeatureFlag`?

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: extra await removed

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

VivekFitkariwala (approved):

PR #1829, opened 2024-08-14
Add support for service account access tokens

### LINKED ISSUE #1752: Bypass Google Authentication for '/experiments' and '/metric' Endpoints
The `api/experiments` and `api/metric/save` endpoints are being requested by the [demo app](https://github.com/CarnegieLearningWeb/upgrade-demo-app) and the [tester app](https://github.com/CarnegieLearningWeb/upgrade-tester), but they currently fail with a `401 (Unauthorized)` error due to lack of Google authentication in the request.

I believe these endpoints are useful for testing UpGrade, and requiring Google authentication makes it difficult to use them. If we are concerned about the security, I think we can require different authentication methods, like an API key for developers.

The `@Authorized()` decorator for these endpoints should be commented out (search for `@JsonController('/experiments')` and `@JsonController('/metric')` in the codebase) to bypass the Google authentication until we implement a different authentication method.

**Describe alternatives you've considered**
Implement different authentication methods, like an API key for developers.

zackcl:
Resolves #1752

### Description
This PR modifies the `AuthService` to add support for validating service account access tokens in addition to the existing ID token validation. This change allows our application to authenticate requests made using service account credentials.

### Changes Made:
1. Updated the `validateUser` method in `AuthService` to handle both ID tokens and access tokens.
2. Implemented a two-step validation process:
   - First, attempt to validate the token as an ID token.
   - If that fails, attempt to validate it as an access token ([reference](https://cloud.google.com/nodejs/docs/reference/google-auth-library/latest/google-auth-library/oauth2client#google_auth_library_OAuth2Client_getTokenInfo_member_1_))

3. For valid service account access tokens, the method now returns `null` instead of throwing an error.
4. Maintained existing functionality for ID tokens to ensure backwards compatibility.
5. Added appropriate error handling and logging for both token types.

### Testing:
- Tested with both ID tokens and service account access tokens.
- Verified that existing ID token functionality works as before.
- Confirmed that service account access tokens are correctly validated.

### Future Considerations:
We might want to implement specific handling for service accounts in the future, such as creating a custom payload (email, name, role) or applying specific authorization rules.

### Reference:
For your reference, here's an example of code to generate an access token:

```
const { GoogleAuth } = require('google-auth-library');

async function authenticateServiceAccount() {
  try {
    const auth = new GoogleAuth({
      keyFilename: 'ppl-ees-603b3a01e004.json', // Path to your service account key file
      scopes: 'https://www.googleapis.com/auth/cloud-platform', // Specify the scopes your application needs
    });

    const client = await auth.getClient();
    const accessToken = await client.getAccessToken();
    
    console.log('Access Token:', accessToken); // Use this access token to make authenticated API requests
    return accessToken;
  } catch (error) {
    console.error('Error authenticating service account:', error);
    throw error;
  }
}

authenticateServiceAccount().catch(console.error);
```


danoswaltCL:
@kcalvo-cli maybe we can look at this in SRE mtg, I thought I created a devops ticket but I can't find it. This would allow us to get a proper automated-test service account credential for our QA engineer so that they can hit authenticated UI endpoints.

Since this is a different kind of user, I do wonder what else we have to be mindful of to make sure this is secure. If we could make sure this account's access is only good in the testing environments and has a short expiry, that may suffice?

danoswaltCL (approved):

PR #1845, opened 2024-08-21
Centralize demo mode DB clearing logic in ExperimentUserService

zackcl:
This PR centralizes the logic for clearing the database (`/clearDB`) in demo mode within the ExperimentUserService. 

Key changes:
- Moved the demo mode check from `ExperimentClientController` to `ExperimentUserService`
- Updated `clearDB` method in `ExperimentUserService` to include demo mode check
- Simplified `/clearDB` endpoints in all versions of `ExperimentClientController`

Benefits:
- Improves code maintainability by centralizing demo mode logic
- Reduces risk of inconsistencies across different controller versions

Testing:
- Verified that `clearDB` functionality works as expected in demo mode
- Confirmed that `clearDB` is properly restricted when not in demo mode

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentClientController.ts
@@ -6,7 +6,7 @@ import {
   Req,
   InternalServerError,
   Delete,
-  OnUndefined,
+  Authorized,
 } from 'routing-controllers';
 import { ExperimentService } from '../services/ExperimentService';
 import { ExperimentAssignmentService } from '../services/ExperimentAssignmentService';
@@ -1098,8 +1098,8 @@ export class ExperimentClientController {
    *          '500':
    *            description: DEMO mode is disabled
    */
+  @Authorized()
   @Delete('clearDB')
-  @OnUndefined(204)
   public async clearDB(@Req() request: AppRequest): Promise<string> {
     return this.experimentUserService.clearDB(request.logger);
   }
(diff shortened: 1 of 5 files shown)


zackcl:
@danoswaltCL I've added the `@Authorized()` decorator to the `/clearDB` endpoint for extra security as you suggested. I've fully tested the endpoint in different versions (v1, v4, v5) locally, and it works as expected. Please feel free to approve this PR.

danoswaltCL:
wait... this endpoint shouldn't be on the client controller, that makes no sense, especially now if using an authenticated endpoint middleware that uses a different auth than the client middleware uses. clients don't call this, it's an admin endpoint that should require a google login user credential.

Maybe can we copy paste this endpoint over to some other spot like the `SettingsService` and delete if from all client controllers? Kinda feels like that would be a better fit (if we really have to have this endpoint).

@VivekFitkariwala @bcb37 @ppratikcr7 any thoughts? this came about from me asking Zack if we really needed this endpoint, it seems like we're flirting with danger

bcb37:
> wait... this endpoint shouldn't be on the client controller, that makes no sense, especially now if using an authenticated endpoint middleware that uses a different auth than the client middleware uses. clients don't call this, it's an admin endpoint that should require a google login user credential.
> 
> Maybe can we copy paste this endpoint over to some other spot like the `SettingsService` and delete if from all client controllers? Kinda feels like that would be a better fit (if we really have to have this endpoint).
> 
> @VivekFitkariwala @bcb37 @ppratikcr7 any thoughts? this came about from me asking Zack if we really needed this endpoint, it seems like we're flirting with danger

It didn't feel like a huge change. It's technically a bit safer because the Authorization decorator is requiring the same authorization as the non-client endpoints. It's an odd endpoint so it's not obvious where it belongs. Maybe it should have its own controller? But if we're going to revisit the need for this eventually anyway, I was ok with leaving it in the client controller with Authorization required. It's necessary for the demo app to function without a lot of re-engineering,

VivekFitkariwala (approved):

VivekFitkariwala:
As @bcb37 said this is an odd endpoint and it is present because we don't have a multitenant system. So I don't have a strong opinion. We must be sure that the DEMO flag is false in production :)

Issue #1780, opened 2024-07-24
FF: Disable or Hide the Menu Options for v2


zackcl:
The menu options that we decided to implement for v2 on the Feature Flags Root and Details pages should be either disabled or hidden in the UI.

<img width="1536" alt="Feature Flags Root menu options for v1" src="https://github.com/user-attachments/assets/671e8630-238e-49d3-a8b6-2ec8cd1ada1e">

<img width="1536" alt="Feature Flags Details page menu options for v1" src="https://github.com/user-attachments/assets/7e3fcce5-f006-4158-8957-a8ff7a0e67e5">

danoswaltCL:
@zackcl can you list out exactly what all should be disabled / hidden for MVP1? cc: @bcb37

Yagnik56:
@zackcl we are going to hide the disabled options right? from menu

zackcl:
@Yagnik56 Yes, that's correct! We should hide them.

zackcl:
QA: Passed

PR #1850, opened 2024-08-22
Removed V2 UI elements

Yagnik56:
hide unused menu items on ff details page and hamburger menu for include and exclude action buttons


zackcl:
I think the "Email Feature Flag Data" menu option should also be hidden. See: https://github.com/CarnegieLearningWeb/UpGrade/issues/1780

zackcl (changes requested):
I think the "Email Feature Flag Data" menu option should also be hidden. See: https://github.com/CarnegieLearningWeb/UpGrade/issues/1780

zackcl (approved):

PR #1858, opened 2024-08-22
Upgrade Rebuild on ECS

shpwe:
* Create cloudformation template for resources
* Create Pipeline Configs
* Add string replacement function for environment.*.ts files
* Update package-lock.json to include replace-in-file package.
* Create new Dockerfile for cl builds.


frontend/angular.json
```diff
@@ -80,6 +80,47 @@
                 }
               ]
             },
+            "prod": {
```
danoswaltCL, line 83: there's already a `prod` build definition above that looks the same as this, but labeled `production`. The values all look the same otherwise

frontend/projects/upgrade/src/environments/environment.prod.ts
```diff
@@ -1,11 +1,11 @@
 export const environment = {
```
danoswaltCL, line 1: we currently use this file for the all the beanstalk env builds... the `apiBaseUrl` and `googleClientId` will be safely swapped out, but `baseHrefPrefix` will cause an issue because it won't be replaced when running `build:prod`, so all the app routes will append '%BASE_HREF_PREFIX%'.

maybe a separate environment `environment.beanstalk.prod.ts` and configuration called `beanstalk` to keep these separate while we switch over?  then we can retain the current build process without the replacements, and point `build:prod` at that environment? `"build:prod": "ng build --configuration beanstalk",

frontend/angular.json
shpwe: Our build CI expects environment names to be prod,qa,staging. If its not an issue, I can delete the `production` config since it wouldnt be used.

frontend/projects/upgrade/src/environments/environment.prod.ts
shpwe: having a seperate config for beanstalk works. WIll update this PR with it.

frontend/settings.env.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  API_BASE_URL: process.env.API_BASE_URL || ' ',
+  BASE_HREF_PREFIX: process.env.BASE_HREF_PREFIX || ' ',
+  GOOGLE_CLIEND_ID: process.env.GOOGLE_CLIEND_ID || ' ',
```
danoswaltCL, line 4: mispelling as CLIEND, should be CLIENT

danoswaltCL (approved):

PR #1862, opened 2024-08-23
Add default FetchHttpClient to lite versions for identical usage as non-lite version

zackcl:
## Context
This PR builds upon the work in #1565, which addresses #150 for importing clientlib without Webpack. The existing solution requires users to provide a custom HTTP client for the `browser-lite` version (as noted in [this comment](https://github.com/CarnegieLearningWeb/UpGrade/pull/1565#issuecomment-2302061297)). While this approach offers flexibility and further size optimization, I wanted to explore an alternative that might strike a different balance between ease of use and bundle size. 

This exploration aims to enable identical usage across all versions while maintaining a relatively small size for the lite version. I'm entirely open to closing this PR if the team prefers to maintain the current implementation.

## Summary
This PR adds a `FetchHttpClient` as the default for lite versions, enabling identical usage to non-lite versions without requiring a custom HTTP client. This change slightly increases bundle size (around 1KB) but significantly improves out-of-the-box functionality.

## Key Changes
- Implemented `FetchHttpClient` using native `fetch` API
- Set `FetchHttpClient` as default for lite versions (Custom HTTP clients can still be provided if needed)
- Updated webpack configuration and `ApiService` accordingly

## Benefits
- Identical usage between lite and non-lite versions
- No need to provide custom HTTP client for lite versions
- Maintained API consistency across all versions

## Trade-offs
Bundle size changes:
- Browser-lite: 41KB -> 42KB (+1KB, ~2.4% increase)
- Browser: 71KB -> 72KB (+1KB, ~1.4% increase)

The minor size increase is offset by improved user experience and ease of use.

## Usage
Both versions now work identically out-of-the-box:

```javascript
const upClient = new UpgradeClient(userId, hostUrl, context);
await upClient.init();
const assignment = await upClient.getDecisionPointAssignment(site, target);
```

### DIFF
--- a/clientlibs/js/jest.config.ts
+++ b/clientlibs/js/jest.config.ts
@@ -11,7 +11,7 @@ const config: Config.InitialOptions = {
       isolatedModules: true,
       diagnostics: false,
     },
-    USE_CUSTOM_HTTP_CLIENT: true,
+    USE_FETCH_DEFAULT_HTTP_CLIENT: true,
     IS_BROWSER: true,
     API_VERSION: 5,
   },
--- a/clientlibs/js/package.json
+++ b/clientlibs/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "upgrade_client_lib",
-  "version": "6.0.0",
+  "version": "5.2.0",
   "description": "Client library to communicate with the Upgrade server",
   "files": [
     "dist/*"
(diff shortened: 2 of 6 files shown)


zackcl:
Closing this as `fetch` isn't supported in older versions of Node.js.

PR #1865, opened 2024-08-26
implement FF exposures for v1

bcb37:
- Add an exposure record in the exposures table when a user first sees a given feature flag
- Display the number of unique user exposures associated with a given feature flag on the list page 'Exposures' column


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -64,6 +65,21 @@ export class FeatureFlagService {
 
     const includedFeatureFlags = await this.featureFlagLevelInclusionExclusion(filteredFeatureFlags, experimentUserDoc);
 
+    // save exposures in db
+    try {
+      const exposureRepo = getRepository(FeatureFlagExposure);
+
+      await Promise.all(
```
VivekFitkariwala, line 72: Instead of using a `map` which will do multiple DB requests, you should use an array to save all the documents in a single request.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -64,6 +65,21 @@ export class FeatureFlagService {
...
+        includedFeatureFlags.map((flag) => {
+          return exposureRepo.save({ featureFlag: flag, experimentUser: experimentUserDoc });
+        })
+      );
+    } catch (err) {
+      const error = new Error(`Error in saving feature flag exposure records ${err}`);
+      (error as any).type = SERVER_ERROR.QUERY_FAILED;
+      throw error;
```
VivekFitkariwala, line 80: Should we throw an error here or log the error and complete the request?
Saving the exposure is not a primary action, this HTTP request should succeed.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -107,7 +123,9 @@ export class FeatureFlagService {
   ): Promise<FeatureFlag[]> {
     logger.info({ message: 'Find paginated Feature flags' });
 
-    let queryBuilder = this.featureFlagRepository.createQueryBuilder('feature_flag');
+    let queryBuilder = this.featureFlagRepository
+      .createQueryBuilder('feature_flag')
+      .leftJoinAndSelect('feature_flag.featureFlagExposures', 'featureFlagExposures');
```
VivekFitkariwala, line 128: This is too much data to query in production. You are only interested in the count of the exposure. Have a separate query to get that count.

backend/packages/Upgrade/src/api/controllers/FeatureFlagController.ts
```diff
@@ -263,9 +246,10 @@ export class FeatureFlagsController {
       ),
       this.featureFlagService.getTotalCount(),
     ]);
+    const nodes = featureFlags.map((flag) => ({ ...flag, exposures: flag.featureFlagExposures.length }));
```
VivekFitkariwala, line 249: This transformation should be done inside the service itself.

VivekFitkariwala (changes requested):

VivekFitkariwala (dismissed):
Some types changes

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -107,7 +124,9 @@ export class FeatureFlagService {
   ): Promise<FeatureFlag[]> {
```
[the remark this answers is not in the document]
bcb37: It looks like `loadRelationCountAndMap()` doesn't change the return type of `getMany()`. So even though the query returns the correct type, type assigned to the return value of `getMany()` is still the type of the model. It's possible to remedy this with a `@VirtualColumn` but that decorator is only available in version 0.3.11.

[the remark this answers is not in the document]
VivekFitkariwala: Ohh I see it. Either we can 

1. Update the typeORM and fix this (This should be done as a separate PR)
2. Use type assertion and add a TODO

```
Promise<(Omit<FeatureFlag, 'featureFlagExposures'> & { featureFlagExposures: number })[]> // For function return

return queryBuilder.getMany() as unknown as (Omit<FeatureFlag, 'featureFlagExposures'> & { // For function return
      featureFlagExposures: number;
    })[];
```

3. ignore it for now and add TODO to add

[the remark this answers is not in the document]
bcb37: I choose 3

VivekFitkariwala (approved):

PR #1874, opened 2024-08-27
use upsert() when creating users

bcb37:


VivekFitkariwala (dismissed):

ppratikcr7 (dismissed):

danoswaltCL:
To document what was said in meeting about this, Ben asked do we want to merge this into 5.2, I believe we agreed that it was good to merge this into 5.2 because these commits were unexpectedly not there and had been tested previously to eliminate at least the 'duplicate key contraint' error that pops up, so should be considered to be pretty safe. therefore yes, commit a version bump and then put in staging to test the hotfix (somehow).

and then @VivekFitkariwala was going to also put in PR for something with indexing the table right?

danoswaltCL (approved):

VivekFitkariwala (approved):

PR #1890, opened 2024-08-30
give precedence to exclude lists

bcb37:
Reverses the precedence of include and exclude lists. Now if a user (or their group or segment) is on an exclude list, with filterMode set to 'excludeAll', they will be excluded from feature flags and experiments, even if they are also on an enabled inlcude list.


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1867,21 +1867,22 @@ export class ExperimentAssignmentService {
...
       } else {
-        if (explicitIndividualInclusionFilteredData.some((x) => x.id === modal.id)) {
-          userIncludedModals.push(modal.id);
-        } else if (explicitIndividualExclusionFilteredData.some((x) => x.id === modal.id)) {
+        if (explicitIndividualExclusionFilteredData.some((x) => x.id === modal.id)) {
           userExcludedModals.push({ id: modal.id, reason: 'filterMode' });
+        } else if (explicitIndividualInclusionFilteredData.some((x) => x.id === modal.id)) {
+          userIncludedModals.push(modal.id);
         } else {
+          console.log('ELSE', userGroups);
```
VivekFitkariwala, line 1875: Remove this comment

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1867,21 +1867,22 @@ export class ExperimentAssignmentService {
...
               )
             ) {
               exclusionFlag = true;
             }
           }
-          if (!exclusionFlag) {
+          if (exclusionFlag) {
             userExcludedModals.push({ id: modal.id, reason: 'filterMode' });
           } else {
             userIncludedModals.push(modal.id);
```
VivekFitkariwala, line 1888: This will include all the users which are not in the explicitGroupExclusionFilteredData. But these users should be in explicitGroupInclusionFilteredData for them to push inside userIncludedModals

VivekFitkariwala (approved):

PR #1900, opened 2024-09-04
adding build config for scheduler lambda and s3 bucket

shpwe:
add email bucket import

add taskRoleArn import

Add automatic deploy for lambda function to qa

add postbuild npm command

removing shared-infrastructure template from repo


kcalvo-cli (approved):
LGTM

PR #1903, opened 2024-09-05
initial commit to add mooclet proxy to backend

danoswaltCL:
This is an initial draft of the Mooclet integration. This is sort of a minimum working example of creating a special "mooclet"-linked experiment and being able to enroll / receive assignments.

[Roadmap and User flow](https://docs.google.com/spreadsheets/d/14tWb1UyThkMinFAVsmkmXRcv9aixcAY6p-LTs--aQuU/edit?gid=1005455694#gid=1005455694)

[Mooclet API ref](https://docs.google.com/spreadsheets/d/1eL3z5zJfpraqEONYMVhiWgGYj6PSm8T8KFKFyeOMnHU/edit?gid=1544484998#gid=1544484998)

[Mooclet PolicyParameters](https://docs.google.com/document/d/1OuclKLK2mi6BPm76TckTVyIeAoBB394oGmXNcCCtiOs/edit#heading=h.ekzddajcso5r)

[Google Drive link to Mooclet docker image and local setup instructions (not sure who has access to research team folder, let me know if you want to try this)](https://drive.google.com/drive/folders/11M37FpMSFwoYFa7LigzbPYoObQfMG5SD)

[EASI repo](https://github.com/danoswaltCL/mooclet-engine) // this will be changing....

[Carnegie Learning Repo](https://github.dev.carnegielearning.com/CarnegieLearning/mooclet-engine)

This would be easier to review in depth over a call, I'll be adding more notes as I go for references, but I wanted to get this draft up for the curious.

- Adds configuration changes to connect to external "mooclet" API service
- Adds `axios`
- Adds Mooclet types to the global types folder
- Adds new `ASSIGNMENT_ALGORITHM`s for algorithms we will support
- Adds a property called `moocletDetails` to the `Experiment` entity
- Adds ability to create a parallel `mooclet` experiment when a mooclet experiment algorithm is selected and it's `policyParameter` (aka: experiment design options) supplied to `POST /experiment`
- Adds ability to bypass UpGrade's assignment algorithm on `/assign` to get condition from mooclet API instead

This does NOT add:
- Syncing Edit Experiment w/ Mooclet API
- Syncing Delete Experiment w/ Mooclet API
- Syncing `/log` calls w/ Mooclet API
- frontend changes yet
- unit tests / integration tests


backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
```diff
@@ -463,6 +464,9 @@ export class ExperimentDTO {
   @IsNotEmpty()
   @IsEnum(EXPERIMENT_TYPE)
   public type: EXPERIMENT_TYPE;
+
```
danoswaltCL, line 467: I don't know how to add validation for this... this interface will have a property called "moocletDetails.policyParameters" that will contain what is basically a config object that is different for each algorithm that we will support. See "policyParameters" interfaces in the Mooclet types file for examples.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1525,15 +1536,15 @@ export class ExperimentAssignmentService {
     }
   }
 
-  private assignExperiment(
+  private async assignExperiment(
```
danoswaltCL, line 1539: because we are going to sometimes potentially be calling out to another service, I had to make assignExperiment return a Promise, which causes me to have to modify this function in a bunch of places. I don't love this, but I wasn't sure how else to accomplish it. Ideas welcome.

types/src/Experiment/enums.ts
```diff
@@ -19,6 +19,10 @@ export enum CONDITION_ORDER {
 export enum ASSIGNMENT_ALGORITHM {
   RANDOM = 'random',
   STRATIFIED_RANDOM_SAMPLING = 'stratified random sampling',
+
```
danoswaltCL, line 22: I am not sure about this! In the frontend, the user will be able to select additional algorithms. We will be able to add them as we test and support them, mooclet has a dozen or so algorithms we could potentially support. So I guess I need to add them to this enum instead of elsewhere with other Mooclet types?

backend/packages/Upgrade/src/database/1725913309740-migrations.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Migrations1725913309740 implements MigrationInterface {
+    name = 'Migrations1725913309740'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "experiment" ADD "moocletDetails" json`);
+        await queryRunner.query(`ALTER TYPE "public"."experiment_assignmentalgorithm_enum" RENAME TO "experiment_assignmentalgorithm_enum_old"`);
+        await queryRunner.query(`CREATE TYPE "public"."experiment_assignmentalgorithm_enum" AS ENUM('random', 'stratified random sampling', 'uniform_random', 'ts_configurable')`);
```
bcb37, line 9: Should the algorithm types be constrained by an enum in the UpGrade db? It seems like that would limit the flexibility of being able to add/define different algorithms in the mooclet server.

danoswaltCL: it's definitely one of the things I'm looking for feedback on. It sure is convenient for several reasons to be able to attach an algorithm type enum to the experiment, it feels like that's where it has to go. Each algorithm we want to add support for later will require at least some minimal code updates because each one has a special set of experiment design options called "policy-parameters" that we'll want to validate, and potentially some bespoke usage and testing, so it will need to be a deliberate code change task anyway to include each supported algorithm.

bcb37: I did notice that you could add algorithms ("policies") arbitrarily in the django mooclet interface, but there's no way to add the corresponding function without changing the API code. That said, we can still have enums in the application without enforcing them in the database schema. An argument for that would be: if we wanted to support adding and removing the ones that are already defined in the mooclet, it would be easier if it didn't always have to involve a database migration - especially in the case of removing algorithms.

backend/packages/Upgrade/src/api/models/Experiment.ts
```diff
@@ -182,5 +182,5 @@ export class Experiment extends BaseModel {
     nullable: true,
     type: 'json',
   })
-  public moocletDetails?: MoocletExperimentRef;
+  public MoocletExperimentRef?: MoocletExperimentRef;
```
bcb37, line 185: This should be `moocletExperimentRef`.

bcb37:
> I put this comment in wrong PR yesterday.
> 
> So to recap the suggestions from @VivekFitkariwala / @bcb37 / @ppratikcr7 in call (ty, this was exactly what I was looking for). Let me know if this matches what was said:
> 
> * Don't update Experiment Entity model, as we won't want to store these details.
> * Do send in user-supplied policy-parameters on the ExperimentDTO when POST / PUT
> * Do send back policy-parameters in response DTO so that they can be edited in UI (but other mooclet metadata may not be needed).
> * Create a new table in db that minimally stores: mooclet id, version ids, policy parameters json (other info TBD) that may be required to associate UpGrade objects with Mooclet objects. The upgrade experiment id would be the foreign key.
> * Update the mooclet service to query this new table instead of depending on a new "experiment.moocletDetails" object stored with the Experiment.
> 
> Point being that it's more efficient to store the data this way, and will allow more straightforward validation and decoupling of information so that we don't have to involve the existing Experiment table in this Mooclet business.
> 
> TBD is Ben's question on how to handle the mooclet assignment algorithm enums. In my changes, I've done this by adding additional mooclet-specific enums to the existing upgrade ASSIGNMENT_ALGORITHM enum list and storing that directly on the Experiment also. I'm not sure how to get around that, as we have to have some kind of algorithm choice for the experiment, it's not random or stratified-random. Does that still seem ok if we are trying to decouple these objects?

I was wondering: Is it necessary at all to store any of the mooclet data in upgrade? If we're creating one mooclet per experiment, couldn't we assume the policy parameters stored in the mooclet db for the given mooclet are the source of truth? Then we wouldn't be maintaining the same information in two places. On experiment creation (or edit) we could populate the assignment algorithm dropdown with the algorithms from the mooclet server. On the backend those could all map to, say,  a `MOOCLET` algorithm type, which would tell UpGrade to fetch the data from the mooclet server by the experiment ID (which could also be the mooclet name).

danoswaltCL:
This is probably true, I think my original thought was to limit the number of calls to the mooclet API for performance reasons, but for the experiment CRUD operations that is really not an important factor. My only concern there is the `/assign` call, where I believe we would always need to do an extra lookup (I think) to get the mooclet id by experiment id, in order to call the requisite mooclet function that gives us an assignment:  

But overall yes we may in fact be able to get by with just storing a table mapping Upgrade Experiment id to Mooclet id, although storing the id of each of the mooclet entities tied to the base-mooclet (like versions id and policyparameters id or variable id) is probably going to be better because in the case of editing a mooclet, that API has you to edit the modified entity directly via PUT /version/:id, PUT /policyParameters...

As for algorithm enums though, somewhere we will need to have a list of which mooclet algorithms we support in that UpGrade version.

danoswaltCL:
hm, each upgrade condition id may need to map to a mooclet version id. how should that be stored? Here's what I'm playing with:

As columns:

```
import { Entity, PrimaryGeneratedColumn, Column, ManyToOne } from 'typeorm';

@Entity()
export class MoocletExperimentRef {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  moocletId: number;

  @Column({ type: 'json', nullable: true })
  versionIdMapList: { [key: string]: string }[]; // <--- { upgradeConditionId : moocletVersionId }

  @Column({ nullable: true })
  policyParametersId: number;

  @Column({ nullable: true })
  variableId: number;
}
```


Or as JSON?
```
import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';

@Entity()
export class MoocletExperimentRef {
  @PrimaryGeneratedColumn()
  id: number;

  @Column()
  moocletId: number;

  @Column({ type: 'json', nullable: true })
  metadata: {
    version?: { [key: string]: string }[];
    policyParametersId?: number;
    variableId?: number;
  };
}
```

bcb37:
> hm, each upgrade condition id may need to map to a mooclet version id. how should that be stored? Here's what I'm playing with:
> 
> As columns:
> 
> ```
> import { Entity, PrimaryGeneratedColumn, Column, ManyToOne } from 'typeorm';
> 
> @Entity()
> export class MoocletExperimentRef {
>   @PrimaryGeneratedColumn()
>   id: number;
> 
>   @Column()
>   moocletId: number;
> 
>   @Column({ type: 'json', nullable: true })
>   versionIdMapList: { [key: string]: string }[]; // <--- { upgradeConditionId : moocletVersionId }
> 
>   @Column({ nullable: true })
>   policyParametersId: number;
> 
>   @Column({ nullable: true })
>   variableId: number;
> }
> ```
> 
> Or as JSON?
> 
> ```
> import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';
> 
> @Entity()
> export class MoocletExperimentRef {
>   @PrimaryGeneratedColumn()
>   id: number;
> 
>   @Column()
>   moocletId: number;
> 
>   @Column({ type: 'json', nullable: true })
>   metadata: {
>     version?: { [key: string]: string }[];
>     policyParametersId?: number;
>     variableId?: number;
>   };
> }
> ```

I'm still not sure I understand the relationships among 'Verision' 'Variable' and 'Value'. Do we need to use all of those? There seems to be a lot of redundancy.

backend/packages/Upgrade/src/api/models/Experiment.ts
danoswaltCL: this property will be going away on the Experiment model anyway

VivekFitkariwala:
> hm, each upgrade condition id may need to map to a mooclet version id. how should that be stored? Here's what I'm playing with:
> 
> As columns:
> 
> ```
> import { Entity, PrimaryGeneratedColumn, Column, ManyToOne } from 'typeorm';
> 
> @Entity()
> export class MoocletExperimentRef {
>   @PrimaryGeneratedColumn()
>   id: number;
> 
>   @Column()
>   moocletId: number;
> 
>   @Column({ type: 'json', nullable: true })
>   versionIdMapList: { [key: string]: string }[]; // <--- { upgradeConditionId : moocletVersionId }
> 
>   @Column({ nullable: true })
>   policyParametersId: number;
> 
>   @Column({ nullable: true })
>   variableId: number;
> }
> ```
> 
> Or as JSON?
> 
> ```
> import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';
> 
> @Entity()
> export class MoocletExperimentRef {
>   @PrimaryGeneratedColumn()
>   id: number;
> 
>   @Column()
>   moocletId: number;
> 
>   @Column({ type: 'json', nullable: true })
>   metadata: {
>     version?: { [key: string]: string }[];
>     policyParametersId?: number;
>     variableId?: number;
>   };
> }
> ```

We should have a separate table for storing the version condition map. This will help in doing the CRUD operations and the performance due to extra Joins will be taken care by caching.
You can use [this](https://typeorm.io/listeners-and-subscribers) to sync the Experiment specific data. This will keep the code decoupled.

danoswaltCL:
Update: Created a mooclet service that extends the experiment service, a couple mooclet entity ref tables instead of experiment json objects, and a function to create both the experiment and the mooclet in a single transaction.

VivekFitkariwala (changes requested):

bcb37 (commented):
These changes work, but I'm not sure about a couple things:
- How are we getting the policy parameter values and sending them to the mooclet API now? There didn't seem to be anything in the entity definition, without sending anything it ends up being null
- How is the mooclet/experiment relationship set up in the database? Should there be an `experimentId` foreign key in the 'mooclet_experiment_ref' table?

backend/packages/Upgrade/src/api/errors/UnprocessableEntity.ts
```diff
@@ -0,0 +1,7 @@
+import { HttpError } from 'routing-controllers';
+
+export class UnprocessableEntityError extends HttpError {
```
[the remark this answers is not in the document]
danoswaltCL: what would be preferable? I throw this when the experiment can't be saved because of maybe unknown Mooclet server issues so I don't have to throw a 500 server error.

danoswaltCL:
I've implemented most of the requested changes, but as I try to get the validation aspect to work I am now thinking.... should I just have a separate Mooclet Controller in the first place? To keep the logic and types even more separate and maybe make this a little more straightforward?

UI will know to send to /mooclet endpoints

```
MoocletExperimentDTO extends ExperimentDTO {
  moocletExperimentRef?: MoocletExperimentRef <--- sent back in response, not in POST request
  moocletPolicyParameters: MoocletPolicyParameters <--- should be validated based on the experiment.assignmentAlgorithm (using TypeORM Discriminator?)
}
```

Mooclet Controller routes CRUD operations on MoocletExperimentDTO
Mooclet Experiment Service extends ExperimentService

maybe some other uses for /mooclet endpoints will be useful, such as for validating policyParameters for the UI so the user entry can be corrected before submitting

@VivekFitkariwala @bcb37

backend/packages/Upgrade/src/api/errors/UnprocessableEntity.ts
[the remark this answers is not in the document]
VivekFitkariwala: I mean, this is not getting used anywhere. Have you created this for future use?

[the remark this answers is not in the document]
danoswaltCL: ah, i thought i was using it, but I had imported and used import { UnprocessableEntityException } from '@nestjs/common' instead! it works the same apparently, I guess that's just as good so I'll delete this

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -0,0 +1,397 @@
...
+          const map = await manager.save(MoocletVersionConditionMap, versionConditionMapEntity);
+          logger.info({
+            message: 'MoocletVersionConditionMap created successfully:',
+            versionConditionMap: JSON.stringify(map),
+          });
+        }
+      }
+    } catch (error) {
+      logger.error({
+        message: 'Failed to sync experiment with Mooclet',
```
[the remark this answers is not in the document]
danoswaltCL: I agree, I'd like to try and that and other tricky scenarios in a follow-up PR once the basics are checked into dev

danoswaltCL:
@VivekFitkariwala @bcb37 I have pushed up changes to resolve most of the questions I think that were posed, I didn't add any fancy validation or try to make a separate controller at this point.

I'm thinking maybe policy-parameter validation and some more edge cases can be handled in follow-up PRs if we can focus on getting this one in good enough shape to finalize the basic architecture.

VivekFitkariwala (changes requested):
Small changes.

backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
danoswaltCL: i will try using discriminator in follow-up PR

PR #1911, opened 2024-09-09
Fix undefined segment errors in Feature Flag Inclusion and Exclusion selectors

### LINKED ISSUE #1910: TypeError: Cannot read properties of undefined (reading 'createdAt') when navigating to some FF details page
6.0.0

**Describe the bug**
When navigating to some FF details page (e.g., Test13 from QA env), the page doesn't properly load the Include/Exclude Lists due to the `TypeError: Cannot read properties of undefined (reading 'createdAt')` error.

**To Reproduce**
1. Go to https://upgrade.qa-cli.net/
2. Sign-in and go to FF root page
3. Click the feature flag named "Test13"
4. See error on the browser console.

**Expected behavior**
The error should not happen.

**Screenshots**
Screenshot 2024-09-10 at 1 18 52 AM

zackcl:
Resolves #1910 

This PR fixes the `TypeError: Cannot read properties of undefined (reading 'createdAt')` error that occurs when navigating to some Feature Flag details pages. This error is caused by attempting to access the `createdAt` property from `segment` when `segment` is not available from the store (This issue started happening after merging #1907 which included `featureFlagSegmentInclusion` in the `flags/paginated` response). The fix involves adding a filter step: `.filter((inclusion) => inclusion.segment)` to ensure segment exists before accessing its properties. This PR also simplifies the existing selector code.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.selectors.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.selectors.ts
@@ -139,35 +139,33 @@ export const selectIsLoadingFeatureFlagDelete = createSelector(
 export const selectFeatureFlagInclusions = createSelector(
   selectSelectedFeatureFlag,
   (featureFlag: FeatureFlag): ParticipantListTableRow[] => {
-    if (!featureFlag || !featureFlag.featureFlagSegmentInclusion) {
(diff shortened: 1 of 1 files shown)


bcb37 (approved):
This works, but I can't reproduce the error on dev now.

VivekFitkariwala (approved):

PR #1920, opened 2024-09-10
add missing endpoint

bcb37:
This was breaking FF imports on dev. Shouldn't these always be added to all environments?


danoswaltCL:
i think i know what happened with these actually.... these two environment files (`beanstalk` and `qa`) did not used to exist, so maybe they were in PR status when those environment files were introduced and we didn't spot it in review.

`beanstalk.prod` and `qa` were added recently while implementing ECS / Jenkins job changes because we are temporarily deploying via 2 different pipelines. The "old" way was to actually always deploy the `environment.prod` config regardless of environment, which is still the case, but it was renamed `beanstalk.prod` to separate it from the "new" build configs. Those configs really are based on each environment. A `staging` one existed already even though it wasn't used, so a `qa` one needed to be added as well. These configs will map to the new domain envs.

danoswaltCL (approved):

PR #1922, opened 2024-09-11
Fix delete exclusion list throwing error

RidhamShah:
This fixes the frontend error throwing on deleting elements of the inclusion list though getting 200 response from backend.


zackcl (approved):

danoswaltCL:
but do we know how/why an inclusion segment could come back without an id?

PR #1924, opened 2024-09-12
Add description under flag name with tooltip

### LINKED ISSUE #1704: FF: Add description under flag name with tooltip
We should be showing a tooltip over the description in small text below the title when present in the root table, same as the experiments and segments tables:

Image

zackcl:
Resolves #1704

This PR displays the description under flag name with a tooltip when the text is truncated (for both the name and description) just like how it works on the Experiments root page.

Screenshot 2024-09-12 at 5 31 36 PM

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
@@ -18,12 +18,20 @@
       <td mat-cell *matCellDef="let flag" class="name-column ft-14-400">
         <a
           [routerLink]="['/featureflags', 'detail', flag.id]"
-          [matTooltip]="flag.name.length >= 30 ? flag.name : null"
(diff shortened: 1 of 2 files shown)


RidhamShah:
@zackcl This works fine, but I also noticed that in Experiment we also provide a tooltip for Table-Header (Name, Status, Tags ... ) but it is not there for FeatureFlags. Can you add that too?

zackcl:
@RidhamShah Why do we need to show a tooltip for table headers? I don't think we need to display the tooltip if we are just showing the same text as the headers. But please let me know if you have any specific suggestions for different tooltip messages.

RidhamShah:
> @RidhamShah Why do we need to show a tooltip for table headers? I don't think we need to display the tooltip if we are just showing the same text as the headers. But please let me know if you have any specific suggestions for different tooltip messages for the headers.

True, but i think we should make it consistent between Experiment, FeatureFlags & Segments

zackcl:
@RidhamShah Currently, nothing is consistent between Feature Flags and other pages since it's a new design. We decided to apply the new design one page at a time, so there may be inconsistencies, but eventually they will be consistent.

bcb37 (commented):
The positioning of the tooltip seems confusing. It appears to cover um the name that the description corresponds to, rather than showing up under the name: 

Screenshot 2024-09-13 at 4 59 22 PM

VivekFitkariwala (approved):

zackcl:
@bcb37 I tried different positioning, but displaying a tooltip above an element seemed most natural to me. It ends up covering other elements even with different positioning anyway. Also, the user can slightly move the cursor to see the name when it's covered.

Issue #1575, opened 2024-05-23
FF Root Table filtering with the search widget


danoswaltCL:
Add filtering functionality to the root table.

- The search widget should affect filtering in the table as you type
- Clicking a tag chip should also affect filtering in the table

Bonus points if we can make this filtering functionality part of a general shared "table utility" service. We'll need this behavior for any table, and as you can see in the existing code in the experiments and segments for example, filtering and sorting code is heavily duplicated and very confusing to figure out, so let's keep it clean and not repeat that!

zackcl:
@Yagnik56 Is this Done? I think we should allow searching by "Tag", and also allow searching by "Tag" by clicking on a tag chip on the table (see my last comment).

zackcl:
@Yagnik56 For your information, since now we support capital letters for user-entered values/tags (#1927), the tags shouldn’t be case-sensitive in matching. If it’s “mathia” in one feature flag but “MATHia” in another, we should be able to click (or search) one and see both.

PR #1927, opened 2024-09-12
Support capital letters for user-entered values/tags

### LINKED ISSUE #1925: FF: Support capital letters for user-entered values/tags
Currently, the user-entered values and tags are automatically converted to lowercase. This makes it not possible to create a tag such as "MATHia" or values/IDs that contain capital letters. Also, we support including capital letters for experiment participants, but not feature flag participants. We should make it consistent.

We should support capital letters for user-entered values/tags for better usability and flexibility.

**Additional context**
Tags shouldn’t be case-sensitive in matching. If it’s “mathia” in one feature flag but “MATHia” in another, we should be able to click (or search) one and see both.

zackcl:
Resolves #1925

Note: The tags being case-insensitive in matching should be handled after https://github.com/CarnegieLearningWeb/UpGrade/issues/1575#issuecomment-2338359452 is resolved.


bcb37 (approved):
This fixes forcing tags and values to be lower case in feature flags. However:
 - tags are still not searchable on the feature flags list page
 - tags are still forced to be lower case in experiments

VivekFitkariwala (approved):

PR #1948, opened 2024-09-13
Update validation to show Warning

RidhamShah:
While importing a FeatureFlag if there is a non-existent Segment reference is present it should see a warning in the import dialog box and can be imported, but for some reason instead of a WARNING it was showing COMPATIBLE.

image

Enhanced the validation process to prevent importing multiple files with duplicate feature flags, before it would allow and throw multiple errors while importing.

image

### DIFF
--- a/backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
+++ b/backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
@@ -804,43 +804,51 @@ export class FeatureFlagService {
   ): Promise<ValidatedFeatureFlagsError[]> {
     logger.info({ message: 'Validate feature flags' });
 
-    const featureFlagsIds = featureFlagFiles
-      .map((featureFlagFile) => {
-        try {
-          return JSON.parse(featureFlagFile.fileContent as string).key;
-        } catch (parseError) {
-          return null;
-        }
-      })
-      .filter((key) => key !== null);
+    const parsedFeatureFlags = featureFlagFiles.map((featureFlagFile) => {
+      try {
+        return {
+          fileName: featureFlagFile.fileName,
+          content: JSON.parse(featureFlagFile.fileContent as string),
+        };
+      } catch (parseError) {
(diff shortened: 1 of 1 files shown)


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -797,6 +797,7 @@ export class FeatureFlagService {
     logger.info({ message: 'Imported feature flags', details: createdFlags });
     return fileStatusArray;
   }
+
   public async validateImportFeatureFlags(
```
VivekFitkariwala, line 801: Can you refactor this code? You are parsing the file twice once to get the `featureFlagsIds` and the second time to get the `validationErrors`. During the featureFlagIds parsing, you are not throwing any error; the second time, returning `INCOMPATIBLE`.
You should parse only once and use that result for further validation.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -874,6 +884,11 @@ export class FeatureFlagService {
         ];
 
         const segments = await this.segmentService.getSegmentByIds(segmentIds);
+
+        if (segmentIds.length !== segments.length) {
```
VivekFitkariwala, line 888: This will fail if `segmentIds` have duplicate values. Use `Set` to remove duplicate values.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code refractored

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

VivekFitkariwala (approved):

bcb37 (approved):

PR #1958, opened 2024-09-16
Closing the experiment stepper after clicking on save is not working as expected with other small edit bugs

Yagnik56:
This PR will fix the issue of the experiment modal not closing after saving changes in edit.

The "Save" button on the overview stepper is not disabled on context/experimentType change, as we had earlier, this might have been changed in some commit. We need to ensure the experimenter moves to the design page to update the decision points and conditions according to the updated context/experimentType, or else the wrong design is mapped with the context or experimentType. The "Save" button should remain enabled for all other edits other than context/experimentType on the Overview page.

If we select the assignment algorithm as SRS but we don't have any factor to select from then clicking on next will throw an error but if it moves back to random it should allow the user to go next which was not happening. This PR fixes that issue by moving the assignment algorithm back to random.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-overview/experiment-overview.component.html
@@ -1,44 +1,21 @@
 <div *ngIf="!(isLoadingContextMetaData$ | async); else loadingExperimentOverview" class="shared-modal--step-container">
   <section class="shared-modal--form-container">
     <form class="experiment-overview dense-3" [formGroup]="overviewForm">
-      <mat-form-field
-        class="name"
-        appearance="outline"
-        subscriptSizing="dynamic"
-      >
+      <mat-form-field class="name" appearance="outline" subscriptSizing="dynamic">
         <mat-label class="ft-14-400">
           {{ 'home.new-experiment.overview.name.placeHolder' | translate }}
         </mat-label>
-        <input
-          class="ft-14-400"
-          matInput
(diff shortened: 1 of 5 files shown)


ppratikcr7:
@Yagnik56 This works fine, but observed that the "Save" button on overview stepper is not disabled on context change, as we had earlier, maybe this got changed in some commit. We need to make sure the experimenter moves to the design page for updating the decision points and conditions as per the updated context, else wrong design is mapped with the context. The "Save" button should remain enabled for all other edits other than context on Overview page.

ppratikcr7 (changes requested):
Please check the review comment.

Yagnik56:
@ppratikcr7 disabled the save button if context or experiment type changes in overview page.

ppratikcr7:
@Yagnik56 It works partially. In case I update the context to say A to B and then again update from B to A to keep it same, the save button is still disabled. For experiment type change this scenario enables the save button. This should happen for context too. Else, the experimenter has to forcibly go to next page to update/save any other value of overview page.

ppratikcr7:
Also, found a new page on this overview page. The Assignment Algorithm when updated from Random to SRS without a stratification factor and tried to move next, will show an error on the stepper, which is expected. But, when you again update it to Random, you are unable to move to next design page and the error message also stays there. Also, the error message is overlapped a bit by the buttons. Can you add a proper padding around it. And align the logging checkbox in a straight left margin line. Please refer below:

Screenshot 2024-09-16 at 5 05 16 PM
Screenshot 2024-09-16 at 5 05 28 PM

VivekFitkariwala (approved):

PR #1960, opened 2024-09-16
return 'warning' compatibility when segments are not found

bcb37:
Send the 'warning' compatibility type when public segments in imported lists are not found.


RidhamShah:
This issue is solved and the same code is present in my [PR](https://github.com/CarnegieLearningWeb/UpGrade/pull/1948).

bcb37:
Duplicate of https://github.com/CarnegieLearningWeb/UpGrade/pull/1948

PR #1961, opened 2024-09-18
feature flag root table sorting functionality issue fixed

Yagnik56:
This PR fixes the sorting issue in the feature flag root table.


zackcl:
I tried this PR, but the sorting works differently from the Experiments root page. On the Experiments root page, it sorts the lists regardless of their capitalization, which I believe is the expected behavior.

Screenshot 2024-09-18 at 6 43 26 PM


However, on the Feature Flags root page, it sorts the lists by flags starting with uppercase and lowercase separately.

Screenshot 2024-09-18 at 6 43 44 PM

I believe this should work the same way as the Experiments root page. My research of some other apps shows they also sort lists regardless of capitalization.

zackcl (approved):

VivekFitkariwala (approved):

PR #1963, opened 2024-09-18
Remove error snackbar for 401 errors that redirects to the login page

### LINKED ISSUE #1846: login/user api fails on login screen probably during routine token refresh
e.g "6.0.0"

**Describe the bug**
The login/user api gives 401 Unauthorised error when we are auto logged out from Upgrade or during routine token refresh.

**To Reproduce**
1. Go to upgrade login page or maybe after restarting the backend after a long time would trigger auto log out for logged in user in Upgrade.
2. See the snackbar error as below.

**Expected behavior**
We should not get that error before clicking the login button and we have some messed up login params.

**Screenshots**
image

zackcl:
Resolves #1846

This PR removes the error snackbar for 401 Unauthorized errors that redirects to the login page.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
+++ b/frontend/projects/upgrade/src/app/core/http-interceptors/http-error.interceptor.ts
@@ -30,9 +30,9 @@ export class HttpErrorInterceptor implements HttpInterceptor {
         if (err.status === 401) {
           // auto logout if 401 response returned from api
           this.authService.authLogout();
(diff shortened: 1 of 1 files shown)


zackcl:
Here's the screen recording showing what happens if it gets redirected to the login page due to 401 error after the fix.

https://github.com/user-attachments/assets/e9d0601b-14fd-414f-8ae9-d614a37890f9

VivekFitkariwala (approved):

PR #1966, opened 2024-09-19
Solved Error when clicking 'add-experiment-button'

RidhamShah:
The error encountered when clicking the "Add Experiment" button was caused by an issue with the metric structure, likely due to an improper configuration in the `.env` file. As the metric structure has been updated, it now requires a context array as one of its properties. However, it was receiving null or undefined instead of an array, which prevented the application from reading the metric's context and resulted in the error.

The solution provided in this PR will address the immediate bug, but the primary focus should be on correcting the metric data configuration in the `.env` file.


ppratikcr7:
So this resolves the error. If the context in metrics is null, we are able to create experiment, and we wont see the metrics in the stepper list and the context on Metrics Tab is blank. @danoswaltCL @VivekFitkariwala Hope this is fine.
Screenshot 2024-09-19 at 1 41 31 PM

ppratikcr7 (approved):

zackcl:
Isn't this the current behavior in https://upgrade.qa-cli.net/? I think we already have some metrics with `null` context. So I think it should be fine?

zackcl (approved):

VivekFitkariwala (approved):

Issue #1341, opened 2024-03-05
Java library changes for feature flags


amurphy-cl:
Update Java client library's `getAllFeaturedFlags` and `getFeatureFlag` methods to call POST `/featureflags`.

Use the new structure for request / response as defined in #1340

getAllFeatureFlags will internally call `POST /featureflags` with a similar structure as getAllExperimentConditions(), with userId and context, and will store the list of key strings on the client.

`getFeatureFlag` should take a string and return `true` if the flags list contains that flag key, and `false` if it is not found.

danoswaltCL:
This doesn't seem like it was done, reopening

amurphy-cl:
Closing as done.

PR #1968, opened 2024-09-19
#1341 update java client library feature flag methods

kawcl:


clientlibs/java/src/main/java/org/upgradeplatform/utils/Utils.java
```diff
@@ -20,16 +19,16 @@ public class Utils
...
-	public static final String INITIALIZE_USER= "api/v5/init";
-	public static final String SET_GROUP_MEMBERSHIP= "api/v5/groupmembership";
-	public static final String SET_WORKING_GROUP= "api/v5/workinggroup";
-	public static final String GET_ALL_EXPERIMENTS= "api/v5/assign";
-	public static final String MARK_EXPERIMENT_POINT= "api/v5/mark";
-	public static final String GET_ALL_FEATURE_FLAGS= "api/v5/featureflag";
-	public static final String LOG_EVENT= "api/v5/log";
-	public static final String SET_ALT_USER_IDS= "api/v5/useraliases";
-	public static final String ADD_MATRIC= "api/v5/metric";
+	private static final String apiVersion = "api/v5";
```
danoswaltCL, line 22: as discussed, will need to be `v6` once the v6 client controller pr is merged in.

kawcl: changed to v6. also removed a redundant request object

danoswaltCL:
i have `hasFeatureFlag` changes for TS lib coming, btw @kawcl @bcb37

clientlibs/java/src/main/java/org/upgradeplatform/responsebeans/InitializeUser.java
```diff
@@ -4,24 +4,15 @@
 import java.util.Map;
 
 public class InitializeUser {
-    private String id;
```
bcb37, line 7: The response object still has the `id` in it so this is failing when I run the request. It seems like maybe they used to be identical but now we may not be able to use the response type to define the request.

clientlibs/java/src/main/java/org/upgradeplatform/client/ExperimentClient.java
```diff
@@ -92,20 +90,20 @@ public void close() {
 
 	// Initialize user with userId
 	public void init(final ResponseCallback<InitializeUser> callbacks) {
-		InitializeUser initUser = new InitializeUser(this.userId, null, null);
+		InitializeUser initUser = new InitializeUser(null, null);
 		initializeUser(initUser, callbacks);
```
bcb37, line 94: In the `initializeUser` method, the request object is created with the response type `InitializeUser`, which should have `id` on it (but no longer does). Is it possible we need a new request bean for init?

bcb37 (dismissed):
Requests and responses differ in v6 for `ExperimentUser` (used by the two 'group' endpoints), and `InitializeUser` (used by /init). I've created new request beans for these that work locally.

danoswaltCL (approved):

PR #1972, opened 2024-09-20
If feature flag context changed, all list will be removed

Yagnik56:
In this PR the issue, when the context of the feature flag is changed from the edit modal all the segments in include/exclude lists will be removed.


zackcl:
Is this PR ready? I just tried this branch but it doesn't seem to work.

zackcl:
I tried this branch and it seems like the warning message shows up even before updating the App Context on the "Edit Feature Flag" modal. Could you please update this?

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.ts
```diff
@@ -152,7 +155,14 @@ export class UpsertFeatureFlagModalComponent {
 
   // Close the modal once the feature flag list length changes, as that indicates actual success
   listenForFeatureFlagGetUpdated(): void {
-    this.subscriptions.add(this.isSelectedFeatureFlagUpdated$.subscribe(() => this.closeModal()));
+    this.subscriptions.add(
+      this.isSelectedFeatureFlagUpdated$.subscribe(() => {
+        this.closeModal();
+        if (this.isContextChanged) {
+          this.router.navigate(['/featureflags']);
```
VivekFitkariwala, line 162: Why should we navigate to featureflags page?

Yagnik56:
> I tried this branch and it seems like the warning message shows up even before updating the App Context on the "Edit Feature Flag" modal. Could you please update this so that the message appears after the App Context is changed?
> 
> Also, the "Save" button sometimes remains disabled on the "Edit Feature Flag" modal even after making changes to the form. I'm not entirely sure how to reproduce this, but this happened to me after editing feature flags a few times (Let me know if you can't reproduce it, I can maybe share a video).

addressed

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.ts
zackcl: Can't we make the modal wait until the deletion is done before it is closed? I think the page should remain on the details page for best user experience. There's no reason to navigate to the root page after editing a feature flag on the details page.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -350,7 +353,22 @@ export class FeatureFlagService {
...
       } = flag;
+
+      if (oldFlagDoc.context !== flagDoc.context) {
+        includeList?.forEach((list) => {
+          this.deleteList(list.segment.id, FEATURE_FLAG_LIST_FILTER_MODE.INCLUSION, user, logger);
+        });
+
+        excludeList?.forEach((list) => {
+          this.deleteList(list.segment.id, FEATURE_FLAG_LIST_FILTER_MODE.EXCLUSION, user, logger);
+        });
```
VivekFitkariwala, line 364: This should be the part of the transaction and should be deleted in a single database call. Please use await.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -351,6 +359,27 @@ export class FeatureFlagService {
         updatedAt,
         ...flagDoc
       } = flag;
+
+      if (oldFlagDoc.context[0] !== flagDoc.context[0]) {
+        // Create delete audit logs for inclusion and exclusion lists
+        if (includeListIds.length) {
```
VivekFitkariwala, line 365: You should use transaction to add the audit logs. Use `Promise.all` for all the three operations

VivekFitkariwala (dismissed):

zackcl (dismissed):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -405,34 +452,71 @@ export class FeatureFlagService {
...
+      }
+
+      // Use Promise.all to run all audit log saving operations concurrently
+      await Promise.all(auditLogPromises);
+    };
+
+    if (entityManager) {
+      await executeTransaction(entityManager);
+    } else {
+      await this.dataSource.transaction(async (manager) => {
```
VivekFitkariwala, line 516: You should not start a new transaction. Use the repository.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
VivekFitkariwala: Add Promise.all for all the requests

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
RidhamShah: code updated

VivekFitkariwala (approved):

PR #1973, opened 2024-09-20
Fix/refresh issues with hashing

### LINKED ISSUE #1904: Refresh (or URL) goes to Experiment list from FF details page
Steps to reproduce:
- Navigate to the details page for a Feature Flag (or Experiment)
- Refresh the page in the browser (or copy the URL and load it in a new tab)

Expected behavior: it should display the details page.

Actual behavior: The Experiment list page is displayed.

danoswaltCL:
Fixes refresh and direct url nav issues related to changing the hash-strategy. Resolves #1904

This also goes ahead and gives a boolean environment flag for `useHashRouting`, off for the "beanstalk" builds, on for the "ecs" builds, so we can more comfortably support both the beanstalk builds as we are used to with the non-hash-looking urls and the new builds with the hash routing on.

These were the scenarios tested (for both `useHashRouting` true and false):

Suppose the route in question is: `/featureflags/detail/blahblahblah` (or any page other than `/home`, not just feature flags)

When you are logged in:
- Refresh should keep you on your page.
- Direct url navigation should open to that route directly.

When not logged in:
- If you are logged out or auth has expired when pasting in that link, you should be prompted with login page. On success, the app should open on your intended route, not `/home`
- However, if you directly log yourself out via the log out button while you are on a specific page, then log back in, the expected behavior is for the app to treat you as a fresh login and that you will in fact be redirected to `/home`. This was because it was determined that it would be odd to log out as one user, and log in as another, and be directed to a resource that the previous user was visiting.


danoswaltCL:
@shpwe @kcalvo-cli  just copying you for awareness that`useHashRouting` will be an option in the frontend environment files, but I don't think it's something that will need to be parameterized for build scripts.

zackcl (approved):

PR #1980, opened 2024-09-24
Fix the visibility of multiple-row chips and the alignment of the icon button

### LINKED ISSUE #1976: Feature Flag Inclusion & Exclusion list values chips row is not visible properly after 3rd row of chips are added
e.g "6.0"

**Describe the bug**
The scroll is not working as expected, when 3rd row of values are added, the top row is half visible. Will create a separate for anyone who wants to take that up.

**To Reproduce**
1. Go to feature flag inclusion or exclusion list
2. Create an Individual list which will take up the 3rd row of chips.
3. Scroll up to see half visible chips in the first row

**Expected behavior**
Scroll should work fine to show proper chips

https://github.com/user-attachments/assets/a7230057-b315-4ad3-8999-3d2605caa515

zackcl:
Resolves #1976


ppratikcr7 (approved):
Works as expected.

PR #1988, opened 2024-09-26
Send 400 bad request error for no user-id header in api request

### LINKED ISSUE #1987: V6 client controller should not return success response on no `User-Id` header
e.g "6.0"

**Describe the bug**
V6 client controller should not return success response on no `User-Id` header

ppratikcr7:
This resolves #1987


backend/packages/Upgrade/src/api/middlewares/UserCheckMiddleware.ts
```diff
@@ -12,8 +12,16 @@ export class UserCheckMiddleware {
   public async use(req: AppRequest, res: AppRequest, next: express.NextFunction): Promise<any> {
     try {
       const user_id = req.get('User-Id');
-      req.logger.child({ user_id });
-      req.logger.debug({ message: 'User Id is:', user_id });
+      if (!user_id) {
+        const error = new Error(`User-Id header not found.`);
+        (error as any).type = SERVER_ERROR.USER_NOT_FOUND;
```
danoswaltCL, line 17: well it's a bad request type, not really a user not found, the server has incomplete info to even try to find a user

ppratikcr7: @danoswaltCL Yeah, I was also thinking to add a new ERROR type: `MISSING_HEADER_USER_ID`. But, got an input from our playpower team to reuse existing one's. I know this is for logged in user data not found from DB. I will update this if the above type looks good to you.

ppratikcr7: Screenshot 2024-09-26 at 2 04 58 PM

danoswaltCL (approved):

ppratikcr7:
@danoswaltCL I was yet to push the commit with your requested change, just showed you that does this looks good. Will create another small PR for the same.

PR #1991, opened 2024-09-26
Send 400 bad request error for no user-id header in api request 

ppratikcr7:
Proper enum for missing `User-Id` in request header.


zackcl (approved):

zackcl:
QA: Passed.

It looks like, currently, we allow passing a user ID for /api/v6/init in either the body or the header of the request, whereas we don't allow passing the user ID in the body for /api/v6/featureflag and /api/v6/assign. Is this expected behavior, or should we not allow passing a user ID in the body for consistency? @ppratikcr7 please create a ticket.

PR #1992, opened 2024-09-26
Refine Feature Flags and Segments List Sorting

zackcl:
This PR fixes the incorrect sorting of the Segments list after refreshing the page. It also updates the Feature Flags sorting to work just like the Segments and Experiments list tables, ensuring that the previous sorting state remains after a page refresh by storing the sorting states in local storage.

The Feature Flags sorting works fine in general, except that it always defaults to sorting by "Name" in ascending order upon page refresh, even though I can confirm that the previous value is stored in local storage. I would appreciate any ideas on how to fix this. We can also merge this one for now and address that issue later, as it's still an improvement over the current.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/local-storage/local-storage.service.ts
+++ b/frontend/projects/upgrade/src/app/core/local-storage/local-storage.service.ts
@@ -6,7 +6,8 @@ import {
   EXPERIMENT_SORT_KEY,
 } from '../experiments/store/experiments.model';
 import { SegmentLocalStorageKeys, SegmentState } from '../segments/store/segments.model';
-import { SEGMENT_SEARCH_KEY, SORT_AS_DIRECTION, SEGMENT_SORT_KEY } from 'upgrade_types';
+import { SEGMENT_SEARCH_KEY, SORT_AS_DIRECTION, SEGMENT_SORT_KEY, FLAG_SEARCH_KEY, FLAG_SORT_KEY } from 'upgrade_types';
+import { FeatureFlagLocalStorageKeys, FeatureFlagState } from '../feature-flags/store/feature-flags.model';
 
 const APP_PREFIX = 'UPGRADE-';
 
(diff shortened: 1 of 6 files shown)


zackcl:
@ppratikcr7 Just for your info, I realized that `featureFlagSortAs` and `featureFlagSortKey` are no longer being stored/used in the code. I think they were stored previously while working on this PR.

Screenshot 2024-09-27 at 10 46 47 PM

I can confirm that the values are correctly stored to and read from the local storage (You can search `FeatureFlagLocalStorageKeys.FEATURE_FLAG_SORT_TYPE` in the code base and add logging).

I think setting the default `sortKey` and `sortAs` in the `feature-flags.reducer.ts` might be overriding the values retrieved from the local storage when the page is refreshed.

```
export const initialState: FeatureFlagState = adapter.getInitialState({
  isLoadingUpsertFeatureFlag: false,
  isLoadingImportFeatureFlag: false,
  isLoadingFeatureFlags: false,
  isLoadingUpdateFeatureFlagStatus: false,
  isLoadingFeatureFlagDetail: false,
  isLoadingFeatureFlagDelete: false,
  isLoadingSelectedFeatureFlag: false,
  isLoadingUpsertPrivateSegmentList: false,
  hasInitialFeatureFlagsDataLoaded: false,
  duplicateKeyFound: false,
  activeDetailsTabIndex: 0,
  skipFlags: 0,
  totalFlags: null,
  searchKey: FLAG_SEARCH_KEY.ALL,
  searchValue: null,
  sortKey: FLAG_SORT_KEY.NAME,
  sortAs: SORT_AS_DIRECTION.ASCENDING,
});
```
But this is how it's also implemented in the `segments.reducer.ts` (which works fine) so I'm not sure about this.

Anyway, I'm not sure how to make the Feature Flags root table to keep the previous sorting state even after refresh just like other pages. I think it's OK to merge this for now unless anyone has ideas or suggestions to fix this minor issue.

zackcl:
@VivekFitkariwala @ppratikcr7 @Yagnik56 Please feel free to merge this PR.

ppratikcr7:
> @VivekFitkariwala @ppratikcr7 @Yagnik56 Please feel free to approve this PR.

@zackcl Yeah, it looks good. I also tried to check but you are correct, the code is similar for segments and experiment, couldn't find why on refresh its always ASC order. Maybe we can create a separate ticket for that and tackle that later.

ppratikcr7 (approved):

frontend/projects/upgrade/src/app/core/local-storage/local-storage.service.ts
```diff
@@ -66,7 +93,8 @@ export class LocalStorageService {
     };
 
     const state = {
-      experiments: experimentState, // experiment state,
+      experiments: experimentState,
+      featureFlagState: featureFlagState,
```
VivekFitkariwala, line 97: @zackcl Change this to `featureFlags` for local storage to work correctly

zackcl: That worked, thank you! I created a PR that fixes this: https://github.com/CarnegieLearningWeb/UpGrade/pull/2005

PR #1993, opened 2024-09-26
handle import validation edge cases

bcb37:
Fixes validation for imports of segments, feature flags and experiments for the following valid json file contents:
- [empty file]
- "abc"
- ""

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentService.ts
@@ -1513,10 +1513,16 @@ export class ExperimentService {
 
     const validationErrors = await Promise.allSettled(
       experimentFiles.map(async (experimentFile) => {
-        let experiment = JSON.parse(experimentFile.fileContent);
-
+        let experiment;
+        try {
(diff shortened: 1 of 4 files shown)


zackcl:
I tested importing the JSON files you mentioned and they all seem to work fine and display the "Invalid JSON format" message on the import modal, except when importing a JSON file with contents `"abc"` or `""` on the "Import Segments" modal. When I import these files, I get a long weird error message instead of "Invalid JSON format".

Screenshot 2024-09-27 at 7 45 14 PM

bcb37:
> I tested importing the JSON files you mentioned and they all seem to work fine and display the "Invalid JSON format" (or "Incompatible") message on the import modal, except when importing a JSON file with contents `"abc"` or `""` on the "Import Segments" modal. When I import these files, I get a long weird error message instead of "Invalid JSON format".
> 
> Screenshot 2024-09-27 at 7 45 14 PM

This is expected, the way segments validation is set up. Those two examples are technically valid JSON, so the segment validator fills in the missing attributes as undefined values. The long message is reporting each of the missing attributes (which is all of them, in this case).

VivekFitkariwala (approved):

PR #1996, opened 2024-09-30
Add 'System' Username to Audit Logs Generated with API Requests

### LINKED ISSUE #1983: Add Username to Audit Logs Generated with API Requests
Currently, the audit logs generated with API requests (e.g., create/delete experiment) don't have a username since the requests are made using an access token generated with a service account.

Screenshot 2024-09-26 at 1 10 48 AM

The audit logs generated with API requests should display the specific account user in some way (e.g., service account ID).

**Additional context**
Related issue: https://github.com/CarnegieLearningWeb/UpGrade/issues/1413 (closed)

zackcl:
Resolves #1983 

This PR adds 'System' username to audit logs generated with API requests. This was done by including system@gmail.com to the payload's email so that the existing [systemUserDoc](https://github.com/CarnegieLearningWeb/UpGrade/blob/dev/backend/packages/Upgrade/src/init/seed/systemUser.ts) can be used in this case. I had to also include `hd` to bypass the domain name checking.

```
payload = {
    hd: env.google.domainName,
    email: 'system@gmail.com',
};
```

Here's the screenshot of audit logs displaying the logs with 'System' username instead of empty username after this change:
Screenshot 2024-09-30 at 7 08 13 PM

### DIFF
--- a/backend/packages/Upgrade/src/auth/AuthService.ts
+++ b/backend/packages/Upgrade/src/auth/AuthService.ts
@@ -44,10 +44,11 @@ export class AuthService {
         if (!tokenInfo || !serviceAccountIds.includes(tokenInfo.aud)) {
           throw new Error('Invalid or unauthorized access token');
         }
+        payload = {
+          hd: env.google.domainName,
+          email: 'system@gmail.com',
+        };
         request.logger.info({ message: 'Access token validated' });
-        // For service account access tokens, we'll return null
-        // We might want to implement specific handling for service accounts here
-        return null;
       } catch (error) {
         request.logger.error(error);
         throw error;


danoswaltCL:
Better than blank for sure, a google service account "user" does have an email and user-like metadata associated with it though if you want to be more explicit: `test-app@upgrade-280713.iam.gserviceaccount.com`. I'm fine with saying "System" just to not be blank though.

image

danoswaltCL (approved):

zackcl:
> Better than blank for sure, a google service account "user" does have an email and user-like metadata associated with it though if you want to be more explicit: test-app@upgrade-280713.iam.gserviceaccount.com. I'm fine with saying "System" just to not be blank though.

Yeah, that would be better, but I'm not sure how to extract the email from a service account access token.

The `tokenInfo` (`const tokenInfo = await client.getTokenInfo(token);`) doesn't include anything like email or name as I checked.

PR #1999, opened 2024-10-01
Import will remove segments with Warning

RidhamShah:
Solve
- Added context to be matched for import FeatureFlag's Segment else throw as WARNING
- The Segments that are throwing warnings are ignored while importing


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -60,7 +60,8 @@ export class FeatureFlagService {
     @InjectRepository() private experimentAuditLogRepository: ExperimentAuditLogRepository,
     @InjectDataSource() private dataSource: DataSource,
     public experimentAssignmentService: ExperimentAssignmentService,
-    public segmentService: SegmentService
+    public segmentService: SegmentService,
+    private segmentArray = []
```
VivekFitkariwala, line 64: This service instance is used for multiple http requests and it is possible that it get overwritten by another request. You should not use private variable in the service.

VivekFitkariwala (changes requested):

ppratikcr7 (approved):

PR #2002, opened 2024-10-01
Disallow Adding, Importing, Editing, and Deleting a Feature Flag for Users with the 'Reader' Role

### LINKED ISSUE #1998: Adding, Importing, Editing, and Deleting a Feature Flag Should Not Be Allowed for Users with the 'Reader' Role
6.0.0

**Describe the bug**
Currently, adding, importing, editing, or deleting a feature flag is allowed for users with the 'Reader' role. We should not allow this.

**To Reproduce**
1. Go to the Profile page and update your role to 'Reader' (note: not sure how to revert to 'Admin' later though).
2. Go to the Experiments/Segments root and details pages, and observe that the options for adding, importing, editing, and deleting an experiment or segment are hidden or disabled.
3. Go to the Feature Flags root and details pages.
4. Observe that the options for adding, importing, editing, and deleting a feature flag are available.

**Expected behavior**
Adding, importing, editing, and deleting a feature flag should not be allowed for users with the 'Reader' role, just like how the Experiments and Segments pages work.

Root page:
- The "Add Feature Flag" button should be hidden or disabled.
- The "Menu" icon button should be hidden (since we currently have only the "Import Feature Flag" menu option).

Details page:
- "Edit Feature Flag", "Duplicate Feature Flag", and "Delete Feature Flag" menu options should be hidden.
- "Add Include List" and "Add Exclude List" buttons should be hidden or disabled.
- All slide toggles should not respond to mouse events, and the cursor should be `not-allowed` on hover.
- All "Edit" and "Remove" icons from the Include/Exclude Lists tables should be greyed out and not respond to mouse events.

zackcl:
Resolves #1998

This PR resolves the issue but the CSS styles for the disabled slide toggles and edit/remove buttons might need to be improved, because currently they are simply greyed out.


zackcl:
It seems that it's quite challenging to make the slide toggles disabled but retain their colored appearance. We have two potential approaches:

1. Set the "disabled" attribute to `true` for `mat-slide-toggle` to make the slide toggle unresponsive, and then override the colors to prevent it from appearing grey.
2. Add a custom class to the slide toggle when it's disabled, and use CSS to make it unresponsive.

I've attempted both approaches, and they both present significant difficulties. For now, I've simply set the "disabled" attribute to true to make the toggles unresponsive. I can revisit this issue later, but if anyone has a robust solution, please let me know.

Screenshot 2024-10-02 at 3 58 56 AM

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-details-page/feature-flag-details-page-content/feature-flag-exclusions-section-card/feature-flag-exclusions-section-card.component.html
```diff
@@ -10,7 +10,7 @@
   <!-- header-right -->
   <app-common-section-card-action-buttons
     header-right
-    [showPrimaryButton]="true"
+    [showPrimaryButton]="(permissions$ | async)?.featureFlags.update"
     [primaryButtonText]="'feature-flags.details.add-exclude-list.button.text' | translate"
```
Yagnik56, line 14: add disable check to `[actionsDisabled]` in `app-feature-flag-exclusions-table` below.

Yagnik56 (changes requested):
Some small changes need to be addressed, but overall PR looks good.

frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-details-page/feature-flag-details-page-content/feature-flag-exclusions-section-card/feature-flag-exclusions-section-card.component.html
zackcl: Good catch! Thank you!

ppratikcr7 (approved):

PR #2006, opened 2024-10-01
remove abandoned libtesters

danoswaltCL:
This removes the client libtesters, they are not going to be maintained and they are gathering dust / dependabot warnings.

A new and improved client-testing suite is planned with our contracted QA guy.


bcb37 (approved):
Was Chance going to look at these as a starting point, or would that just confuse matters?

Issue #1576, opened 2024-05-23
FF Root page table: Pagination on scroll


danoswaltCL:
Additional calls to fetch the next 'page' of paginated results should happen 'on scroll', per the existing behavior in other table.

While new data is pending, the "progress-indicator" banner at the very top of the table should "spin" like it does for experiments (aka, don't hide the table with the spinner used for loading initial data).

ppratikcr7:
@danoswaltCL I dont see paginated scroll happening for feature flags page as we have in experiments. All flags are shown on the page at once and a side scroll will scroll through the entire page.

ppratikcr7:
> [@danoswaltCL](https://github.com/danoswaltCL) I dont see paginated scroll happening for feature flags page as we have in experiments. All flags are shown on the page at once and a side scroll will scroll through the entire page.
> 
>  Screen.Recording.2024-09-12.at.2.05.33.PM.mov

@danoswaltCL Any views on this?

ppratikcr7:
> > [@danoswaltCL](https://github.com/danoswaltCL) I dont see paginated scroll happening for feature flags page as we have in experiments. All flags are shown on the page at once and a side scroll will scroll through the entire page.
> > Screen.Recording.2024-09-12.at.2.05.33.PM.mov
> 
> [@danoswaltCL](https://github.com/danoswaltCL) Any views on this?

@danoswaltCL ?

ppratikcr7:
@zackcl So I confirmed, its still a bug, paginated nodes in 1st load itself is total Feature Flags: 22
It should be 20 as per skip/take.
![Image](https://github.com/user-attachments/assets/d76f94e0-df61-4f71-af11-c714eef042e0)

ppratikcr7:
I will QA this.

ppratikcr7:
QA passed

PR #2020, opened 2024-10-04
use wheel for scroll

danoswaltCL:
`SharedModule` should be giving us access to the `ScrollDirective`, which is how the `(scrolled)` directive is being used elsewhere in the app, but for some reason it is not working that way, possibly there is something about being a standalone that makes this work differently.

The `(wheel)` directive seems to work fine though in place of `scrolled`, and a few other pieces that work similarly from the experiment list table seem to need to be included... I don't 100% know why this works and `scrolled` doesn't, but it does.

This'll re-close [#1576](https://github.com/CarnegieLearningWeb/UpGrade/issues/1576)


ppratikcr7 (approved):

danoswaltCL:
@zackcl Wait, did you get `scrolled` to work? If `scrolled` is working for you then let's use that

zackcl:
@danoswaltCL No, it's not working completely. The last page is not being fetched for some reason.

> However, when scrolled is used, it seems like the last page is not being fetched currently for some reason.

PR #2021, opened 2024-10-04
remove legacy feature flag components and store directories

danoswaltCL:


frontend/projects/upgrade/src/environments/environment.ts
```diff
@@ -11,7 +11,7 @@ export const environment = {
   test: false,
   baseHrefPrefix: '',
   useHashRouting: false,
-  googleClientId: '135765367152-pq4jhd3gra10jda9l6bpnmu9gqt48tup.apps.googleusercontent.com',
+  googleClientId: '656936260684-hdsihcgcdrcbeom72n56o8n5rb2mtc58.apps.googleusercontent.com',
```
bcb37, line 14: As discussed, this change shouldn't be included

bcb37 (dismissed):

bcb37 (approved):

PR #2026, opened 2024-10-04
Refine Feature Flags Root Table Pagination on Scroll

zackcl:
Reference: https://github.com/CarnegieLearningWeb/UpGrade/pull/2020#issuecomment-2393459616

This PR removes the use of the `(wheel)` directive, which was triggering the fetching of flags and displaying the loading spinner even with slight scrolling. Instead, this PR implements an `IntersectionObserver` to control when the `this.fetchFlagsOnScroll()` function is called.

Here's a screen recording demonstrating the changes:

https://github.com/user-attachments/assets/3c720baa-4e2a-44fe-8c06-12bc5312001c


danoswaltCL:
Interesting, makes sense to me.

danoswaltCL (approved):

Issue #2027, opened 2024-10-04
The filterMode (includeAll) is not applied to an imported Feature Flag


zackcl:
**Version where bug was found:**
6.0.0

**Describe the bug**
When I import a feature flag with the `filterMode` set to `includeAll`, the "Include All" slide toggle is not enabled.

**To Reproduce**
1. Create a feature flag
2. Enable the "Include All" slide toggle
3. Export the feature flag design (JSON)
4. Delete the feature flag and import the exported JSON file
5. Observe that the "Include All" slide toggle is disabled

**Expected behavior**
The `filterMode` (includeAll) should be applied to an imported feature flag

**Additional context**
Here's an example feature flag design (JSON) file with the `filterMode` set to `includeAll`: [Include All Test.json](https://github.com/user-attachments/files/17260979/Include.All.Test.json)

PR #2032, opened 2024-10-07
(Release) Refine Disabled Slide Toggles and Action Buttons on 'Reader' mode 

zackcl:
This PR fixes the unresolved [comment](https://github.com/CarnegieLearningWeb/UpGrade/pull/2002#issuecomment-2386788231) from PR #2027.

Changes include:
- Retaining colored appearance of the slide toggles when disabled
- Setting cursor to `not-allowed` when hovering over disabled elements (slide-toggle, expand button, action buttons)

https://github.com/user-attachments/assets/2d372730-4ba9-4e8f-a322-b717d8999e1f

### DIFF
--- a/frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-details-participant-list-table/common-details-participant-list-table.component.scss
+++ b/frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-details-participant-list-table/common-details-participant-list-table.component.scss
@@ -76,6 +76,46 @@
     .enable-column {
       width: 20%;
       text-align: center;
+
+      ::ng-deep .mat-mdc-slide-toggle {
+        cursor: not-allowed;
+        .mdc-switch {
+          cursor: not-allowed;
+          &--unselected {
+            .mdc-switch__track {
+              opacity: 1.0;
+              &::before {
+                background: var(--mdc-switch-unselected-track-color);
+              }
+            }
+
+            .mdc-switch__shadow {
+              background: var(--mdc-switch-unselected-handle-color);
+            }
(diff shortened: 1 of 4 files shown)


frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-details-participant-list-table/common-details-participant-list-table.component.scss
```diff
@@ -76,6 +76,46 @@
     .enable-column {
       width: 20%;
       text-align: center;
+
+      ::ng-deep .mat-mdc-slide-toggle {
```
danoswaltCL, line 80: The `::ng-deep .mat-mdc-slide-toggle` styles look like duplicate code, and it's pretty likely to get used elsewhere as we go... can you try moving this to `styles.scss` or something like that to generalize it?

danoswaltCL:
yeesh that looks like it was not easy to figure out

zackcl:
@danoswaltCL Thank you for your suggestion. I have resolved your comment. It still works the same way, but I have updated the opacity of the action buttons to 1.0, just like in the non-reader mode, because I think it more clearly communicates that those buttons are available but disabled due to user permissions. Also, it's consistent with how the slide toggles are not greyed out. I think it would make it even clearer to display a tooltip saying something like 'Not allowed for users with the "Reader" role' when hovering over these UI elements. Maybe we could implement this for v2.

https://github.com/user-attachments/assets/ebe25861-e431-469b-81fa-a08f4c4df6ce

danoswaltCL (approved):

PR #2040, opened 2024-10-10
(Release) Fix Not-Allowed Cursor Appearing on Enabled Slide Toggles

zackcl:
This PR fixes the "not-allowed" cursor appearing over the enabled slide toggles for users with the "Admin" role.

Related PR: #2032

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ab_testing_backend",
-  "version": "6.0.0-prerelease.1",
+  "version": "6.0.0-prerelease.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ab_testing_backend",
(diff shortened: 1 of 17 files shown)


bcb37 (dismissed):
This looks ok. Do we bump the prerelease suffix when it's merged?

danoswaltCL:
> This looks ok. Do we bump the prerelease suffix when it's merged?

I think the person approving can be the person on the hook to pip the version (if needed), then they can approve and merge without needing a re-approve from someone else in a follow-up PR.

I say if needed because it's really only needed the moment before we re-deploy to staging, we don't have to neccessarily worry about pipping every bugfix. I don't want that to hold up merging in code.

danoswaltCL (approved):

danoswaltCL:
(i went ahead and did it, I'll re-deploy)

PR #2049, opened 2024-10-12
updated error code for experiment endpoint and added same …

Yagnik56:
This PR updates the error code for the experiment endpoint and includes relevant changes in the swagger code.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
@@ -11,6 +11,7 @@ import {
   CurrentUser,
   Req,
   QueryParams,
+  Params,
 } from 'routing-controllers';
 import { Experiment } from '../models/Experiment';
 import { ExperimentNotFoundError } from '../errors/ExperimentNotFoundError';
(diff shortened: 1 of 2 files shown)


backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
```diff
@@ -836,15 +841,17 @@ export class ExperimentController {
...
-          JSON.stringify({ type: SERVER_ERROR.INCORRECT_PARAM_FORMAT, message: ' : id should be of type UUID.' })
-        )
-      );
+  public async one(
+    @Params({ validate: true }) { id }: IdValidator,
+    @Req() request: AppRequest
+  ): Promise<ExperimentDTO> {
+    const experiment = await this.experimentService.getSingleExperiment(id, request.logger);
+
+    if (!experiment) {
```
VivekFitkariwala, line 850: You don't need this if statement and throw the exception. We already have OnUndefined decorator which will set the proper HTTP code if undefined is returned.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -213,7 +213,11 @@ export class ExperimentService {
 
   public async getSingleExperiment(id: string, logger?: UpgradeLogger): Promise<ExperimentDTO | undefined> {
     const experiment = await this.findOne(id, logger);
-    return this.reducedConditionPayload(this.formatingPayload(experiment));
+    if (experiment) {
+      return this.reducedConditionPayload(this.formatingPayload(experiment));
+    } else {
+      return null;
```
VivekFitkariwala, line 219: You should return undefined here for the controller to throw appropriate error code. Event the function definition expects undefined.

VivekFitkariwala (approved):

PR #2054, opened 2024-10-14
fix config string replacement

shpwe:


frontend/set_build_variables.js
```diff
@@ -3,7 +3,7 @@ replace = require("replace-in-file");
 
 var replacements = [];
 var envKeys = Object.keys(envVars);
-
+var environment = process.env.ENV || 'develop';
```
danoswaltCL, line 6: maybe this will never really be undefined, but there's no `develop` environment file for that fallback, I assume that would error

shpwe: right, setting default to qa

Issue #2045, opened 2024-10-11
Replace custom Error with pre defined HttpError for reflecting message property of error


ppratikcr7:
**Version where bug was found:**
e.g "6.0.0"

**Describe the bug**
Currently, the custom Errors thrown in backend api's will remove message property in the response. This can be resolved using pre-defined HttpError.

PR #2065, opened 2024-10-17
hotfix to throw http 404 error for api calls other than init for new users

ppratikcr7:
This resolves an issue caught during QA of ticket #2045 where we got 404 httperror for hitting init api call for new users. This will filter the init api call to not throw 404 error on user not found.


backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -298,14 +298,22 @@ export class ExperimentUserService {
     return this.userRepository.save(newDocument);
   }
 
-  public async getUserDoc(experimentUserId, logger): Promise<RequestedExperimentUser> {
+  public async getUserDoc(
+    experimentUserId: string,
+    logger: UpgradeLogger,
+    api?: string
```
VivekFitkariwala, line 304: This is not a good way. The function is doing more than what the name is suggesting.

backend/packages/Upgrade/src/api/services/ExperimentUserService.ts
```diff
@@ -298,14 +298,22 @@ export class ExperimentUserService {
...
+  ): Promise<RequestedExperimentUser> {
     const experimentUserDoc = await this.getOriginalUserDoc(experimentUserId, logger);
     if (experimentUserDoc) {
       const userDoc = { ...experimentUserDoc, requestedUserId: experimentUserId };
       logger.info({ message: 'Got the user doc', details: userDoc });
       return userDoc;
     } else {
-      throw new HttpError(404, `Experiment User not found: ${experimentUserId}`);
```
VivekFitkariwala, line 308: The error should not be thrown from here

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v5.ts
```diff
@@ -179,7 +179,7 @@ export class ExperimentClientController {
   ): Promise<Pick<ExperimentUser, 'id' | 'group' | 'workingGroup'>> {
     request.logger.info({ message: 'Starting the init call for user' });
     // getOriginalUserDoc call for alias
-    const experimentUserDoc = await this.experimentUserService.getUserDoc(experimentUser.id, request.logger);
+    const experimentUserDoc = await this.experimentUserService.getUserDoc(experimentUser.id, request.logger, 'init');
```
VivekFitkariwala, line 182: Create a function in the controller which should return the experiment user doc if exist else should throw an error

VivekFitkariwala (changes requested):

danoswaltCL:
if this is a bugfix for release/6.0, it should be based from that

ppratikcr7:
> if this is a bugfix for release/6.0, it should be based from that

No, this was an enhancement kind of ticket, which was intended for proper errors codes from api's. This was intended to go into dev.

VivekFitkariwala (approved):

PR #2075, opened 2024-10-23
Feature/mooclet backend create delete experiment

danoswaltCL:
Thank you @VivekFitkariwala and @bcb37 for all the comments, this PR finalizes (hopefully) the backend changes needed for **creating** and **deleting** a Mooclet-based experiment.

- Creates `MoocletExperimentRef` model
- Creates `MoocletVersionConditionMap` model
- Extends `ExperimentDTO` as `MoocletExperimentDTO` with `moocletPolicyParameters`
- Creates a single-transaction sync on Experiment creation to create Mooclet in parallel with Upgrade Experiment
- Creates a single-transaction sync for Experiment deletion to delete Mooclet in parallel with Upgrade Experiment

Update 12/10:
- Rolls back all Mooclet resources and/or Upgrade Experiment if errors during creation
- Begins an attempt at validation of the mooclet-policy-parameters (doesn't seem to catch everything as expected but can be refined)
- Adds a `MOOCLET_POLICY_MAP` object in the `types` folder that maps supported mooclet algorithm string names to their `class-validator` validator class, so that can be used in both frontend and backend for validating policy parameters based on the algorithm name. When we want to support a new algorithm, we would add the new policy name and a validator class for it here.

```
import 'reflect-metadata';
import { ASSIGNMENT_ALGORITHM } from "../Experiment/enums";
import { MoocletTSConfigurablePolicyParametersDTO } from "./MoocletTSConfigurablePolicyParametersDTO";

const MOOCLET_POLICY_SCHEMA_MAP = {
    [ASSIGNMENT_ALGORITHM.MOOCLET_TS_CONFIGURABLE]: MoocletTSConfigurablePolicyParametersDTO
}

export {
    MOOCLET_POLICY_SCHEMA_MAP,
    MoocletTSConfigurablePolicyParametersDTO
}
```

This can be used in frontend to fill in the dropdown by taking an array of the keys: `Object.keys(MOOCLET_POLICY_SCHEMA_MAP)`

Default values can be created automatically by creating a new instance of the class-validator object:
`const defaultPolicyParams = new MoocletTSConfigurablePolicyParametersDTO()`

```
import { MOOCLET_POLICY_SCHEMA_MAP } from "upgrade_types";

const supportedMoocletPolicies = Object.keys(MOOCLET_POLICY_SCHEMA_MAP);
const moocletPolicyParameterDTO = MOOCLET_POLICY_SCHEMA_MAP[experiment.assignmentAlgorithm]
const defaultMoocletPolicyParameters = moocletPolicyParameterDTO ?? new moocletPolicyParameterDTO();

```

Follow-up PRs will address:
- modify/edit experiment flow
- assign
- mark
- logging and syncing metrics with adaptive experiment rewards


types/src/Mooclet/interfaces.ts
```diff
@@ -0,0 +1,152 @@
...
+  learner: string;
+  mooclet: number;
+  version: number;
+  policy: number;
+  value: number;
+  text: string;
+  timestamp: string;
+}
+
+export interface ExperimentCondtitionToMoocletVersionParams {
```
VivekFitkariwala, line 90: Typo

types/src/Mooclet/interfaces.ts
```diff
@@ -0,0 +1,152 @@
+/**********************************************************************
+ * types for the Mooclet services. I don't know where to put these so they're here!
+ */
+export interface MoocletProxyRequestParams {
```
VivekFitkariwala, line 4: Is this type going to be used on the frontend also? Only shared types should go here.

VivekFitkariwala (changes requested):
Small changes

danoswaltCL:
@VivekFitkariwala @bcb37 comments resolved, I will start writing out unit tests for this today if there are no more major changes requested.

backend/packages/Upgrade/.env.test
VivekFitkariwala, line 1: These changes should be removed

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
```diff
@@ -952,28 +953,23 @@ export class ExperimentController {
    */
 
   @Post()
+  @UseBefore(ValidateMoocletPolicyParametersMiddleware)
```
VivekFitkariwala, line 956: This middleware will be going to get used in multiple end point?

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
danoswaltCL: No it should just be here, is there a different way you'd recommend?

danoswaltCL:
@VivekFitkariwala 

Example of two mooclet policy parameters types. "ts_configurable" and "weighted_random" would be enums, and these would be examples of valid objects.

```
{
"ts_configurable":
    {
        "prior": {
            "success": 1,
            "failure": 1
        },
        "batch_sizes": 1,
        "max_rating": 1,
        "min_rating": 0,
        "uniform_threshold": 0,
        "tspostdiff_thresh": 0,
        "outcome_variable_name": "whatever"
    }

"weighted_random":
  {
     "probability_distribution": {
         "Arm1": 0.2,
        "Arm2": 0.8
     }
  }
}
```

danoswaltCL:
@VivekFitkariwala I incorporated your suggestions for validator. I'm still not seeing it quite work for nested properties like for the "prior" property (it accepts everything). It's actually okay to omit this value, as it will create a default new Prior() object, but if supplied, it should only allow numbers for success and failure. I'm done spinning my wheels on this problem for now though, I'd like to move forward and refine the validation later, this is good enough for now.

I also incorporated your suggestions we talked about before break in `handleCreateMoocletTransaction` so that rollback will happen on correctly for each part, hopefully the comments are clear.

```
private async handleCreateMoocletTransaction(
    manager: EntityManager,
    params: SyncCreateParams
  ): Promise<ExperimentDTO> {
    const moocletPolicyParameters = params.experimentDTO.moocletPolicyParameters;

    // create Upgrade Experiment. If this fails, then mooclet resources will not be created, and the UpGrade experiment transaction will abort
    const experimentResponse = await this.createExperiment(manager, params);

    // create Mooclet resources. If this fails, it will internally rollback any mooclet resources created, and the UpGrade experiment transaction will abort
    const moocletExperimentRefResponse = await this.orchestrateMoocletCreation(
      experimentResponse,
      moocletPolicyParameters,
      params.currentUser,
      logger
    );

    logger.info({
      message: 'Mooclet experiment created successfully:',
      moocletExperimentRef: JSON.stringify(moocletExperimentRefResponse),
    });

    // additionally, create MoocletExperimentRef and VersionConditionMaps.
    // If either of THESE fail, we will need to rollback the mooclet resources here also before aborting the UpGrade experiment transaction
    try {
      await this.saveMoocletExperimentRef(manager, moocletExperimentRefResponse);
      await this.createAndSaveVersionConditionMaps(manager, moocletExperimentRefResponse);

      experimentResponse.moocletPolicyParameters = moocletPolicyParameters;

      return experimentResponse;
    } catch (error) {
      await this.orchestrateDeleteMoocletResources(moocletExperimentRefResponse);
      throw error;
    }
  }
  ```

danoswaltCL:
@VivekFitkariwala please let me know if you have a chance to review these changes, i've included some extra unit tests for the error flow also to make sure that's working as it should.

VivekFitkariwala (approved):

danoswaltCL:
@bcb37 @zackcl @ppratikcr7 i plan on merging this end of day, let me know if there's any final things to consider, as it is a fairly big change, mostly it is contained in its own logic but experiment service and db will be affected. next thing will be integrating w/ Zack's frontend PR to finalize that as well.

Issue #1024, opened 2023-09-21
'log' calls fail in the java client library


bcb37:
The shape of response data changed from version "zero" to v1.

The log response as of v1 may have had an array added by mistake, as there are single-element arrays (example below).
We should either change the client to handle the new shape or fix the oddity in response data introduced in v1.

```
[
    {
        "0": {
            "createdAt": "2023-09-21T20:24:20.757Z",
            "updatedAt": "2023-09-21T20:39:07.795Z",
            "versionNumber": 5,
            "id": 1018,
            "uniquifier": "1",
            "timeStamp": "1970-01-01T00:00:00.000Z",
            "data": {
                "totalTimeSeconds": 10
            },
            "userId": "boof"
        }
    },
    {
        "0": {
            "createdAt": "2023-09-21T20:24:20.757Z",
            "updatedAt": "2023-09-21T20:39:07.799Z",
            "versionNumber": 6,
            "id": 1018,
            "uniquifier": "1",
            "timeStamp": "1970-01-01T00:00:00.000Z",
            "data": {
                "totalTimeSeconds": 20
            },
            "userId": "boof"
        }
    }
]
```

amurphy-cl:
Resolution: change response to 200, don't send the data object, change client so it doesn't expect the object

bcb37:
Fix by [this pr](https://github.com/CarnegieLearningWeb/UpGrade/pull/2081)

PR #2081, opened 2024-10-28
add quicktests for the clientlibs

danoswaltCL:
This improves usability of the Main.java test script, renaming it to `QuickTest.java`, matching the style of the ts `quickTest.ts` file.
 
These can be used to smoke test the client libs on the release branch on our deployed envs.

The "log" call for the Java client lib seems to have something off with its response, but since it is not a feature anybody uses at the moment, and because the response is not important at all anyway, I have commented it out, will spawn issue.


bcb37:
> The "log" call for the Java client lib seems to have something off with its response, but since it is not a feature anybody uses at the moment, and because the response is not important at all anyway, I have commented it out, will spawn issue.

Is that different from [this issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/1024)

bcb37 (approved):
This worked for me. I have a fix for the log method. Can make a pr.

danoswaltCL:
@bcb37 Yeah do you want to commit to this branch if you know a fix for that? could easily be same basic issue. I was being lazy and didn't look into it very deeply.

danoswaltCL:
cool, works for me if you want to approve, i'll pr to merge down to dev also.

bcb37 (approved):

Issue #2090, opened 2024-11-01
Errors in data export, Payload appears to copy data from Condition and duplicate rows


amurphy-cl:
Version 6.0

- [x] In new data export, the filed for Payload seems to copy data from Condition, rather than including the actual payload. Expected behavior is that the Payload field will represent the experiment payload. 

- [] Also, rows are duplicated (see `markExperimentPointTime` fractional seconds). 

Experiment export attached. Experiment in stage is https://upgrade.qa-cli.com/home/detail/bdb0b762-04ce-45db-a929-aaa9fc85e264  name "TC-RB-ISP-test". 

[amurphy@carnegielearning.com_simpleExport2024-11-04T15_35_40.651Z.csv](https://github.com/user-attachments/files/17648357/amurphy%40carnegielearning.com_simpleExport2024-11-04T15_35_40.651Z.csv)

ppratikcr7:
Observed multiple duplicate rows due to all site target's mapped to condition payloads instead of the actual mapped site target.

PR #2095, opened 2024-11-06
Bugfix for experiment data export storing condition name in condition payload column

ppratikcr7:


zackcl:
@ppratikcr7 Could you please explain how/where this PR resolves the 2 issues mentioned in #2090?

1. In new data export, the filed for Payload seems to copy data from Condition, rather than including the actual payload. Expected behavior is that the Payload field will represent the experiment payload. 

2. Also, rows are duplicated (see `markExperimentPointTime` fractional seconds). 

Also, does anyone know a good way to test/validate this? I'm not sure how to properly QA this. I don't think I can send experiment data to my email when running locally (I get the `CredentialsProviderError: Could not load credentials from any providers` error).

If anyone is confident in approving this PR, please feel free to do so.

backend/packages/Upgrade/src/api/services/AnalyticsService.ts
```diff
@@ -199,24 +206,26 @@ export class AnalyticsService {
             enrollmentCompleteDate: item.enrollmentCompleteDate,
             details: [],
           };
+          experimentMap.set(item.experimentId, experiment);
           acc.push(experiment);
         }
-        experiment.details.push({
+        // add the design details for the experiment for each experimentQueryResult item
+        experimentMap.get(item.experimentId)?.details.push({
```
bcb37, line 213: If this was overwriting the details for existing experiments, can't we fix that by simply closing the `if` statement on line 221 (old 212), rather than creating the map structure?

ppratikcr7: ye we had that in the old code, I felt using a Map would be more efficient. You want me to revert this part? Wont make much of a difference though.

PR #2097, opened 2024-11-07
close the modals if auth fails

Yagnik56:
This PR will fix the issue of Modal staying open despite auth failure and UI moves to the login screen.


zackcl:
It works fine as expected!

PR #2109, opened 2024-11-12
fix header and moved overflow to table body to make it scrollable

Yagnik56:
This PR fixed the header as static and moved overflow to the table body to make it scrollable.
image


zackcl (dismissed):

zackcl:
Note: I've autoFocus: true has been re-added after merge-conflict.

zackcl (approved):

PR #2131, opened 2024-11-25
update filter mode when ALL row is removed in include table

Yagnik56:
This PR will fix the issue of filter mode not updating when the include ALL row is deleted in the include participants table.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
@@ -304,6 +304,9 @@ export class ExperimentParticipantsComponent implements OnInit {
   }
 
   removeMember1(groupIndex: number) {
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
```diff
@@ -304,6 +304,9 @@ export class ExperimentParticipantsComponent implements OnInit {
   }
 
   removeMember1(groupIndex: number) {
+    if (groupIndex === 0 && this.members1.controls.at(0).get('type').value === 'All') {
```
danoswaltCL, line 307: Are you sure these chained calls will never catch a weird `undefined` or null case? 

`this.members1.controls.at(0).get('type').value === 'All')`
 `this.participantsForm.get('inclusionCriteria').setValue(INCLUSION_CRITERIA.INCLUDE_SPECIFIC)`

PR #2141, opened 2024-11-26
Hotfix: fix export data

bcb37:
This pr:
- removes the queries to 'monitored_decision_point' and 'monitored_decision_point_log' and replaces multiple rows with different `MarkExperimentPointTime` values with a single row with `MarkExperimentPointTime` coming from the enrollment table
- avoids including metrics columns that do not correspond to queries defined in the experiment
- fixes the display of metric keys

### DIFF
--- a/backend/packages/Upgrade/src/api/repositories/QueryRepository.ts
+++ b/backend/packages/Upgrade/src/api/repositories/QueryRepository.ts
@@ -5,23 +5,6 @@ import repositoryError from './utils/repositoryError';
 
 @EntityRepository(Query)
 export class QueryRepository extends Repository<Query> {
-  public async findNamesByExperimentId(experimentId: string): Promise<string[]> {
-    const queries = await this.createQueryBuilder('query')
-      .select(['name'])
-      .where('"experimentId" = :experimentId', { experimentId })
-      .getRawMany()
-      .catch((errorMsg: any) => {
-        const errorMsgString = repositoryError(
-          'QueryRepository',
-          'findNamesByExperimentId',
-          { experimentId },
-          errorMsg
-        );
-        throw errorMsgString;
-      });
-    return queries.map((query) => query.name);
-  }
-
   public async deleteQuery(id: string, entityManager: EntityManager): Promise<void> {
     await entityManager
       .createQueryBuilder()
--- a/backend/packages/Upgrade/src/api/services/AnalyticsService.ts
+++ b/backend/packages/Upgrade/src/api/services/AnalyticsService.ts
@@ -234,11 +234,6 @@ export class AnalyticsService {
       }
 
       const queryData = await this.logRepository.getLogPerExperimentQuery(experimentId);
-      // query name id mapping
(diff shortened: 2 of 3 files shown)


ppratikcr7 (commented):
@bcb37 So wanted to confirm one thing, you are now showing the attribute metric (`timeSeconds`) instead of the class (`masteryWorkspace`) for a grouped metric in the metrics column, right? Should we show the complete grouped metric (Class -> Key -> Attribute)? I have attached screenshot for other to see the difference. The old export gives 2 rows for each of the 2 students and the new one as per this PR gives 1 row for each of the 2 students. This is working perfectly as expected. 

Screenshot 2024-11-27 at 6 38 28 PM


Also, didn't get the 2nd point of this PR description. Which metric column will not be shown? We already show the metrics that are used in the experiment.

bcb37:
Also, didn't get the 2nd point of this PR description. Which metric column will not be shown? We already show the metrics that are used in the experiment. Can you please elaborate this scenario a bit. Thanks
 
@ppratikcr7  it was including metrics columns from other experiments because the metrics list wasn't restricted to just the metrics in the current experiment. So I fixed that.
The screenshot is from a prod export of an experiment that does not have queries "Percent Graduated", "Swap", etc.

Screenshot 2024-11-27 at 2 42 34 PM

ppratikcr7 (dismissed):

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
```diff
@@ -467,39 +467,13 @@ export class AnalyticsRepository extends Repository<AnalyticsRepository> {
       .addGroupBy('individualEnrollment.experimentId')
       .addGroupBy('individualEnrollment.conditionId')
       .addGroupBy('individualEnrollment.partitionId')
+      .addGroupBy('individualEnrollment.createdAt')
```
VivekFitkariwala, line 470: We don't need to join tables in the query as we are only getting the data from the `individualEnrollment`  table so do we need so many `addGroupBy`?
```
.leftJoin(
        DecisionPoint,
        'decisionPointData',
        'decisionPointData.experimentId = individualEnrollment.experimentId AND decisionPointData.id = individualEnrollment.partitionId'
      )
```
This part can be removed as the partition data is coming from the `experimentsData`.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/repositories/AnalyticsRepository.ts
bcb37: @VivekFitkariwala  This is necessary because we need the **names** of the site and target for the 'Site' and 'Target' columns. Those are not in the data defined in the the type `ExperimentDetailsForCSVData`. We only get the decisionPointId. To get the names of the site and target we have to join on the DecisionPoint table.

VivekFitkariwala (approved):

PR #2143, opened 2024-11-28
Allow importing experiments with old payload 

Yagnik56:
This PR allows experiments with versions below 5.3.0 to be imported with older payload structures.

### DIFF
--- a/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
+++ b/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
@@ -182,7 +182,7 @@ export class PartitionValidator {
   public excludeIfReached: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
```diff
@@ -201,6 +201,25 @@ class ConditionPayloadValidator {
   public decisionPoint?: string;
 }
 
+class OldConditionPayloadValidator {
```
VivekFitkariwala, line 204: You can create a base abstract class for `OldConditionPayloadValidator` and `ConditionPayloadValidator`.
It is not a big deal.

VivekFitkariwala (approved):

VivekFitkariwala (approved):

PR #2145, opened 2024-12-02
list endpoint and segment migration

bcb37:
The return value for the /delete segment endpoint (along with the service and repository methods in its call stack) was falsely typed as `Promise<Segment>` . It was actually returning `Promise<Segment[]>`, which the front end was set up to expect. Now the typing is correct (and consistent with the swagger annotation). This pr adjusts the front end code accordingly.


backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -308,6 +328,20 @@ export class SegmentService {
...
   public async deleteSegment(id: string, logger: UpgradeLogger): Promise<Segment> {
     logger.info({ message: `Delete segment by id. segmentId: ${id}` });
+    const segmentDoc = await this.segmentRepository.findOne({
+      where: { id: id },
+      relations: ['individualForSegment', 'groupForSegment', 'subSegments'],
+    });
+    if (!segmentDoc) {
+      throw new Error(SERVER_ERROR.QUERY_FAILED);
+    }
+    await Promise.all(
```
VivekFitkariwala, line 338: These should happen in a transaction

VivekFitkariwala (changes requested):

Issue #2147, opened 2024-12-03
417 Error in "Vanderbilt" studies


danoswaltCL:
**Version where bug was found:**
6

**Describe the bug**
Referencing this slack thread: https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1732071862779939
Here is the experiment:

[upgrade-280713-63033db33a3a.json](https://github.com/user-attachments/files/17986487/upgrade-280713-63033db33a3a.json)

This is an example user that apparently was able to cause this error:

```
Group not defined for experiment User: {
  "createdAt": "2022-10-20T16:32:56.503Z",
  "id": "117e6fc0-2cc1-470b-b30c-ad616719b91e",
  "requestedUserId": "mx-117e6fc0-2cc1-470b-b30c-ad616719b91e-searching_for_patterns_na_alg1_24",
  "group": {
    "schoolId": [
      "ppxsx5fb48763a3734"
    ],
    "classId": [
      "zsqbfdxtmmnzzwux06li7g"
    ],
    "instructorId": [
      "29bf77ad-06b5-4d67-a89f-ebfa12de06e4"
    ]
  },
  "workingGroup": {
    "classId": "zsqbfdxtmmnzzwux06li7g",
    "schoolId": "ppxsx5fb48763a3734",
    "instructorId": "29bf77ad-06b5-4d67-a89f-ebfa12de06e4"
  },
  "experiment": []
}
```

**To Reproduce**
I can't! I tried and tried. I'm baffled. I'm not sure I understand everything going on this part of [this code](https://github.com/CarnegieLearningWeb/UpGrade/blob/dev/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts#L799),

danoswaltCL:
Well I've reproduced it now... a lot, I was misreading what I was expecting because this "417" error actually has nothing to do with an HTTP response. It only appears in the logs and db, and worse, it appears to be doing this for every single assign request for a user who is not globally excluded and has not previously enrolled. I don't know what the intent was, this error was probably supposed to wrapped in some more specific logical context, but it's just nonsensically logging this and inserting it into the error table for nearly every user in assign-prog for any group assignment that is currently enrolling, as far as I can tell. I guess we have not seen this previously because group was not used in prod?

Regardless, it should be cut off from the db, and luckily it appears has no bearing on the actual assignments being returned, which seem to all be working as they should.

[...]

PR #2148, opened 2024-12-04
remove errorneous error logging and db inserts in group experiment assign call

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/issues/2147

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -863,36 +863,11 @@ export class ExperimentAssignmentService {
       });
     }
 
-    // exclude experiments which are not previously assigned and throw error
+    // exclude experiments which are not previously assigned
     const experimentToExclude = experiments.filter((experiment) => {
       return groupExperimentAssignedIds.indexOf(experiment.id) === -1;
     });
 
-    const experimentToExcludeIds = experimentToExclude.map((experiment) => experiment.id);
-
-    // throw error user group not defined and add experiments which are excluded
-    experimentToExclude.forEach(({ id, name }) => {
-      logger.error({
-        message: `Experiment Id: ${id},
-      Experiment Name: ${name},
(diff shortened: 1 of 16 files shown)


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -863,36 +863,11 @@ export class ExperimentAssignmentService {
...
-    // exclude experiments which are not previously assigned and throw error
+    // exclude experiments which are not previously assigned
     const experimentToExclude = experiments.filter((experiment) => {
       return groupExperimentAssignedIds.indexOf(experiment.id) === -1;
     });
 
-    const experimentToExcludeIds = experimentToExclude.map((experiment) => experiment.id);
-
-    // throw error user group not defined and add experiments which are excluded
-    experimentToExclude.forEach(({ id, name }) => {
```
VivekFitkariwala, line 874: We should keep the logger and comment out the insertion in the database.

VivekFitkariwala (dismissed):

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
danoswaltCL: it's not just the insert that is a problem, when i was testing this yesterday, the log and insert both are called for no reason for most users, regardless of if they have an invalid group or not.

danoswaltCL: https://github.com/CarnegieLearningWeb/UpGrade/issues/2147#issuecomment-2515876690 i can demonstrate this in my local in mtg

danoswaltCL: https://github.com/CarnegieLearningWeb/UpGrade/issues/2149

bcb37 (dismissed):

bcb37 (approved):

PR #2161, opened 2024-12-12
Fix/solving 500 errors

RidhamShah:
Cheery-pick some commits to throw perfect 404 User not found errors instead of throwing 500.


danoswaltCL:
@amurphy-cl see above comment, should we merge and deploy to prod, or are we okay with just putting in dev?

PR #2163, opened 2025-01-02
Feature/mooclet validator

VivekFitkariwala:
To test this PR make a `Post` request to `/api/experiments` with `moocletPolicyParameters` for testing validation.

```
"assignmentAlgorithm": "ts_configurable",
"moocletPolicyParameters": {
    "assignmentAlgorithm": "ts_configurable",
    "batch_size": 1
}
```
Based on the type of `"moocletPolicyParameters"."assignmentAlgorithm"` the associated validator will run. The discriminator will already remove `"moocletPolicyParameters"."assignmentAlgorithm"` this property in the `moocletPolicyParameters`.
From frontend, we need to add `assignmentAlgorithm` inside `moocletPolicyParameters` for validator to work correclty.

### DIFF
--- a/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
+++ b/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
@@ -3,6 +3,7 @@ import {
   IsArray,
   IsBoolean,
   IsDateString,
+  IsDefined,
   IsEnum,
   IsInt,
   IsNotEmpty,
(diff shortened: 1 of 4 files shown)


danoswaltCL:
ok, i have tried this and have seen it work when adding "assignmentAlgorithm" to the mooclet parameters. I tried to see if I could send without adding that property, there were some suggestions I found, but it really does seem like it needs to be there.

One other thing is that surprisingly this will still allow a moocletPolicyParams object pass without validation if the property isn't present. Any ideas on that?

VivekFitkariwala:
If none of the discriminators matches it will use `MoocletPolicyParameters` class for the validation. I have added appropriate validation on `assignmentAlgorithm` for this to work correctly.

danoswaltCL:
closing, I've incorporated these ideas in the branch 👍

PR #2169, opened 2025-01-17
add assign mooclet proxy to assignment service

danoswaltCL:
This adds assignment proxy to experimentAssignmentService for Mooclet experiments.

Experiments that are created with a Mooclet assignment algorithm will generate two entities in the db to map the relationship between the Upgrade Experiment entities and its Mooclet counterparts, "MoocletExperimentRef" and "MoocletVersionConditionMap", which is joined to the MoocletExperimentRef document as an array of direct relationships between Mooclet "versions" and UpGrade "conditions".

On `/assign`, if a user has not already been enrolled in the mooclet experiment they have been assigned, a call will be proxied to the Mooclet API to be given a new "version" (condition).

When fetching a new version, the MoocletExperimentRef will be fetched via UpGrade experimentId to find the condition that this version is mapped to, which is what is returned in the response as normal Upgrade condition assignment.

When this assignment is marked, the user will be enrolled as normal in the UpGrade experiment and subsequent calls to /assign will not reach out to the Mooclet server.

Note! the assignment retrieved from Mooclet for ts_configurable is not constant like it is in upgrade, you could get any of the versions involved on each call until it is marked, so this is something we should discuss.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -1590,16 +1590,14 @@ export class ExperimentAssignmentService {
   private async getConditionFromMoocletProxy(experiment: Experiment, user: ExperimentUser) {
     const userId = user.id;
 
-    const condition = await this.moocletExperimentService.getConditionFromMoocletProxy(experiment, userId);
-
-    return condition;
+    return await this.moocletExperimentService.getConditionFromMoocletProxy(experiment, userId);
   }
 
   private assignRandom(
     experiment: Experiment,
     user: ExperimentUser,
     enrollmentCount?: { conditionId: string; userCount: number }[]
-  ) {
+  ): ExperimentCondition {
     const randomSeed =
       experiment.assignmentUnit === ASSIGNMENT_UNIT.INDIVIDUAL ||
       experiment.assignmentUnit === ASSIGNMENT_UNIT.WITHIN_SUBJECTS
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -68,6 +68,8 @@ import { UserStratificationFactorRepository } from '../repositories/UserStratifi
 import { UserStratificationFactor } from '../models/UserStratificationFactor';
 import { RequestedExperimentUser } from '../controllers/validators/ExperimentUserValidator';
 import { In } from 'typeorm';
+import { env } from '../../env';
+import { MoocletExperimentService } from './MoocletExperimentService';
 @Service()
 export class ExperimentAssignmentService {
   constructor(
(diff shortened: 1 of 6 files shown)


danoswaltCL:
@bcb37 this is ready for review but I am asking the Mooclet team folks if sending the "learner" is needed when getting condition. I can see that it will add `null` for `learner id` in the `value` table when not sending it, and it adds the learner id when i do, but I can't tell if it actually affects anything (for ts_configurable).

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1559,11 +1563,43 @@ export class ExperimentAssignmentService {
...
+    const condition = await this.moocletExperimentService.getConditionFromMoocletProxy(experiment, userId);
+
+    return condition;
+  }
+
   private assignRandom(
     experiment: Experiment,
     user: ExperimentUser,
     enrollmentCount?: { conditionId: string; userCount: number }[]
-  ): ExperimentCondition {
```
bcb37, line 1566: Is there a way to keep the return type here?

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -1559,11 +1563,43 @@ export class ExperimentAssignmentService {
...
+      return this.getConditionFromMoocletProxy(experiment, user);
+    } else {
+      return this.assignRandom(experiment, user, enrollmentCount);
+    }
+  }
+
+  private async getConditionFromMoocletProxy(experiment: Experiment, user: ExperimentUser) {
+    const userId = user.id;
+
+    const condition = await this.moocletExperimentService.getConditionFromMoocletProxy(experiment, userId);
```
bcb37, line 1593: Do we want to create this variable just to return it?

backend/packages/Upgrade/src/api/services/MoocletDataService.ts
```diff
@@ -199,21 +199,19 @@ export class MoocletDataService {
     return response;
   }
 
-  /* not yet implemented */
-
-  // public async getVersionForNewLearner(moocletId: number, userId: string) {
-  //   const endpoint = `/mooclet/${moocletId}/run?learner=${userId}`;
+  public async getVersionForNewLearner(moocletId: number, userId: string) {
```
bcb37, line 202: Is the return type (`Promise<MoocletVersionResponseDetails>`) inferred here?

bcb37 (commented):
This looks ok to me. Just a couple questions.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
danoswaltCL: yes probably, i can add that back, not sure why removed

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
danoswaltCL: we don't have to, I guess I like to be explicit vs saving a couple of characters... it's also slightly easier to debug later if need be, all the data is assembled there in the local scope to spy or log out.

backend/packages/Upgrade/src/api/services/MoocletDataService.ts
danoswaltCL: i can add a tyep.

danoswaltCL:
Got a response and it seems that we not only should send the learner for this /mooclet/:id/run call, but also when sending a "reward", which is not something I was testing. I will need to check that out next week to see what the affect is, but I think we do in fact want what is currently coded up after all.

bcb37 (approved):

PR #2186, opened 2025-01-28
Angular 19 update

bcb37:
Update from angular 17 to 19:

Changes from running automated migrations for both updates (17->18 & 18->19):
- syntax changes in `CoreModule` and `AppModule`
- Modifications to configs: 'angular.json' and 'tsconfig.json' (18->19)
- Addition of `standalone:boolean` attribute to components

Manual changes after 18->19 update:
- addition of 'm2' prefix to some sass functions and variables (migration to newer material version)
- moving `AuthModule` import from `CoreModule` to `AppModule` - This seemed to fix a circular dependency issue. It doesn't seem like a simple task to keep it in `CoreModule`. Is there strong a reason we'd want to?


frontend/tsconfig.json
```diff
@@ -7,13 +7,12 @@
     "declaration": false,
     "moduleResolution": "node",
     "strict": false,
+    "esModuleInterop": true,
```
VivekFitkariwala, line 10: It might be a good idea to remove `import * ` from the code.
In a separate PR.

frontend/projects/upgrade/src/dev-tools/key-value-bubble/key-value-bubble.component.ts
```diff
@@ -4,10 +4,11 @@ import { IKeyValueBubbleStyles } from './key-value-bubble.model';
 import { KeyValueBubbleService } from './key-value-bubble.service';
 
 @Component({
```
VivekFitkariwala, line 6: @bcb37 , this change has introduced a lint problem and a similar problem in all the files. Can you check?

danoswaltCL: omg we should just delete this whole dev-tools module, I forgot it existed.

frontend/tsconfig.json
bcb37: Probably a good idea at some point. In the frontend code, there are no instances of `import *` for 3rd party deps, only modules defined locally.

frontend/package-lock.json
```diff
@@ -89,112 +89,105 @@
       }
     },
     "node_modules/@angular-devkit/architect": {
-      "version": "0.1703.11",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.1703.11.tgz",
-      "integrity": "sha512-YNasVZk4rYdcM6M+KRH8PUBhVyJfqzUYLpO98GgRokW+taIDgifckSlmfDZzQRbw45qiwei1IKCLqcpC8nM5Tw==",
+      "version": "0.1901.4",
+      "resolved": "https://cli-467155500999.d.codeartifact.us-east-1.amazonaws.com/npm/cli-npm-artifacts/@angular-devkit/architect/-/architect-0.1901.4.tgz",
```
VivekFitkariwala, line 93: It should be from npm registry

bcb37: That's a good point. We may have to get permission to keep the deps there, but it certainly makes sense in an open-source repo.

bcb37: Yep. Reverting this as per discussion in SRE meeting.

VivekFitkariwala: I am able to test it locally

VivekFitkariwala:
_moving AuthModule import from CoreModule to AppModule - This seemed to fix a circular dependency issue_
@bcb37  Can you give more detail about this?

VivekFitkariwala (approved):

danoswaltCL (approved):
seems like everything works normally as far as I can tell

PR #2201, opened 2025-02-05
feature flag caching on any change to flags or lists, no wait on exposures save

danoswaltCL:
This will cache the feature flags when fetched by context.

- The flag cache is cleared by context when any flag is added, deleted, or updated in any way including enabling/disabling.
- On list add, delete, or modify, including enabling/disabling, the flag cache will clear all contexts.
- Also, I removed the `await` when calling exposures.

I tested as well as I could in local with JMeter load tests, it certainly does make a difference but it's not really a good representation of a real test on staging.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
+++ b/backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
@@ -33,6 +33,7 @@ import {
   FEATURE_FLAG_LIST_FILTER_MODE,
   FEATURE_FLAG_LIST_OPERATION,
   ListOperationsData,
+  CACHE_PREFIX,
 } from 'upgrade_types';
 import { UpgradeLogger } from '../../lib/logger/UpgradeLogger';
 import { FeatureFlagValidation } from '../controllers/validators/FeatureFlagValidator';
@@ -49,6 +50,7 @@ import { User } from '../models/User';
 import { diffString } from 'json-diff';
 import { SegmentRepository } from '../repositories/SegmentRepository';
 import { ExperimentAuditLog } from '../models/ExperimentAuditLog';
+import { CacheService } from './CacheService';
 
 @Service()
 export class FeatureFlagService {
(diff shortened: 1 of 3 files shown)


backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
```diff
@@ -485,6 +511,7 @@ export class FeatureFlagService {
     logger: UpgradeLogger
   ): Promise<Segment> {
     await this.createDeleteListAuditLogs([segmentId], filterType, currentUser);
+    this.cacheService.resetPrefixCache(CACHE_PREFIX.FEATURE_FLAG_KEY_PREFIX);
```
VivekFitkariwala, line 514: add `await`
Or you can use `Promise.all` to run things parallelly

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
VivekFitkariwala, line 1: Add `await` for the cache service calls.
Logically, it won't make any difference, but it is the correct way.

VivekFitkariwala, line 1: The cache clear needs to be added in the `update` function.

backend/packages/Upgrade/test/unit/services/FeatureFlagService.test.ts
```diff
@@ -154,6 +155,14 @@ describe('Feature Flag Service Testing', () => {
       providers: [
         DataSource,
         FeatureFlagService,
+        {
+          provide: CacheService,
+          useValue: {
+            delCache: jest.fn().mockResolvedValue(undefined),
+            resetPrefixCache: jest.fn().mockResolvedValue(undefined),
+            wrap: jest.fn().mockImplementation((key, cb) => cb()),
+          }
```
VivekFitkariwala, line 164: `Eslint` problem here. Need a `,`.
Do you have the Eslint extension installed?

VivekFitkariwala (changes requested):
Small changes.

backend/packages/Upgrade/src/api/services/FeatureFlagService.ts
danoswaltCL: it is there, it's in the `updateFeatureFlagInDB` part

backend/packages/Upgrade/test/unit/services/FeatureFlagService.test.ts
danoswaltCL: sure, the eslint job that runs on commit hook didn't catch  and it builds and passes everything, so I didn't see it. i think we are not all getting consistent linting action going on if vscode and our commit hooks aren't agreeing.

danoswaltCL:
@VivekFitkariwala updated

VivekFitkariwala (approved):

PR #2205, opened 2025-02-07
create mooclet when a mooclet experiment is imported

bcb37:


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -509,6 +509,16 @@ export class ExperimentService {
     user: UserDTO,
     logger: UpgradeLogger
   ): Promise<ValidatedExperimentError[]> {
+    const { experiments, validatedExperiments } = await this.verifyExperiments(experimentFiles, logger);
```
VivekFitkariwala, line 512: This function is no longer needed.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -136,6 +138,51 @@ export class MoocletExperimentService extends ExperimentService {
...
+    experiments: ExperimentDTO[],
+    currentUser: UserDTO,
+    logger: UpgradeLogger
+  ): Promise<Experiment[]> {
+    const createdExperiments = [];
+    for (const experiment of experiments) {
+      try {
+        if ('moocletPolicyParameters' in experiment) {
+          if (!env.mooclets?.enabled) {
+            throw new BadRequestError(
```
VivekFitkariwala, line 162: If we throw an error, even valid experiments will not created.
This should be part of the `verifyExperiment` and the experiment should be skipped.
Users should see this problem before the actual import.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
```diff
@@ -1270,7 +1270,7 @@ export class ExperimentController {
     @CurrentUser() currentUser: UserDTO,
     @Req() request: AppRequest
   ): Promise<ValidatedExperimentError[]> {
-    return this.experimentService.importExperiment(experiments, currentUser, request.logger);
+    return this.moocletExperimentService.syncImportExperiment(experiments, currentUser, request.logger);
```
danoswaltCL, line 1273: `moocletExperimentService` is the entrypoint for all imports? That feels a little funny, is there something that makes `experimentService.importExperiment` unable to accommodate delegating control to `moocletExperimentService` for mooclet imports only, while remaining the same for existing imports?

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -136,6 +136,45 @@ export class MoocletExperimentService extends ExperimentService {
...
+    return validatedExperiments;
+  }
+
+  private async syncAddBulkExperiments(
+    experiments: ExperimentDTO[],
+    currentUser: UserDTO,
+    logger: UpgradeLogger
+  ): Promise<Experiment[]> {
+    const createdExperiments = [];
+    for (const experiment of experiments) {
```
VivekFitkariwala, line 156: Can `for` loop be replaced with `map` so the creation will be parallel?

VivekFitkariwala (changes requested):
One more small change. Sorry missed it earlier.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
bcb37: Probably, I'll try

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
bcb37: I wasn't able to import the `MoocletExperimentService` into the `ExperimentService`, since the former 'extends' the latter. Any ideas about a delegation pattern?

bcb37: We'll have to follow the same pattern with exports because we have to call `moocletExperimentService.getPolicyParametersByExperimentId()`, which we can't do from `experimentService`

danoswaltCL: i didn't get a chance to look at this today! I feel like I ran into something similar at some point, I will take a look tomorrow to see if there's a straightforward alternative.

danoswaltCL:
Okay, I think I have a good way of doing this, rather than dump all the code here I pushed up a branch with what I was playing around with: [example branch](https://github.com/CarnegieLearningWeb/UpGrade/tree/example-mooclet-imports-delegation)

The gist is that the core logic that has nothing to do with Mooclet stuff can all stay in the experiment service, but if mooclet is enabled, we'll detour briefly into the moocletExperimentService (from the ExperimentController to avoid the circular import issues). It will take care of mooclet experiment creation business only, and then it will call `super` to return control back to ExperimentService to handle the remaining "normal" experiments.

I tested it out with importing a mix of mooclet and native experiments, seems to work, let me know if this is any better or just a rearrangement. I think it keeps the services clearer in a similar pattern as is used elsewhere.

```typescript
@Post('/import')
  public async importExperiment(
    @Body({ validate: true })
    experiments: ExperimentFile[],
    @CurrentUser() currentUser: UserDTO,
    @Req() request: AppRequest
  ): Promise<ValidatedExperimentError[]> {
    const { experiments: experimentList, validatedExperiments } = await this.experimentService.verifyExperiments(
      experiments,
      request.logger
    );

    if (env.mooclets.enabled) {
      await this.moocletExperimentService.syncAddBulkExperiments(experimentList, currentUser, request.logger);
    } else {
      await this.experimentService.addBulkExperiments(experimentList, currentUser, request.logger);
    }

    return validatedExperiments;
  }

```

danoswaltCL:
It looks like currently `addBulkExperiments` is not using the experiments return value in that import function in `dev` anyway, but the reason it is there is that `addBulkExperiments` does apparently also get used by the `/batch` function, which looks like it is there for testing and does use the returned experiments, which I was not aware of.

That would imply needing to figure out what is needed to support that function regardless of the changes I proposed... if necessary, I don't know the usage of this and/or if Mooclet experiments are ever going to be mocked in the testing module that uses this method.

As for trying to keep the mooclet feature services separate even if it feels like jumping through hoops presently, my pitch is this: we have features like "within-subjects" experiments that are intertwined in the core code in a non-modular way, and although it has never been turned on, it has a huge presence in the codebase that has to be worked around, maintained, mocked in unrelated testing, etc, and has at times caused surprising bugs to pop up.

Mooclets may indeed not go to prod for some time, if ever, and my hope is that while it is off, we won't have to look at, test it, or worry about that code at all. And when it's on and there's a new feature to add or a bug to fix related to mooclet, it will be obvious where to do that work.

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -570,52 +560,51 @@ export class ExperimentService {
...
-        LOG_TYPE.EXPERIMENT_DESIGN_EXPORTED,
-        { experimentName: experiment.name },
-        user
-      );
-      return this.reducedConditionPayload(this.formatingPayload(this.formatingConditionPayload(experiment)));
-    });
-
-    return formattedExperiments;
-  }
+  // public async exportExperiment(user: UserDTO, logger: UpgradeLogger, experimentIds?: string[]): Promise<Experiment[]> {
```
danoswaltCL, line 566: comments can be removed

danoswaltCL:
this looks good to me at a glance, I can take a closer look tomorrow so we can try to get this squared away before your vacation

danoswaltCL (approved):

Issue #1934, opened 2024-09-12
mooclet backend: metric logs sync to rewards


danoswaltCL:

zackcl:
Here is the example design of the read-only Mooclet Reward Metric section:

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/07cfc2a5-8b2b-418b-80b4-118f5219f60d" />

danoswaltCL:
@zackcl looks good, I created a separate story for the stepper part so it can be worked in parallel #2203

PR #2207, opened 2025-02-08
add reward metric and query on mooclet create

danoswaltCL:
This adds experiment specific reward metric and a query for percent success on mooclet experiment creation.

On experiment delete, this experiment-specific reward metric will also be deleted.

#1934


danoswaltCL:
@zackcl this has been updated, frontend will need to send `rewardMetricKey`. There are tests failing, so I'm trying to address those or try to figure out if it's just me? It seems to pass the QA checks...

@VivekFitkariwala @bcb37 please take a look, this one is a bit of a blocker for the rest of the mooclet stuff until it's in. ty ty

zackcl (approved):

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
```diff
@@ -843,6 +843,7 @@ export class ExperimentController {
       } else {
         const policyParametersResponse = await this.moocletExperimentService.getPolicyParametersByExperimentId(id);
```
VivekFitkariwala, line 844: You are making two queries to get the same mooclet experiment and extracting `policyParameter` from one and `rewardMetricKey` from another

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
danoswaltCL: They both pull a reference to the same moocletExperimentRef that's true. But  `getPolicyParametersByExperimentId(id);` also fetches an updated copy of the policyParameters from the mooclet server.

Would you like me to combine these so we just make the one query to get moocletExperimentRef?

danoswaltCL: @VivekFitkariwala I updated this to combine

PR #2225, opened 2025-02-14
Disallow Empty Metrics Rows For Experiment Creation/Update

### LINKED ISSUE #2224: Empty rows in the Metrics table should not be allowed to proceed
6.1.0

**Describe the bug**
When creating or editing an experiment, empty rows in the Metrics table are currently allowed to proceed (Unlike how the Include/Exclude tables work). This causes an error during experiment creation/update because the `queries` field contains a null element (`[null]`) in the experiment data.

**To Reproduce**
1. Click "Add Experiment"
2. Navigate to the Metrics step
3. Click "Add Metric" to add a new row (or two rows)
4. Click "Next" to proceed to the next step with the rows being empty
5. Create the experiment
6. Observe error in the console

zackcl:
Resolves #2224 

When "Next" is clicked while empty rows exist:
Screenshot 2025-02-14 at 10 38 25 PM


danoswaltCL (approved):
looks good to me

PR #2229, opened 2025-02-15
⬆️ Updating node version

VivekFitkariwala:
Removing the `esm` flag from ts-node and convert `.mts` to `.ts`

### DIFF
--- a/backend/packages/Upgrade/commands/banner.ts
+++ b/backend/packages/Upgrade/commands/banner.ts
@@ -1,12 +1,12 @@
-import chalk from 'chalk';
 import figlet from 'figlet';
 
 figlet.text(process.argv[2], (error: any, data: any) => {
   if (error) {
     return process.exit(1);
   }
-
-  console.log(chalk.blue(data));
-  console.log('');
-  return process.exit(0);
+  import('chalk').then((chalk) => {
+    chalk.default.blue(data);
+    console.log('');
+    return process.exit(0);
+  });
 });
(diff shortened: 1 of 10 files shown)


backend/packages/Upgrade/commands/banner.ts
```diff
@@ -1,12 +1,12 @@
...
   }
-
-  console.log(chalk.blue(data));
-  console.log('');
-  return process.exit(0);
+  import('chalk').then((chalk) => {
+    chalk.default.blue(data);
+    console.log('');
+    return process.exit(0);
+  });
```
bcb37, line 11: `figlet.text()` can return `void`. For me the banner worked if I changed the above to the following:
```
if (error) {
    process.exit(1);
  } else {
    import('chalk').then((chalk) => {
      console.log(chalk.default.blue(data));
      process.exit(0);
    });
  }
```

ppratikcr7:
@VivekFitkariwala I was trying to confirm the fix, but I don't see the changes for reverting zack's volume related updates and removing only a specific folder in the last commit. Am I missing something? 

Also, when I am trying to run the docker, I see a hashed volume instead of our defined `upgradedbbdata` as we had earlier. 

Screenshot 2025-02-20 at 1 11 21 PM

Also, I see the backend crash with an error in ErrorHandlerMiddleware. Do anyone else see this?

Screenshot 2025-02-20 at 1 12 58 PM

zackcl:
@ppratikcr7 I also see a hashed volume instead of our defined upgradedbbdata, but I didn't get that error in the backend. The app seems to be working fine for me (I tried creating an experiment and a feature flag).

ppratikcr7:
> @ppratikcr7 I also see a hashed volume instead of our defined upgradedbbdata, but I didn't get that error in the backend. The app seems to be working fine for me (I tried creating an experiment and a feature flag).

Maybe I see the error due to CORS_WHITELIST env value not added in my local env file. I will check after updating that.

PR #2233, opened 2025-02-19
refetch metrics on exp upsert

danoswaltCL:
related to #2207, metrics need refetched on experiment upsert so the UI will know about the new reward metric. This will work for create, edit, and import.


danoswaltCL:
@zackcl can you review this one?

VivekFitkariwala (approved):

zackcl (approved):

Issue #2191, opened 2025-01-31
mooclet backend: send "rewards" on /log


danoswaltCL:
Dependent on [#1934](https://github.com/CarnegieLearningWeb/UpGrade/issues/1934)

When users log "reward" metrics, we will trigger a lookup for experiment and user condition enrollment, and proxy a call to post the reward to the Mooclet server.

Requirements:

If client `/log`'s a metric that matched to a `rewardMetricKey` property in MoocletExperimentRef table, this will trigger a call to send the "reward" value to the Mooclet server.

The data needed for sending the reward value to the Mooclet API is:

- `moocletId`
- `outcomeVariableName`
- `versionId`
- `value` (i.e reward, will be 1 or 0)

So I think we will need to:
- Find `MoocletExperimentRef` by `rewardMetricKey`

If found, get the versionId:
- Find user's condition in `IndividualEnrollment` using user id and experiment id.
- Find the version id by matching the condition id to the corresponding "version" in `MoocletExperimentRef.versionConditionMaps`

Determine `value` (reward)
- The `rewardMetricKey` will always either give `SUCCESS` or `FAILURE`.
- Success = 1, Failure = 0

Get `outcomeVariableName` from MoocletExperimentRef (we should add this property to the MoocletExperimentRef as a column and add it on experiment creation instead of searching in the policyParameters for it here).

POST these details to /value:

```
{
  moocletId,
  outcomeVariableName,
  versionId,
  value
}
```

PR #2234, opened 2025-02-19
add mooclet reward service and sync rewards on log

danoswaltCL:
#2191


danoswaltCL:
I will take a look at why there's a ci/cd build failure

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -767,6 +769,10 @@ export class ExperimentAssignmentService {
     logger.info({ message: `Add data log userId ${userId}`, details: jsonLog });
     const keyUniqueArray: { key: string; uniquifier: string }[] = [];
 
+    if (env.mooclets.enabled) {
+      this.moocletRewardsService.parseLogsForActiveRewardMetricKey(userDoc, jsonLog, logger);
```
VivekFitkariwala, line 773: Can we have a better name for this function, as this also sends the rewards?
`sendRewardsForActiveMetricsKey`

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,207 @@
...
+      id: uuid(),
+      metric: rewardMetricKey,
+      datatype: IMetricMetaData.CATEGORICAL,
+      allowedValues: [BinaryRewardMetricAllowedValue.SUCCESS, BinaryRewardMetricAllowedValue.FAILURE],
+    };
+
+    return this.metricService.saveAllMetrics([metric], [context], logger);
+  }
+
+  public async parseLogsForActiveRewardMetricKey(
```
VivekFitkariwala, line 71: The code is jumping from one function to another instead, which is making it difficult to follow can we have all the control in a single function, and all the modular functions should return the result after doing specific action to the control function?

```
// Get all active mooclet experiment

// Get log attributes

// Find experiment with reward metrics keys which exists in the log attributes

// Determine versions for rewards

// Send rewards
```

VivekFitkariwala (dismissed):

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,207 @@
...
+      } = log;
+
+      const attributesKeys = Object.keys(attributes);
+      for (const key of attributesKeys) {
+        this.findExperimentByRewardMetricKey(key, attributes[key], params);
+      }
+    });
+  }
+
+  private findExperimentByRewardMetricKey(key: string, value: string | number, searchParams: FindRewardParams): void {
```
bcb37, line 108: There's no guarantee that `value` here will be either 'SUCCESS' or 'FAILURE'. when, for any reason, /log is called with something else, a reward variable is created with no value. That's probably fine if we're sure those are ignored w.r.t updating the algorithm. But if we're sending data that'll just be ignored (or maybe won't), it could save traffic to just not send it in those cases (or error out).

danoswaltCL: true, we should log a warning about a bad value and not send anything.

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
danoswaltCL: yes i can do that

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,207 @@
...
+    return this.metricService.saveAllMetrics([metric], [context], logger);
+  }
+
+  public async parseLogsForActiveRewardMetricKey(
+    user: RequestedExperimentUser,
+    logs: ILogInput[],
+    logger: UpgradeLogger
+  ): Promise<void> {
+    try {
+      const moocletExperimentRefs = await this.moocletExperimentRefRepository.getRefsForActivelyEnrollingExperiments();
```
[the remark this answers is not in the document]
danoswaltCL: sure couldn't hurt, i'll try that

danoswaltCL:
@VivekFitkariwala @bcb37 comments addressed, I refactored the main function quite a bit to be easier to follow the sequence of things.

I didn't add caching the get moocletExpirimentRef call, wanted to discuss that. 
- One change I made was to switch the order, so first parse the logs for valid rewards and return early if none. so only the logs that are very likely to send a reward will grab that active mooclet ref data, which is much smaller frequency of extra db calls now.
- Caching still couldn't hurt, but invalidating the cache for this query would be kind of a pain. That being said, the TTL for our caches are a matter of seconds, so invalidation isn't terribly important, we could potentially cache this call and not worry about invalidation.

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,252 @@
...
+        logger.warn({
+          message: 'Reward metrics were logged, but no active experiments matched the rewardMetricKeys',
+          moocletExperimentRefs,
+          validSimpleMetricAttributes,
+        });
+        return;
+      }
+
+      // For each match, find the user's condition, map to version id, and create reward object
+      experimentsMatchingLoggedRewardKey.forEach(async (experimentRewardPair) => {
```
VivekFitkariwala, line 110: `forEach` will execute this code serially. Here, `map` should be used to make this code run in parallel.

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,252 @@
...
+        const { moocletExperimentRef, rewardMetricValue } = experimentRewardPair;
+        const enrollment = await this.findEnrolledCondition(user.id, moocletExperimentRef.experimentId, logger);
```
VivekFitkariwala, line 112: A db call is happening for each experiment. We can fetch all the enrollment documents in a single call.

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
```diff
@@ -0,0 +1,252 @@
...
+            value: BinaryRewardMetricValueMap[rewardMetricValue],
+            mooclet: moocletExperimentRef.moocletId,
+            version: versionId,
+          };
+          logger.info({
+            message: 'Sending reward to mooclet',
+            reward,
+            user,
+          });
+          this.moocletDataService.postNewReward(reward, logger);
```
VivekFitkariwala, line 147: `await` should be used, and if there is a way to send all the rewards in a single call, that would be better.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/MoocletRewardsService.ts
danoswaltCL: it looks like I possibly _can_ send more than one reward at a time, I dug into their API and the django framework allows this if I send append `/create_many`, aka: `/value/create_many` and when I test this, I surprisingly can send multiple values in an array.

This isn't documented by them though, so I'm not 100% sure this will work as intended, would need to test out sending multiple. For MVP, I'd like to just send them one at a time as I've tested this quite a bit, and the limited usage that is planned for experiments with Mooclets at this point is that 1 `/log` will equal 1 reward at a time, because they will be firing a log close to moment they have completed the task that requires a reward, not in batches.

danoswaltCL:
@VivekFitkariwala comments have been addressed, please take another look

VivekFitkariwala (approved):

PR #2237, opened 2025-02-21
Add CORS_WHITELIST=localhost to .env.example

### LINKED ISSUE #2236: Add CORS_WHITELIST=localhost to .env.example
The `CORS_WHITELIST=localhost` should be added to the `.env.example` file so that the users can create a proper `.env` file out of the example file.

zackcl:
Resolves #2236

I'm not sure if we should also add `CORS_WHITELIST=localhost` to the `backend/packages/Upgrade/.env.docker.local.example` file, but it seems to be working without adding it.


danoswaltCL:
Interesting that it works without it, can't hurt to have it in there as an example, makes sense to me.

Issue #2235, opened 2025-02-20
bug: segment list members can have hidden characters


danoswaltCL:
We had a production issue where users who should have gotten a flag because they were in an include list were not seeing it. Eventually I found the issue was a hidden `\r` carriage return character in some of the `groupid`, like `"groupId":"108485675hpoo\r"`.

I found 4 existing segments in prod that also have this issue with some of the members, probably however they were copy pasted into the file that got imported.

We will need a hotfix to trim out white space and hidden characters on segment creation / update.

PR #2238, opened 2025-02-21
Hotfix/trim segment ids before insert

danoswaltCL:
#2235

To recap the issue, `\r` hidden carriage returns were sneaking into some of the segment ids in the json, which renders them useless. The requirement here should be that this fix addresses sanitizing those ids to remove any whitespace or hidden carriage returns or newline characters.

This is a hotfix because several pre-existing segment lists in prod had this issue, and it may have led to wrong assignments even though it looks like the ids are perfectly fine.

This adds a utility in Segment Service to trim the userids and/or groupids before being saved. This is the first time I've really looked at this code, so if there's a better approach let me know.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/SegmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/SegmentService.ts
@@ -626,6 +626,6 @@ export class SegmentService {
   }
 
   private trimAndRemoveHiddenChars(value: string): string {
-    return value.trim().replace(/[\r\n]/g, '');
+    return value.replace(/[\r\n\t]/g, '').trim();
   }
 }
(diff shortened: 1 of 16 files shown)


zackcl (approved):

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -616,4 +624,8 @@ export class SegmentService {
       .getRepository(Segment)
       .findOne({ where: { id: segmentDoc.id }, relations: ['individualForSegment', 'groupForSegment', 'subSegments'] });
   }
+
+  private trimAndRemoveHiddenChars(value: string): string {
+    return value.trim().replace(/[\r\n]/g, '');
```
bcb37, line 629: It's probably even less likely, but it might be good to include '\t' in the trimming too.

danoswaltCL:
Pushed up an update to include `\t`, and also swap the order to trim and replace, the replace should happen before trim because otherwise a string like `"oops \r"` would trim nothing then remove `\r`, leaving `"oops "` with a space at the end. Test updated to reflect that.

bcb37 (approved):

PR #2251, opened 2025-02-27
[Closed] Prepare new Segments UI directory

### LINKED ISSUE #2239: ENV var to toggle on the Segments UI in UI
User Story
As a dev, I can toggle UI access to new Segments features in my environment so that when the lefthand Segments link is clicked, I will conditionally see the new or old UI

Acceptance Criteria
- add ENV var for `SegmentsRefreshToggle`
- when true, we see can access the new UI via lefthand Segments link
- when false, old UI still there via lefthand Segments link

### LINKED ISSUE #2240: Prepare new Segments UI directory
Like when we did Feature Flags refresh, we will need to maintain the old and the new simultaneously for a while, so we will need to keep stuff from colliding and getting messy.

One step I took for FF was to label put a `_LEGACY` suffix on all existing files, classes, and types related to Segments.

I preferred that option over creating a bunch of files called "new" or "refresh", so we can name our new files naturally and then delete all `_LEGACY` resources later.

It also forces more mindful copy-pasting when the old pieces will work perfectly fine, I want to encourage rebuilding the data-store parts and pulling in the pieces as needed.

Requirements:
- Use a `_LEGACY` suffix on existing files, classes, types to prevent collisions
- Create a new `core/segments` directory with the typical files, they can start with a bare amount of info, obvious stuff can be copied over.

Image

- Create a new `features/dashboard/segments` dir with empty `pages` and `modals` dirs to start things out, to model this like the feature-flags setup.

Image

zackcl:
Resolves #2239, resolves #2240
- Renamed the `frontend/projects/upgrade/src/app/core/segments` folder to `frontend/projects/upgrade/src/app/core/segments_LEGACY` 
- Added `_LEGACY` suffix on existing files, classes, types inside the renamed `frontend/projects/upgrade/src/app/core/segments_LEGACY` folder
- Created the new `frontend/projects/upgrade/src/app/core/segments` folder (currently empty)
- Renamed `frontend/projects/upgrade/src/app/features/dashboard/segments` folder to `frontend/projects/upgrade/src/app/features/dashboard/segments-legacy`
- Created the new `frontend/projects/upgrade/src/app/features/dashboard/segments` folder that contains empty `modals` and `pages` folders
- Added `segmentsRefreshToggle` ENV var to toggle between old and new UI via lefthand Segments link (Currently doesn't do anything)
- Reference Feature Flag PRs: #1353, #1405

Sorry for the big PR. I've manually renamed classes and types using the "Rename Symbol" in VSCode, referring to changes made to legacy feature flag code (#1353).


danoswaltCL:
hmmmmmm doing this for the UI feature components makes sense, but now I'm wondering if we should leave the stuff in the `core` folder as-is, it's not quite the same as in FF where the data was going to be majorly different. Here it will mostly stay the same, and a lot of it is "new" already because it was updated to support the feature flags list stuff, so we'll end up re-creating basically the same stuff... 

Maybe the better approach will be to selectively do this:

- Leave the core stuff as-is. IF AND WHEN we are knowingly making a change in the `core` directory that modifies existing types / actions / selectors / effects / service methods, THEN we put _LEGACY on those at that point in time. This would be for things that we want the existing name in particular, we should want the natural naming of things.

zackcl:
Closing due to the new PR #2252 which excludes the changes made to the `core` directory.

PR #2252, opened 2025-02-27
[WIP] Prepare new Segments UI directory

### LINKED ISSUE #2239: ENV var to toggle on the Segments UI in UI

### LINKED ISSUE #2240: Prepare new Segments UI directory

zackcl:
Resolves #2239, resolves #2240

- Renamed `frontend/projects/upgrade/src/app/features/dashboard/segments` folder to `frontend/projects/upgrade/src/app/features/dashboard/segments-legacy`
- Created the new `frontend/projects/upgrade/src/app/features/dashboard/segments` folder that contains empty `modals` and `pages` folders
- Added `segmentsRefreshToggle` ENV var to toggle between old and new UI via lefthand Segments link (Currently doesn't do anything)


frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -4,7 +4,7 @@ import { MatConfirmDialogComponent } from '../components/mat-confirm-dialog/mat-
 import { DeleteFeatureFlagModalComponent } from '../../features/dashboard/feature-flags/modals/delete-feature-flag-modal/delete-feature-flag-modal.component';
 import { ImportFeatureFlagModalComponent } from '../../features/dashboard/feature-flags/modals/import-feature-flag-modal/import-feature-flag-modal.component';
 import { UpsertFeatureFlagModalComponent } from '../../features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component';
-import { UpsertPrivateSegmentListModalComponent } from '../../features/dashboard/segments/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component';
+import { UpsertPrivateSegmentListModalComponent } from '../../features/dashboard/segments-legacy/modals/upsert-private-segment-list-modal/upsert-private-segment-list-modal.component';
```
danoswaltCL, line 7: Ah, can you move this component to the new `segments/modals` directory actually, and let vscode update the import paths? This was the new component developed during the FF stuff for adding / editing lists, so it belongs with the new stuff. I looked through the other components and I think this is the only one that needs to move.

danoswaltCL (approved):

PR #2253, opened 2025-03-03
Display N/A for weights on the stepper/details page for Mooclet experiments

### LINKED ISSUE #2211: mooclet frontend: mooclet experiments shouldn't show condition weights
In mooclet experiments, condition weights are controlled by "priors" initially and will change based on "rewards" that have come in as enrolled users give feedback.

That means that the "weights" shown in the design stepper and in the design tab should be hidden, they aren't going to be used or accurately represent the current condition weights.

Requirements:

If a mooclet experiment:

In stepper: 
- Disable (or hide?) the weights checkbox
- Pre-fill N/A for the weights values

In Design tab: 
- Add N/A

In Data tab?
- maybe show N/A there also

Image

zackcl:
Resolves #2211

Changes made:
- Feature: Hide Weight Equally checkbox on the stepper for Mooclet experiments
- Feature: Display `N/A` for weights on the stepper and details page for Mooclet experiments
- Enhancement: Refined styles for tables on the details page
- Enhancement: Disable factorialExperimentDesignForm in the stepper when editing a Mooclet experiment
- Bugfix: Suppress the initial `DESIGN_TYPE_CHANGED` warning message on the stepper when editing a factorial experiment
- Enhancement: Prevent weights being reset to 0 when the Weight Equally checkbox is unchecked (for consistency with factorial conditions table)
- Bugfix: Reset the assignment algorithm to `RANDOM` when the stepper is closed instead of when the design step is cleared by the updated experiment type

FYI, I didn't address the bugs I mentioned in this PR regarding the Include checkboxes in the factorial conditions table, as it wasn't as simple as I thought. I will create a separate ticket for that.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
+++ b/frontend/projects/upgrade/src/app/core/experiment-design-stepper/experiment-design-stepper.service.ts
@@ -670,6 +670,10 @@ export class ExperimentDesignStepperService {
     );
   }
 
+  clearAssignmentAlgorithm(): void {
+    this.currentAssignmentAlgorithm$.next(ASSIGNMENT_ALGORITHM.RANDOM);
+  }
+
   clearDecisionPointTableEditModeDetails(): void {
     this.store$.dispatch(experimentDesignStepperAction.actionClearDecisionPointTableEditDetails());
   }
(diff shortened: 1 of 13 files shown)


danoswaltCL:
looks good, ty

danoswaltCL (approved):

PR #2264, opened 2025-03-06
Updating Autoscaling CPU Threshold

kcalvo-cli:
Modifying the autoscaling CPU threshold to ensure resources are properly allocated during periods of increased load.


shpwe (approved):
LGTM - This would also need to make its way into the release branches (since staging uses the release branches for automatic deploys.)

PR #2272, opened 2025-03-10
✨ Global excludes segment per context

VivekFitkariwala:
Create multiple exclude segments per segment.

### DIFF
--- a/backend/packages/Upgrade/src/init/seed/globalExcludeSegment.ts
+++ b/backend/packages/Upgrade/src/init/seed/globalExcludeSegment.ts
@@ -30,13 +30,13 @@ export async function createGlobalExcludeSegment(logger: UpgradeLogger): Promise
   const segmentService = Container.get<SegmentService>(SegmentService);
 
   // Get the global exclude segment
-  const globalExcludedSegment = await segmentService.getAllGlobalExcludeSegments(logger);
+  const globalExcludeSegments = await segmentService.getAllGlobalExcludeSegments(logger);
 
   const contexts = Object.keys(env.initialization.contextMetadata);
 
   // Filter context for which global exclude segment is not present
   const contextWithNoGlobalExclude = contexts.filter(
-    (con) => !globalExcludedSegment.some(({ context }) => context === con)
+    (con) => !globalExcludeSegments.some(({ context }) => context === con)
   );
 
   // Exit the function if all context has global exclude segment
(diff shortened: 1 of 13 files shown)


backend/packages/Upgrade/src/init/seed/globalExcludeSegment.ts
```diff
@@ -3,28 +3,86 @@ import { SegmentService } from '../../../src/api/services/SegmentService';
...
-    } catch (err) {
-      const error = new Error('Error while creating Global Exclude Segment');
-      logger.error(error);
-      throw error;
-    }
+
+  // Get the global exclude segment
+  const globalExcludedSegment = await segmentService.getAllGlobalExcludeSegments(logger);
+
+  const contexts = Object.keys(env.initialization.contextMetadata);
```
zackcl, line 35: I think it would be better to rename `globalExcludedSegment` to `globalExcludeSegments` for clarity.

zackcl:
I tested what happens after deploying this branch and noticed some potential issues.

For testing, I created 5 segments and listed some segment IDs in the Global Exclude with the "ALL" context on the `dev` branch:

Screenshot 2025-03-11 at 5 55 24 PM

Screenshot 2025-03-11 at 5 55 20 PM

After switching to your branch and restarting UpGrade, I observed that only the 3 segments listed in the Global Exclude were copied over to the new Global Excludes, each with their own context:

Screenshot 2025-03-11 at 6 00 54 PM

Screenshot 2025-03-11 at 6 01 01 PM

### Issues Noticed
1. Missing Individual and Group Type IDs:
The individual and group type IDs I defined earlier are now missing. This might be fine if our deployed environments only have segments in the Global Exclude, but it's something to confirm.

2. Segments Copied to all Contexts:
All 3 segments were copied to every Global Exclude, regardless of the context. This means we'd need to manually delete segments that don't belong to the right Global Exclude.

This might be acceptable (and still better than not excluding anyone), but it introduces manual cleanup steps.

### Possible Alternative Approach
Instead of automatically copying the existing segments from the "ALL" context, we could skip this step and handle it manually after deployment. This would ensure only the necessary segments are added to each context.

VivekFitkariwala:
I have added the filtering of segments as well as groups according to the `CONTEXT_METADATA` information we have.

zackcl (approved):

zackcl:
@VivekFitkariwala Could you please resolve conflicts?

VivekFitkariwala:
@zackcl done

danoswaltCL:
@zackcl i pushed a commit to remove the unused import that was failing the test runner and was gonna merge, but i noticed it looks like there were a couple commits added after you approved, would you be able to quickly check if it still looks good and we can merge this?

zackcl:
@danoswaltCL @VivekFitkariwala Tested it again and still worked as expected.

zackcl (approved):

Issue #2255, opened 2025-03-04
Segments Paginated Endpoints


kawcl:
There should be two endpoints for getting the segments.

Get -> global segments
The global segment will be limited, so pagination is not needed. It should return all the global segment in the database.

POST -> paginated segments
This should return the list of non-global segments in a paginated way. It should be similar to how the experiment and feature flag is implemented.

We will ignore supporting sorting/searching on 'Status' with pagination for now.

- Create a db migration to add 'tags' to segments.

VivekFitkariwala:
@danoswaltCL @kawcl, can you check the story and remove the need-refinement flag?

danoswaltCL:
I edited GET to POST for paginated call.

zackcl:
@bcb37 Just a reminder - feel free to take this one. I also created a frontend-only PR (#2283) that adds a function to make requests to the `/segments/paginated` endpoint for your reference.

PR #2280, opened 2025-03-12
Segments Root Section Card Table + Data Fetch

### LINKED ISSUE #2243: Segments Root Section Card Table
The `content` slot of the Segments Root Section Card will contain a Mat-Table. There is no special "Common" component for this, but there are common utility functions that in the `common-table-helpers` service that should be reused here.

We already pretty much have the exact code we need if we reference the Feature Flag Root section card and table, the difference will just be the type of data and the columns, and the new "show global segments" toggle.

So it can all be hooked up in this story to complete the table + search widget capabilities and the global filter. 

Requirements:

- Table headers = ['Name', 'Status', 'Updated at', 'App Context', 'Tags', 'Lists']
- Should fetch segments from `segments/paginated`
- Should show load-spinner while loading in the table space.
- If no segments, should display a blank state: [Figma](https://www.figma.com/design/VF9Tck8QIOG1yLvCXUOP6W/UpGrade-UI-Latest-18?node-id=36230-22213&t=JmyixIBBAJrpioFD-0)
- With segments present, search functionality in widget should work as expected.
- The global segments toggle should filter out global segments when turned off.
- Collapse should hide the table.

zackcl:
Resolves #2243

The `SegmentRootSectionCardComponent` and `SegmentRootSectionCardTableComponent` are implemented similarly to their corresponding components in Feature Flags. They can be further updated once we add the paginated segments endpoint (#2255)


danoswaltCL:
i'd like to hold up until #2255 is ready

danoswaltCL:
@zackcl @bcb37 pushed up connecting the the paginated call with `fetchSegmentsPaginated()`, table works for paginated call on load and on scroll, old experience still works with `fetchSegmentsAll()`... I didn't name it `legacy` because while updating this I see that we will continue to want to keep the GET segments call around: when we're creating lists for FF or experiments, we'll need to fetch all available segments for the dropdown.

danoswaltCL:
ha true, ok I'll approve and merge.

danoswaltCL (approved):

PR #2282, opened 2025-03-12
Feature/frontend env gitignore

danoswaltCL:
#2281 

The existence of an `environment.ts` seems to be needed by angular.json to do its thing, so gitignore not really an option for that. so I created an `environment.local.example.ts` and will gitignore `environment.local.example.ts`.

So going forward, everyone will just need to create a new `environment.local.ts` file, and `npm start` will now use this build instead of `environment.ts`.

`environment.ts` s effectively not actually used by anything, just needs to be there in order to be referenced by imports in the code, even though it really will be replaced with an appropriate environment file.

Will follow up with documentation of this in gitbook.


frontend/angular.json
```diff
@@ -209,6 +209,46 @@
...
               ]
             },
+            "local": {
+              "fileReplacements": [
+                {
+                  "replace": "projects/upgrade/src/environments/environment.local.ts",
+                  "with": "projects/upgrade/src/environments/environment.ts"
+                }
+              ],
+              "optimization": true,
```
zackcl, line 219: Shouldn't this be reversed? Looks like "replace" is all set to `environment.ts` in other parts.

```
            "prod": {
              "fileReplacements": [
                {
                  "replace": "projects/upgrade/src/environments/environment.ts",
                  "with": "projects/upgrade/src/environments/environment.prod.ts"
                }
              ],
```

zackcl:
> The existence of an environment.ts seems to be needed by angular.json to do its thing, so gitignore not really an option for that. 

What do you mean? Can't we ask the user to create the `environment.ts` (out of `environment.example.ts`) before building the app?

danoswaltCL:
no, that's what I tried first, but `environment.ts` has a special role in the build process, it essentially is a placeholder that gets replaced at build time when you pass in a configuration.

`import { environment } from './environments/environment';` <--- this is the environment.ts file, replaced by the build configuration in angular.json. It has to exist even if it doesn't get really used itself.

If we try to only create it in local and use gitignore to avoid pushing up our local configs, that will make it disappear from the branch on remote. That's what we want for a local build config file, but we can't do it with `environment.ts` specifically because it will fail on ci/cd because it won't exist on remote anymore.

So, environment.local.ts avoids that.

zackcl (approved):

zackcl (changes requested):
I approved it by mistake. Please resolve the comment I mentioned.

zackcl (approved):

PR #2285, opened 2025-03-13
✨ Adding validations

VivekFitkariwala:
Adding validation of policy parameters

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
@@ -570,7 +570,7 @@ interface ExperimentPaginationInfo extends PaginationResponse {
  *   - name: Experiments
  *     description: CRUD operations related to experiments
  */
-// @Authorized()
+@Authorized()
 @JsonController('/experiments')
 export class ExperimentController {
   constructor(
--- a/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
+++ b/backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
@@ -570,7 +570,7 @@ interface ExperimentPaginationInfo extends PaginationResponse {
  *   - name: Experiments
  *     description: CRUD operations related to experiments
  */
-@Authorized()
+// @Authorized()
 @JsonController('/experiments')
 export class ExperimentController {
   constructor(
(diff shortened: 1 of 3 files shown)


VivekFitkariwala:
@bcb37, validations are working. Can you check it once?

backend/packages/Upgrade/src/api/controllers/ExperimentController.ts
```diff
@@ -570,7 +570,7 @@ interface ExperimentPaginationInfo extends PaginationResponse {
  *   - name: Experiments
  *     description: CRUD operations related to experiments
  */
-@Authorized()
+// @Authorized()
```
ppratikcr7, line 573: @VivekFitkariwala This should be uncommented, right.

types/src/Mooclet/MoocletTSConfigurablePolicyParametersDTO.ts
```diff
@@ -1,39 +1,78 @@
...
+    });
+  };
+};
+
 export class MoocletTSConfigurablePolicyParametersDTO extends MoocletPolicyParametersDTO {
   @IsDefined()
   @ValidateNested()
   @Type(() => Prior)
-  prior: Prior = new Prior();
+  prior: Prior;
```
danoswaltCL, line 72: Without the default values for prior and it's properties, the frontend can't create a default object, the priors as `"prior": { "success": 1, "failure": 1 }` should show up here in the editor.

image

danoswaltCL: The validation isn't quite working for me on the frontend when I fill in prior manually either, it allows me to hit "next" with an object like below. It does catch it on submit and give correct message, however the other validations will raise an error immediately as the user types.

{
  "prior": {
    "success": "d",
    "failure": 2
  },

bcb37:
> @bcb37, validations are working. Can you check it once?

In addition to Dan's comment about the default values (which works if we're defining the types within the root),  we also need to disallow extraneous attributes. That works if you add `{ whitelist: true, forbidNonWhitelisted: true, forbidUnknownValues: true }` as the value for 'validate' in the controller.
But we do want to be able to use those default values.

types/src/Mooclet/MoocletTSConfigurablePolicyParametersDTO.ts
VivekFitkariwala: We should not add the default values in the class constructor because class transform uses this class constructor to create the definition and assign the object (to be tested) to the constructed object. If we are adding the default values, the constructor definition also contains the value which we don't want.
For testing, add the default value for `Prior` in the types
Remove the `"prior": { "success": 1, "failure": 1}`. 
It will not throw an error.

I can add the appropriate config for the frontend side for validation to work correctly.

VivekFitkariwala:
@bcb37 I don't think that we have two different instances of `reflect-metadata` because I am able to re-create the issue even after pasting the code in the backend. It seems like it is related to the discriminator.

danoswaltCL:
To be continued, validation still need attention but closing as we need the default behavior.

PR #2291, opened 2025-03-13
bring back experimentId null logic for now

danoswaltCL:
https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1741886587538479?thread_ts=1741880529.130829&cid=C055CGCRP6Z

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -196,18 +196,24 @@ export class ExperimentAssignmentService {
     });
 
     if (experiments.length) {
-      const dpExpExists = experiments.filter((exp) => exp.id === experimentId);
+      if (experimentId) {
+        const dpExpExists = experiments.filter((exp) => exp.id === experimentId);
 
-      if (!dpExpExists.length) {
-        const error = new Error(
-          `Experiment ID not provided for shared Decision Point in markExperimentPoint: ${userId}`
-        );
-        (error as any).type = SERVER_ERROR.INVALID_EXPERIMENT_ID_FOR_SHARED_DECISIONPOINT;
-        (error as any).httpCode = 404;
-        logger.error(error);
-        throw error;
+        if (!dpExpExists.length) {
+          const error = new Error(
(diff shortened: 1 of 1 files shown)


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -196,18 +196,24 @@ export class ExperimentAssignmentService {
     });
 
     if (experiments.length) {
-      const dpExpExists = experiments.filter((exp) => exp.id === experimentId);
+      if (experimentId) {
+        const dpExpExists = experiments.filter((exp) => exp.id === experimentId);
```
bcb37, line 200: `dpExpExists` seems like that name of a boolean, rather than of an array that will always have either 0 or 1 elements. (Is there a reason `find()` wasn't used here; I think it would be more efficient)

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -196,18 +196,24 @@ export class ExperimentAssignmentService {
...
+        if (!dpExpExists.length) {
+          const error = new Error(
+            `Experiment ID not provided for shared Decision Point in markExperimentPoint: ${userId}`
+          );
+          (error as any).type = SERVER_ERROR.INVALID_EXPERIMENT_ID_FOR_SHARED_DECISIONPOINT;
+          (error as any).httpCode = 404;
+          logger.error(error);
+          throw error;
+        }
+        experiments = dpExpExists;
```
bcb37, line 211: After this line (or line 214), `experiments` will also always be an array with either 0 or 1 element - even though later on `map()` is used, etc.

backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -196,18 +196,24 @@ export class ExperimentAssignmentService {
...
+      } else {
```
bcb37, line 212: Maybe a comment here to say that we're running this function to guess which experiment the user is being excluded from - is it guaranteed to produce the same result as the pooling logic?

bcb37 (approved):
approving for a quick temporary fix

PR #2297, opened 2025-03-18
Implement SegmentListsSectionCardComponent

### LINKED ISSUE #2288: Implement SegmentListsSectionCardComponent
### Requirements 
- `SegmentListsSectionCardComponent` should use `CommonSectionCardComponent`, `CommonSectionCardTitleHeaderComponent` and `CommonSectionCardActionButtonsComponent`
- `SegmentListsTableComponent` should use `CommonDetailsParticipantListTableComponent` (This could be created as a separate story)

### Design
Image

Image

### LINKED ISSUE #2312: Implement SegmentListsTableComponent
### Requirements 
- `SegmentListsTableComponent` should use `CommonDetailsParticipantListTableComponent`

### Design
Image

Image

zackcl:
Resolves #2288, resolves #2312

Still work in progress. 

Reference from @bcb37 :
>  I think I remember we wanted to keep the old component around for displaying old-style segments. Old-style segments will always have at least one non-private member. I think that was how we were going to make the distinction, So, create will always create an empty segment, which we can then add lists to. But when you go to the details page for a segment, it will be the old-style page for any public segment that has one or more of: individuals, groups, non-private subsegments.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page-content/segment-lists-section-card/segment-lists-table/segment-lists-table.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page-content/segment-lists-section-card/segment-lists-table/segment-lists-table.component.html
@@ -1,1 +1,10 @@
-<p>segment-lists-table works!</p>
+<div class="table-container">
+  <app-common-details-participant-list-table
+    [tableType]="tableType"
+    [dataSource]="dataSource$ | async"
+    [actionsDisabled]="actionsDisabled"
+    [isLoading]="isLoading$ | async"
+    noDataRowText="segments.details.lists.card.no-data-row.text"
+    (rowAction)="onRowAction($event)"
+  ></app-common-details-participant-list-table>
+</div>
--- a/frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page-content/segment-lists-section-card/segment-lists-table/segment-lists-table.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page-content/segment-lists-section-card/segment-lists-table/segment-lists-table.component.ts
@@ -1,14 +1,28 @@
 import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output } from '@angular/core';
 import { ParticipantListRowActionEvent } from '../../../../../../../../core/feature-flags/store/feature-flags.model';
+import { CommonDetailsParticipantListTableComponent } from '../../../../../../../../shared-standalone-component-lib/components/common-details-participant-list-table/common-details-participant-list-table.component';
+import { CommonModule } from '@angular/common';
+import { TranslateModule } from '@ngx-translate/core';
+import { SegmentsService } from '../../../../../../../../core/segments/segments.service';
+import { FEATURE_FLAG_LIST_FILTER_MODE } from 'upgrade_types';
 
 @Component({
   selector: 'app-segment-lists-table',
-  imports: [],
   templateUrl: './segment-lists-table.component.html',
   styleUrl: './segment-lists-table.component.scss',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [CommonDetailsParticipantListTableComponent, CommonModule, TranslateModule],
 })
 export class SegmentListsTableComponent {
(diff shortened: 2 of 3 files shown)


frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page.component.ts
```diff
@@ -1,12 +1,71 @@
...
+
+  ngOnInit() {
+    this.segmentIdSub = this.route.paramMap.subscribe((params) => {
+      const segmentId = params.get('segmentId');
+      if (segmentId) {
+        this.segmentsService.fetchSegmentById(segmentId);
+      }
+    });
+
+    this.segment$ = this.segmentsService.selectedSegment$.pipe(filter((segment) => !!segment));
```
danoswaltCL, line 43: how come we have to do this filter? can segments come as undefined or null?

frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page.component.ts
```diff
@@ -1,12 +1,71 @@
...
+
+    // Determine if we should use the legacy view based on segment structure
+    this.shouldUseLegacyView$ = this.segment$.pipe(map((segment) => this.shouldUseLegacyUI(segment)));
+  }
+
+  // Logic to determine if we should use the legacy UI
+  private shouldUseLegacyUI(segment: Segment): boolean {
```
danoswaltCL, line 50: nice, i'd suggest making this a selector, or at least put into service. It would work essentially the same, but ngrx selectors are built for this exact kind of purpose, it's a data-layer job not a component job.

zackcl:
@danoswaltCL Thanks for the early reviews. I had to stop in the middle because I need to work on the mooclet reward stuff. Feel free to take this one if you’d like.

frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page.component.html
```diff
@@ -1,4 +1,14 @@
-<app-common-page>
-  <app-segment-details-page-header header></app-segment-details-page-header>
-  <app-segment-details-page-content content></app-segment-details-page-content>
-</app-common-page>
+<ng-container *ngIf="segment$ | async as segment">
+  <ng-container *ngIf="shouldUseLegacyView$ | async; else newView">
+    <!-- Legacy View -->
+    <view-segment></view-segment>
```
danoswaltCL, line 4: i think this'll work for viewing, but what about the actions like adding / importing / editing segments that can be launched from the legacy view though? maybe we can re-wire those to the newer modals where possible when the toggle is on so we don't have to keep supporting those components?

zackcl: Only the details page will be displayed in the legacy view where you can't add or import a segment. I think we should use the legacy modal for editing legacy segments since they are very different (e.g., the legacy edit modal covers both overview details and segment members unlike the new design). 

I think there could be a possibility of displaying everything in the new segment details page though (so that we don't need to use the legacy view). Maybe we can discuss this in the meeting.

zackcl:
@danoswaltCL I’ve resolved your comments. I think I’m done with this PR unless I missed something. I wasn’t able to check the redesigned Segment Details page yet because it’s still redirecting to the old one, and I currently can’t create a new segment due to the missing "Add Segment" modal. I can continue working on this later if needed.

zackcl:
OK, I made a POST request to `http://localhost:3030/api/segments` with the following body:
```
{
    "name": "Hola2",
    "description": "",
    "context": "mathstream",
    "userIds": [],
    "groups": [],
    "subSegmentIds": [],
    "type": "public",
    "status": "Unused"
}
```
And I was able to see the new details page instead of the old one.
Screenshot 2025-03-26 at 3 37 15 AM


I would still need to fix the text in the UI. I will fix that soon.

zackcl:
I've fixed the labels in the Lists section card, and made the empty table component visible.

Screenshot 2025-03-26 at 7 32 49 PM

zackcl:
Feel free to review and merge this PR.

zackcl:
I went ahead and implemented the `SegmentListsTableComponent` as well (#2312).

Screenshot 2025-03-26 at 9 31 23 PM

danoswaltCL (approved):
Looks great!

PR #2299, opened 2025-03-19
Bugfix: Created and Updated dates for fresh and imported experiments-ff-segments

### LINKED ISSUE #2213: Update "Created on" date for imported experiments to reflect import date and show blank "Updated On" date for experiments/feature-flag/segments
Currently the "created on" date of imported experiments reflect the original date of creation. Imported experiments' create date should instead reflect the date of import. This makes the create date confusing when and if an older experiment is used as a template. 

Image

ppratikcr7:
This PR resolves #2213 

It will resolve the below things:

1. The created date for the imported experiment/feature-flag/segment will have the current created date instead of old date as per the design JSON.
2. The updated date for the imported and freshly created experiment/feature-flag/segment will have a blank (-) date as they are freshly created and not updated yet. Below are attached screenshot for the same.
3. Once we update the experiment/feature-flag/segment, we will store the updated date and update the `versionNumber` for these tables and based on the `versionNumber`, we will show the updated on date.
4. Added date-time for experiments and segments details page as in feature-flags pages' updated-date field for uniformity and it helps differentiate the created and updated times, else we just show dates which can be same for the same date. (Let me know in case this is not intended)
5. Replaced the `async` import from frontend tests as its deprecated from `@angular/core/testing` after angular update. We need to use `waitForAsync` instead. Please find attached screenshot.
6. Updated a spelling error for the `isMoocletAssignmentAlgorithm` in ExperimentDTO.

Screenshot 2025-03-19 at 2 47 20 PM

Blank updated dates for freshly created or imported segments:

Screenshot 2025-03-19 at 5 24 50 PM

Complete date-time for created at as well, earlier we had date-time for only updated-at in Feature-Flag:

Screenshot 2025-03-18 at 7 10 33 PM

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentService.ts
@@ -1623,7 +1623,7 @@ export class ExperimentService {
     experiment.id = uuid();
     // delete createdAt date to let typeorm handle it and initialize versionNumber as fresh experiment version to detect updates
     delete experiment.createdAt;
-    experiment.versionNumber = 1;
+    delete experiment.versionNumber;
     this.deduceFactors(experiment);
     this.deduceConditions(experiment);
     this.deducePartition(experiment);
--- a/backend/packages/Upgrade/src/api/services/ExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentService.ts
@@ -1619,8 +1619,11 @@ export class ExperimentService {
     return errorString;
   }
 
-  private deduceExperimentDetails(experiment: ExperimentDTO): ExperimentDTO {
+  private deduceExperimentDetails(experiment: Experiment): Experiment {
     experiment.id = uuid();
+    // set current createdAt date and initialize versionNumber as fresh experiment version to detect updates
+    experiment.createdAt = new Date();
+    experiment.versionNumber = 1;
     this.deduceFactors(experiment);
     this.deduceConditions(experiment);
     this.deducePartition(experiment);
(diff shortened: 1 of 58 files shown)


ppratikcr7:
**Note**: Here as observed we dot update the `feature-flag` and `segment` tables on updating the inclusion/exclusion lists and segment-members. Also, toggle update for feature-flag wont update the `feature-flag` table, so the `versionNumber` wont be incremented. Should we update these tables on the above mentioned scenarios, though it might seem an over-kill for the ticket created.

danoswaltCL (dismissed):
makes sense to me

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1619,8 +1619,11 @@ export class ExperimentService {
     return errorString;
   }
 
-  private deduceExperimentDetails(experiment: ExperimentDTO): ExperimentDTO {
+  private deduceExperimentDetails(experiment: Experiment): Experiment {
     experiment.id = uuid();
+    // set current createdAt date and initialize versionNumber as fresh experiment version to detect updates
+    experiment.createdAt = new Date();
```
VivekFitkariwala, line 1625: Instead of setting the value, we should delete the value and let TypeORM set the value.

VivekFitkariwala (changes requested):

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1619,8 +1619,11 @@ export class ExperimentService {
     return errorString;
   }
 
-  private deduceExperimentDetails(experiment: ExperimentDTO): ExperimentDTO {
+  private deduceExperimentDetails(experiment: Experiment): Experiment {
     experiment.id = uuid();
+    // delete createdAt date to let typeorm handle it and initialize versionNumber as fresh experiment version to detect updates
+    delete experiment.createdAt;
+    experiment.versionNumber = 1;
```
VivekFitkariwala, line 1626: Why not delete the `versionNumber`?

VivekFitkariwala (changes requested):

VivekFitkariwala (approved):

PR #2300, opened 2025-03-19
Feature/new import segments response endpoint

danoswaltCL:
This adds a separate validation endpoint to support the common import modal style.

It simply returns a different response format, I'm guessing at what the 'warning' type should be, basing this on what FF is doing. In the interest of keeping this PR focused I didn't try to do any additional refactoring.

It also consolidates some of the types that were created as feature-flag-specific, but really are generic types that will work for other imports, and puts them into `types` so the frontend can use them as well.

I would like to add some tests, but I didn't see any current validation endpoint tests? so I think that's something I can add in a follow up ticket that more comprehensively adds tests for all of them, as imports is a common place we get bugs. (frontend PR will follow up that uses these, I've tested and they work, but I've still got some more work on that, this is ready to go).


backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -365,6 +367,33 @@ export class SegmentService {
...
+    logger.info({ message: `Validating segments` });
+    const validatedSegments = await this.validateSegmentsData(segments.map((seg) => ({ filename: seg.fileName, segment: null })));
+    const validatedImportResponses = validatedSegments.importErrors.map((seg) => {
+      if (!seg.error) {
+        return {
+          fileName: seg.fileName,
+          error: null,
+          compatibilityType: IMPORT_COMPATIBILITY_TYPE.COMPATIBLE,
+        };
+      } else if (seg.error.includes('Please import subSegment')) {
```
bcb37, line 409: If we're relying on this hard-coded string to make this determination, maybe we should create a constant so that it's value is defined in one place.

bcb37 (commented):
This looks good, just a couple questions:

Do we still need the old /validate endpoint (and its service method `validateSegments()`).

backend/packages/Upgrade/src/api/services/SegmentService.ts
danoswaltCL: yeah i know it's gross. Really the `compatibilityType` should be applied within the `checkSegmentsValidity` function when that error message itself is applied. Looking at it now I think I don't like this really either.

my hope was to do a more cosmetic change to just send a modified response without deeper refactoring into the `checkSegmentsValidity` function, because we need to keep the 'old' way around for the old modal and the actual logic doesn't need changed, it's just slightly different enough of a response that modifying it isn't so easy to shoehorn in something that makes both modals happy.

Let me go through this and make all the error strings constants, that's not so bad and good practice anyway, I won't be able to sleep at night if it's just one of them changed. I'll update this.

danoswaltCL:
yeah we still need to keep it around, the old modal will need the old response type while it is still kicking. If it were just adding the "compatibilityType" to the response I'd be able to just reuse the endpoint by adding that property and let the old modal ignore it, but the old modal wants an empty array to indicate success (it only returns errors), while the new modal wants an object for each file and its compatibility type, so it's not quite the same.

backend/packages/Upgrade/src/api/controllers/SegmentController.ts
```diff
@@ -497,6 +498,43 @@ export class SegmentController {
...
+   *       description: Validating Segments with response for Common Import Modal
+   *       consumes:
+   *         - application/json
+   *       parameters:
+   *         - in: body
+   *           name: params
+   *           description: Segment file
+   *           required: true
+   *           schema:
+   *            type: object
```
VivekFitkariwala, line 514: The schema should be updated as it is updated to the array. There are other places where this update needs to be done, which should have been done as part of earlier PRs.
```
type: array
items:
  type: object
  properties:
    fileName:
      type: string
    fileContent:
      type: string
```

VivekFitkariwala (changes requested):
The PR looks good.
Few problems to be fixed.

VivekFitkariwala (approved):

PR #2302, opened 2025-03-20
Feature/java client apache connector

danoswaltCL:
Smoke test of this worked for me testing with the `QuickTest.java` file (formerly Main.java). I installed JDK 17 on my local, and then when I target 17 here in properties for the jersey client instead of 8 I was able to replicate the PATCH error. After installing the updated dependency for apache connector as it was in the PR from years back, the error went away.

```
<properties>
		<jersey.version>2.30.1</jersey.version>
		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
		<maven.compiler.source>17</maven.compiler.source>
		<maven.compiler.target>17</maven.compiler.target>
	</properties>
```

Let me know what else we should consider to test, and if there are other deps we should go ahead and update alongside of this let's go ahead.

### DIFF
--- a/clientlibs/java/pom.xml
+++ b/clientlibs/java/pom.xml
@@ -9,7 +9,7 @@
 		at the same time that happen to rev to the same new version will be caught 
 		by a merge conflict. -->
 
-	<version>6.1.0</version>
+	<version>6.1.1</version>
 	<build>
 		<plugins>
 			<plugin>
@@ -20,10 +20,10 @@
 		</plugins>
 	</build>
 	<properties>
-		<jersey.version>2.16</jersey.version>
+		<jersey.version>2.30.1</jersey.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<maven.compiler.source>1.8</maven.compiler.source>
-		<maven.compiler.target>1.8</maven.compiler.target>
+		<maven.compiler.source>17</maven.compiler.source>
+		<maven.compiler.target>17</maven.compiler.target>
 	</properties>
 	<dependencies>
 		<!-- https://mvnrepository.com/artifact/org.testng/testng -->
--- a/clientlibs/java/pom.xml
+++ b/clientlibs/java/pom.xml
@@ -53,6 +53,11 @@
 			<artifactId>jersey-hk2</artifactId>
 			<version>2.30.1</version>
 		</dependency>
+		<dependency>
+            <groupId>org.glassfish.jersey.connectors</groupId>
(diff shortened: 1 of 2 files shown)


clientlibs/java/pom.xml
```diff
@@ -20,10 +20,10 @@
 		</plugins>
 	</build>
 	<properties>
-		<jersey.version>2.16</jersey.version>
+		<jersey.version>2.30.1</jersey.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<maven.compiler.source>1.8</maven.compiler.source>
-		<maven.compiler.target>1.8</maven.compiler.target>
+		<maven.compiler.source>17</maven.compiler.source>
+		<maven.compiler.target>17</maven.compiler.target>
```
jerith666, line 26: They unified `source` and `target` into a new `release` parameter:

```suggestion
		<maven.compiler.release>17</maven.compiler.release>
```

clientlibs/java/pom.xml
```diff
@@ -9,7 +9,7 @@
 		at the same time that happen to rev to the same new version will be caught 
 		by a merge conflict. -->
 
-	<version>6.1.0</version>
+	<version>6.1.1</version>
```
jerith666, line 12: Even though this library has seen very few updates recently, I still think the comment just above is reasonable.  So:

```suggestion
	<version>6.1.1</version> <!-- 6.1.1: switch to jersey-apache-connector -->
```

jerith666 (commented):
I dug around some more and can't find any evidence of why we didn't go with this approach back in 2023.  I guess we'll just have to rediscover it if there was some real problem.  :-/

Anyway, the substance looks good here, just a couple of minor tweaks.

jerith666:
Also there seem to be several failing NPM tests, probably we'll need another PR to land to clean those up before this one lands?

danoswaltCL:
That build failure is to be ignored, I don't know why it can't be suppressed/ignored on dev, I've asked. But long story short is that task fails on dev because it is only for beanstalk builds, we have no dev beanstalk env anymore, and we haven't felt like fixing errors for a pointless task, but it lingers around because it still needs to exist for staging and prod. At the time it seemed like the switchover to ECS was going to happen more imminently, but it hasn't, so we've just sort of ignored it.

the equivalent build task for ECS (upgrade-service) passes.

I am going to assume this was some dangling task that Jenn had been working on when she left and got lost in the void.

jerith666 (approved):
Okay, makes sense; the ECS switchover is actually happening now, so hopefully we can in fact get rid of that check soon.

PR #2303, opened 2025-03-20
Add paginated and global endpoints for segments

bcb37:
- Adds POST /paginated endpoint to segments with a request body analogous to the other paginated endpoints
- Adds GET /global segments endpoint to get all global segments (for all contexts)
- Adds 'tags' to the segments model, creates db migration

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
+++ b/backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
@@ -55,10 +55,10 @@ export class SegmentPaginatedParamsValidator {
   @IsOptional()
   @ValidateNested()
   @Type(() => ISegmentSearchParamsValidator)
-  public searchParams: ISegmentSearchParamsValidator;
+  public searchParams?: ISegmentSearchParamsValidator;
 
   @IsOptional()
   @ValidateNested()
   @Type(() => ISegmentSortParamsValidator)
-  public sortParams: ISegmentSortParamsValidator;
+  public sortParams?: ISegmentSortParamsValidator;
 }
--- a/backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
+++ b/backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
@@ -1,0 +1,64 @@
+import { Type } from 'class-transformer';
+import { IsNotEmpty, IsNumber, IsOptional, ValidateNested, IsEnum, IsString } from 'class-validator';
+import { SORT_AS_DIRECTION } from 'upgrade_types';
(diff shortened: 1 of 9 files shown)


backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
```diff
@@ -0,0 +1,64 @@
+import { Type } from 'class-transformer';
+import { IsNotEmpty, IsNumber, IsOptional, ValidateNested, IsEnum, IsString } from 'class-validator';
+import { SORT_AS_DIRECTION } from 'upgrade_types';
+
+// TODO: Move to upgrade types
```
danoswaltCL, line 5: does this TODO need resolved?

danoswaltCL:
This looks good to me based on scanning the other implementations of this type of thing, seems like what we want, a GET `/segments/global` fetch that can happen on app load probably, and the paginated call for the new table. But @VivekFitkariwala should take a look.

backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
bcb37: Probably eventually. I copied that over from the feature flag file; moving those interfaces broke stuff, but it's something that both the frontend and the backend use.

backend/packages/Upgrade/src/api/controllers/validators/SegmentPaginatedParamsValidator.ts
```diff
@@ -0,0 +1,64 @@
...
+export class SegmentPaginatedParamsValidator {
+  @IsNotEmpty()
+  @IsNumber()
+  public skip: number;
+
+  @IsNotEmpty()
+  @IsNumber()
+  public take: number;
+
+  @IsOptional()
```
VivekFitkariwala, line 55: Typescript should also reflect optional parameters.
` public searchParams?: ISegmentSearchParamsValidator;`
` public sortParams?: ISegmentSortParamsValidator;`

backend/packages/Upgrade/src/api/models/Segment.ts
```diff
@@ -37,6 +37,9 @@ export class Segment extends BaseModel {
   })
   public type: SEGMENT_TYPE;
 
+  @Column('text', { array: true, nullable: true })
+  public tags: string[];
```
VivekFitkariwala, line 41: The typescript should be fixed here
`public tags?: string[]`

There are other parameters where this problem already exists. We should fix this in a separate PR, I guess.

VivekFitkariwala (changes requested):
Minor fixes. The PR looks good to me.

VivekFitkariwala (approved):

PR #2311, opened 2025-03-25
import segments modal / common import modal

danoswaltCL:
-This repurposes the feature-flag import modal into a common import modal is data-type agnostic about the files it is validating, displaying, and importing.

It uses a dependency injection token passed in through the config data that is passed in from common-dialog service to pass in a "common-import-type-adapter" class that implements a `ImportServiceAdapter` interface, which really just is a wrapper around the methods needed to validate, import, etc.

That will allow us to import any data type with this reusable modal as long as it adheres to that interface. 

- Implements this for segments, and refactors the usage for feature flag, include list, and exclude list import modals.
- Fixes issues expand toggle (now it will open individual error details instead of expanding all)
- Fixes issues with import button enabled when it shouldn't be.
- Minor refactor to prefer observable patterns over promises
- This also includes a minor refactor for menu-buttons in the `common-section-card-action-buttons`... `name` was serving a double purpose as a label and an action identifier, and the way it was implemented was kinda different for every spot I found it (sometimes a translation key, sometimes an enum, sometimes just a string that was duplicated all over the place).

So I cleaned it up by separating it into a `label` for standard text key, and an `action` enum.

```
export interface IMenuButtonItem {
  name: string;
  disabled: boolean;
}
```

```
export interface IMenuButtonItem {
  action: string;
  label: string; // translation key
  disabled: boolean;
}
```


bcb37 (commented):
I looked this over a bunch and tested it. Looks ok to me, fwiw.

zackcl:
@danoswaltCL I found the "Import Feature Flag" modal design from the old Figma project file: https://www.figma.com/design/GaKw3nDscTnMEOzIf4CAa7/UpGrade-UI-Latest-14-(Final)?node-id=8283-9405&t=Yu13lzUFdS34DkqM-0

It appears that we want to allow importing compatible/warning files even if there are incompatible files in the modal.

Image

Image

As I tested the app, I found that the "Import" button is disabled if any incompatible file exists.

Image

Could you please update this so that the button is enabled if one or more files are compatible? Sorry about the not being clear about this in the story.

zackcl:
When I try to import a Feature Flag JSON file in the "Import Segment" modal, I'm getting the "Cannot read properties of undefined (reading 'toLowerCase')" error.

Screenshot 2025-04-01 at 2 15 04 AM

danoswaltCL:
@zackcl ok will look into these, ty for testing it out, will get the syntax error thing fixed.

For the behavior of warning / compatible, what should happen when I click import? Is the expected behavior of the modal that it should not send those "incompatible" files in the actual submit call (that would be how I'd assume), I'm not sure that would be clear to the user what "import" will do in a mixed compatibility list?

What if a "warning" files actually throws an error anyway? For example, it shows "warning" for feature flags when they have sub-segments that need to be imported, but the import actually fails, for instance. There's some pre-existing behavior that's not entirely consistent that may require some attention outside of this ticket.

zackcl:
> Is the expected behavior of the modal that it should not send those "incompatible" files in the actual submit call
Yes, we should ignore the incompatible files.

> I'm not sure that would be clear to the user what "import" will do in a mixed compatibility list?
That’s why we show the message "*Incompatible files will not be imported" below the table. Feel free to update the message if you think it could be clearer.

> What if a "warning" files actually throws an error anyway?
The idea is that "warning" files should still be importable without any issues (i.e., they contain all required properties), but they may include outdated fields or be missing some optional information. That’s why we note in the modal that these files should be reviewed post-import.

danoswaltCL:
couldn't replicate the banana.json FF issue, it loads up okay for me when validated?

image

danoswaltCL:
@zackcl fixed up comments, plus a couple other small things I noticed, but couldn't replicate the syntax error. please take another look. (The "*Incompatible files will not be imported" message had not previously been a part of the component, i've added that now.)

One other thing that I was wondering though... the error messages in the Figma design... those were only meant as placeholder for the real messages right? The way this was wired up for Feature Flags it hard-codes a specific warning message and a specific incompatible message, I think maybe some confusion on the requirements, but we should really be showing the actual message from the backend right? We should have that error message available in the /validation responses, just a matter of using that instead (and making sure it doesn't look crazy)

zackcl:
@danoswaltCL My bad. I realized I didn’t add `validateSegmentsImport` to the `environment.local.ts` file, and that caused the weird issue. I wish we could produce a clearer error when any of these environment variables is undefined, since it’s a pretty common issue if someone pulls the latest changes but forgets to update their environment file.

zackcl:
> One other thing that I was wondering though... the error messages in the Figma design... those were only meant as placeholder for the real messages right?

I personally think such generic placeholder messages in the modal is sufficient for non-developer users unless we expect them to actually debug and fix the file. More detailed error messages could be logged in the backend, though, for our debugging purposes.

zackcl (approved):

PR #2320, opened 2025-03-29
add-edit-duplicate segment modal functionality

### LINKED ISSUE #2245: Add Segment Modal
The 'Add Segment' modal is essentially the same as the 'Add Feature Flag' modal, except it does not include the 'Key' field.

**Requirements:**
- Name, Description, App Context, Tags fields should appear on the modal
- Creating a new segment should send a POST request to the `api/segments` endpoint

**Screenshot:**

Image

ppratikcr7:
This PR resolves #2245

I have added all the necessary code for add, edit and duplicate segment modal. Currently this PR needs the segmenr table list component for a listener to close the modal after add/edit/duplicate segment. Also, tags and listType are not getting stored in the backend as of now, that also needs to be handled in a separate PR. We will need to test this PR once all these missing parts are in the dev branch. Rest all seems to work fine as expected.

EDIT DAN:
1. This has separate actions and effects from the 'legacy' segments logic so it can be changed later if we want to update the endpoint (like to add tags), but is currently able to work fine using the old endpoint in basically the same way.
2. This adds Duplicate segment ability, but I left it disabled because I'm not sure it's what we want, I wanted to talk about that to figure out what the expected behavior of that is supposed to be.


zackcl:
I'm able to create two segments with the same name under the same app context. I think we should not allow this.

Screenshot 2025-04-02 at 10 09 00 PM

zackcl:
I don't see the "Duplicate Segment" option from the menu currently.

Screenshot 2025-04-02 at 10 12 52 PM

Here's the design in Figma for reference:

Duplicate Segment

danoswaltCL:
> I'm able to create two segments with the same name under the same app context. I think we should not allow this.
> 
> Screenshot 2025-04-02 at 10 09 00 PM

Ok, I see the old modal disallows this to prevent the issue, I can do that but the backend should really not allow it to happen either.

danoswaltCL:
@zackcl I added in a backend change to do a duplicate name check on submit and error if yes. Also shows error in the form that makes the form invalid until changed. The error will persist and keep getting checked while the name or context changes, removing the error if changed, putting it back if the name and context again match.

I put this in WIP still, I haven't added in the duplicate modal changes requested.

zackcl:
@danoswaltCL Shouldn't we disable the "App Context" dropdown on the"Duplicate Segment" modal? I thought that was our decision.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-segment-modal/upsert-segment-modal.component.ts
```diff
@@ -268,6 +270,25 @@ export class UpsertSegmentModalComponent {
     this.segmentService.modifySegment(segmentRequest);
   }
 
+  createDuplicateRequest({ name, description, tags }: SegmentFormData, sourceSegment: Segment): void {
+    const userIds = sourceSegment.individualForSegment.map((user) => user.userId);
+    const groups = sourceSegment.groupForSegment;
+    const subSegmentIds = sourceSegment.subSegments.map((subSegment) => subSegment.id);
```
bcb37, line 276: We need the actual segments here, not just the IDs. We should be re-creating the private segments for each duplicate, not sharing them. This also means that we'll probably need a different endpoint for this. It could be very similar.

bcb37 (commented):
A couple things about duplicating.

danoswaltCL:
> @danoswaltCL Shouldn't we disable the "App Context" dropdown on the"Duplicate Segment" modal? I thought that was our decision. Currently, I'm able to change the "App Context" on your branch.

oops, yes

danoswaltCL:
@zackcl @bcb37 updated but https://github.com/CarnegieLearningWeb/UpGrade/pull/2349 should go first, then I can use the changes there to update this to close the modal without using interceptor.

backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -377,7 +378,9 @@ export class SegmentService {
     return queryBuilder;
   }
 
-  public upsertSegment(segment: SegmentInputValidator, logger: UpgradeLogger): Promise<Segment> {
+  public async upsertSegment(segment: SegmentInputValidator, logger: UpgradeLogger): Promise<Segment> {
+    await this.checkIsDuplicateSegmentName(segment.name, segment.context, logger);
```
bcb37, line 382: This method is also used when editing - so now editing always fails unless you're changing the name.

bcb37 (changes requested):
I found one more thing.

Also, when you edit a new-style segment, it deletes all the lists.

frontend/projects/upgrade/src/app/features/dashboard/segments/modals/upsert-segment-modal/upsert-segment-modal.component.ts
```diff
@@ -1,10 +1,221 @@
...
+      appContext = sourceSegment.context;
+    }
+    const segmentRequest: UpdateSegmentRequest = {
+      id,
+      name,
+      description,
+      context: appContext,
+      userIds: [],
+      groups: [],
+      subSegmentIds: [],
```
bcb37, line 254: We need subSegmentIds from the segment here (same as with duplicate)

danoswaltCL:
@bcb37 updated, ready for re-re-review

bcb37:
> @bcb37 updated, ready for re-re-review

This works now. Duplicating still adds the same lists, rather than cloning them - but I just put up a pr (https://github.com/CarnegieLearningWeb/UpGrade/pull/2350) to fix that.

bcb37 (approved):

PR #2321, opened 2025-03-31
♻️ Refactor preview user

VivekFitkariwala:
Adding cache for getting Preview user by Id and removing creation of exclusion doc during the state change to Preview State

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -182,7 +182,7 @@ export class ExperimentAssignmentService {
     }
 
     const userId = userDoc.id;
-    const previewUser: PreviewUser = await this.previewUserService.findOne(userId, logger);
+    const previewUser: PreviewUser = await this.previewUserService.findOneFromCache(userId, logger);
     const { workingGroup } = userDoc;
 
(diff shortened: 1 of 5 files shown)


backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -918,6 +918,14 @@ export class ExperimentAssignmentService {
   }
 
   public async dataLog(userDoc: RequestedExperimentUser, jsonLog: ILogInput[], logger: UpgradeLogger): Promise<Log[]> {
+    if (!userDoc) {
```
bcb37, line 921: This test will not filter out preview users, since you can /init them and they'll have a userDoc

danoswaltCL:
@bcb37 in AWS I can see this call is extremely negligible, 17 calls per second, 0.01 ms per call. There are 2 rows of data being queried and unlikely to ever go up much more than that. 

Do we want to spend more time on this PR, let it go, merge it as is if it seems safe?

bcb37:
> @bcb37 in AWS I can see this call is extremely negligible, 17 calls per second, 0.01 ms per call. There are 2 rows of data being queried and unlikely to ever go up much more than that.
> 
> Do we want to spend more time on this PR, let it go, merge it as is if it seems safe?

I'd say t's safe; the cache stuff makes sense and seem to work. If we want to revisit getting previewing working we may have to make more changes anyway.

PR #2324, opened 2025-04-02
create separate card for global exclude segments

bcb37:
- fetches global exclude segments separately, using the new endpoint
- populates a separate segment table card, above the regular segment list table card, with the global exclude segments
- the global exclude card is collapsed by default
- global exclude table is sortable but not searchable
- "Show Global Excludes" removed from UI

### DIFF
--- a/frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
+++ b/frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
@@ -225,8 +225,15 @@ export interface SegmentState extends EntityState<Segment> {
   sortAs: SORT_AS_DIRECTION;
 }
 
+export interface GlobalSegmentState extends EntityState<Segment> {
+  isLoadingGlobalSegments: boolean;
+  sortKey: SEGMENT_SORT_KEY;
+  sortAs: SORT_AS_DIRECTION;
+}
+
 export interface State extends AppState {
   segments: SegmentState;
+  globalSegments: GlobalSegmentState;
 }
 
 export interface SegmentFile {
--- a/frontend/projects/upgrade/src/app/core/segments/store/segments.selectors.ts
+++ b/frontend/projects/upgrade/src/app/core/segments/store/segments.selectors.ts
@@ -57,20 +57,24 @@ export const selectFeatureFlagSegmentsExclusion = createSelector(
   (state) => state.allFeatureFlagSegmentsExclusion
 );
 
-export const selectSelectedSegment = createSelector(
-  selectRouterState,
+export const selectAllSegmentEntities = createSelector(
   selectSegmentsState,
   selectGlobalSegmentsState,
-  (routerState, segmentState, globalSegmentState) => {
-    if (routerState?.state && (segmentState?.entities || globalSegmentState?.entities)) {
+  (segmentState, globalSegmentState) => ({
+    ...segmentState.entities,
+    ...globalSegmentState.entities,
+  })
+);
+
+export const selectSelectedSegment = createSelector(
+  selectRouterState,
+  selectAllSegmentEntities,
+  (routerState, allSegmentEntities) => {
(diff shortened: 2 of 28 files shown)


frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
```diff
@@ -225,8 +225,15 @@ export interface SegmentState extends EntityState<Segment> {
   sortAs: SORT_AS_DIRECTION;
 }
 
+export interface GlobalSegmentState extends EntityState<Segment> {
```
danoswaltCL, line 228: This is interesting. It's not exactly a pattern we've used, having 2 feature-level states and reducers that share the same ngrx files. I think it makes sense though, it's separate enough in that it's loaded up and presented independently, but there's obvious major overlap and I don't think anyone's going to get confused. I guess I have nothing productive to say, just needed to type and talk it out, looks good 👍

frontend/projects/upgrade/src/app/core/segments/store/segments.selectors.ts
```diff
@@ -50,12 +60,17 @@ export const selectFeatureFlagSegmentsExclusion = createSelector(
 export const selectSelectedSegment = createSelector(
```
danoswaltCL, line 60: We can do this without the nested ternary logic if we have a new selector like `selectAllSegmentEntities` return all segments. Then the existing `selectSelectedSegment` can just swap in this selector in place of `selectSegmentsState` without changing the logic that's there.

I have a feeling we're going to want to be able to select all global + regular segments for other purposes as well that should be a standard selector in our collection.

```typescript
export const selectAllSegmentEntities = createSelector(
  selectSegmentsState,
  selectGlobalSegmentsState,
  (segmentState, globalSegmentState) => ({
    ...segmentState.entities,
    ...globalSegmentState.entities
  })
);
```

```typescript
export const selectSelectedSegment = createSelector(
  selectRouterState,
  selectAllSegmentEntities,
  (routerState, allEntities) => {
    if (routerState?.state && allEntities) {
      const {
        state: { params },
      } = routerState;
      return allEntities[params.segmentId];
    }
    return undefined;
  }
);
```

frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-root-page/segment-root-page-content/segment-global-section-card/segment-global-section-card-table/segment-global-section-card-table.component.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  get SEGMENT_TRANSLATION_KEYS() {
+    return SEGMENT_TRANSLATION_KEYS;
+  }
+
+  get SEGMENT_ROOT_COLUMN_NAMES() {
+    return SEGMENT_ROOT_COLUMN_NAMES;
+  }
+
+  changeSorting(event) {
```
danoswaltCL, line 66: I don't know if others have the same mania I do for slaying redundant code, `changeSorting` and `sortTable` are implemented multiple times the same way, these can be made `common-table-helper` utilities or some other shared way to reduce it at least in the 'refreshed' features.

there's a lot of copy pasta around table functionality though, not saying that needs to be a part of this PR. it may just be too baked in this way to be worth refactoring now, but I'd like to create a ticket for a rainy day.

bcb37: Yeah, I considered trying to abstract the sorting (since we didn't want search here too), but I concluded that it probably would have required a lot of changes in a lot of places.

danoswaltCL (approved):

PR #2352, opened 2025-04-09
remove github workflows for beanstalk account and old versioning scripts

danoswaltCL:
:fire:


shpwe:
how are the java libs getting built is deleting these? we dont have a jenkins job doing any java builds as of yet. `.github/workflows/clientlibs-java-build.yml`

danoswaltCL:
Right... I have always built the java one myself and uploaded to nexus and I did it the other day to upload to codeartifact, I never actually used this github workflow job.

There's a javascript/node npm package (clientlibs-js) too with a build/publish workflow here... That gets sent to the public npm repository because of the open-source nature (for some reason we only published the java one internally).

We haven't had to publish a new version in a long while, I don't remember if I used the jobs here to build it but I guess I did for publishing to the main npm registry because it requires a token in the secrets that I'm not privy to. So maybe I should leave that one...

shpwe:
> Right... I have always built the java one myself and uploaded to nexus and I did it the other day to upload to codeartifact, I never actually used this github workflow job.
> 
> There's a javascript/node npm package (clientlibs-js) too with a build/publish workflow here... That gets sent to the public npm repository because of the open-source nature (for some reason we only published the java one internally).
> 
> We haven't had to publish a new version in a long while, I don't remember if I used the jobs here to build it but I guess I did for publishing to the main npm registry because it requires a token in the secrets that I'm not privy to. So maybe I should leave that one...

ok, well if there is no need to continue publishing them publicly and if we are fine with not having this to publish the internal package, then this cleanup is fine. Else, i would leave those for now and only remove them once we have a replacement in jenkins. Up to you.

bcb37:
> imma leave the js client publish one then, we're still going to want to publish to npm registry. and this auto-merge dependabot one I'll leave in here also...
> 
> @bcb37 I'm not sure I've noticed this auto-merge job actually working as we definitely have to merge in some... but maybe we're not seeing all of them because this job is merging most of them automatically. I'd like to keep that in place if it is working, that was the recommendation Mike Fugate made way back when I first started and he set this stuff up. If it's not working then we can decide if it should go or be fixed...

I haven't seen auto-merge work either. Do those prs automatically get the approval by Mike Fugate with the message '@dependabot merge' which then gets a thumbs up emoji from dependabot? Every one of them has that exactly, but I don't think it ends up merging them, for whatever reason.

bcb37 (approved):

PR #2357, opened 2025-04-09
more dead code removal

danoswaltCL:
environment.json setup replaced by new infrastructure setup, feature-flags is permanent now so no longer needs a frontend toggle, key-value bubble is just a custom clunky debugger thing that has gone unused.


frontend/projects/upgrade/src/app/app.module.ts
```diff
@@ -69,10 +40,6 @@ export const getEnvironmentConfig = (http: HttpClient, env: Environment) => {
   ],
   providers: [
     { provide: ENV, useValue: environment },
-    provideAppInitializer(() => {
-      const initializerFn = getEnvironmentConfig(inject(HttpClient), inject(ENV));
```
bcb37, line 73: Both `inject()` and `provideAppInitializer()` are now unused and can be removed from the imports

bcb37 (commented):
Just a couple nits.

bcb37 (approved):

Issue #2334, opened 2025-04-04
Export Segment Design Modal


danoswaltCL:
When clicking on the Export Segment Design option in the details overview actions menu, it should open a confirmation modal. On confirm click, the segment json file should be downloaded in browser and the modal should close.

PR #2365, opened 2025-04-10
export segment design modal

danoswaltCL:
#2334


bcb37 (changes requested):
This doesn't include the lists' members.
We should either include the members (like 'Duplicate' does - and like export does for Feature Flags) or zero out the subSegments array.

bcb37 (approved):

PR #2367, opened 2025-04-10
mooclet: fixes for marking and getting persistant assignment

danoswaltCL:
This adds "learner" to the `/value` (reward) call and adds a method to handle marking condition.

We will always "mark" and enroll the actual condition that was sent in request (if valid), like in within-subjects. That has implications for outside factors potentially causing an imbalance if one condition for some reason is more likely to get marked in client. But there's not a good way around that I can find.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -561,7 +561,8 @@ export class ExperimentAssignmentService {
       },
       exclusionReason,
       status,
-      condition
+      condition,
+      logger
     );
     if (experiment.enrollmentCompleteCondition) {
       await this.checkEnrollmentEndingCriteriaForCount(experiment, logger);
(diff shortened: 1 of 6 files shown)


bcb37 (commented):
This all works and makes sense with one exception:
I still don't see the result of calling ../run?learning=[userId] ever settling on a consistent response.
- I enrolled a learner, then sent a reward and saw that reward added to the 'values' table with the correct learner and version. But I was still getting different responses when I hit the 'run' endpoint with that user id a bunch of times. 

This might not matter because, as you point out, we'll never hit the 'run' endpoint after a learner has marked, even if /assign is called - because then it gets the assignment from the enrollment table. And it does store the association between the learner (user) and the condition, but the consistency of that is actually coming from the UpGrade database (i.e. the params will only ever contain the condition from the enrollment table when the reward call is made).

danoswaltCL:
`/run?learning=[userId]` is this the line you're running or a typo in the comment? should be `learner`.

bcb37:
> `/run?learning=[userId]` is this the line you're running or a typo in the comment? should be `learner`.

typo

danoswaltCL:
@bcb37 amended my comments, please re-review at your convenience

bcb37 (approved):

PR #2371, opened 2025-04-11
Implement Delete Segment Modal

### LINKED ISSUE #2332: Segments new Delete Modal
When clicking delete, we should see the common delete confirmation modal. This should hit the existing delete endpoint.

After delete, we should nav back to segments root page and the segments list should be refetched.

zackcl:
Resolves #2332


bcb37 (approved):
This looks ok. It's copy-pasted from the feature flag one, which makes me wonder if we can generalize these any more before we make a third one for experiments.
Say we want to change something about how these modals look or feel in the future, we'd have to go in and change all three. But maybe the common modal component is general enough?

PR #2395, opened 2025-04-16
Fix parameter handling and pagination filter in segment/flag effects

### LINKED ISSUE #2394: Improper subscription handling and parameter misalignment in effects
6.1.0

## Problem
The current implementation of feature flags and segments effects has two issues:

1. Subscription management: Using direct `.subscribe()` inside switchMap without proper cleanup
2. Parameter misalignment: Filter function parameters don't match the values provided by withLatestFrom

## Impact
These issues may lead to:
- Potential memory consumption issues as the app runs over time
- Incorrect filtering logic when deciding whether to fetch more data
- Unpredictable behavior when search parameters change

## Affected Components
- segments.effects.ts (fetchSegmentsPaginated$ effect)
- feature-flags.effects.ts (fetchFeatureFlags$ effect)

## Proposed Fix
Replace manual subscription with proper reactive pattern using withLatestFrom and update parameter ordering in filter functions.

zackcl:
Resolves #2394 

This PR fixes two critical issues in segment and feature flag effects:

1. Parameter misalignment in filter functions that could cause incorrect pagination behavior
2. Replaces improper subscription inside switchMap with proper reactive pattern using withLatestFrom

### DIFF
--- a/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
+++ b/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.effects.ts
@@ -15,6 +15,7 @@ import {
   selectSortAs,
   selectSearchString,
   selectIsAllFlagsFetched,
+  selectFeatureFlagPaginationParams,
 } from './feature-flags.selectors';
 import { selectCurrentUser } from '../../auth/store/auth.selectors';
 import { CommonExportHelpersService } from '../../../shared/services/common-export-helpers.service';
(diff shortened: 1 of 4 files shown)


danoswaltCL:
Thank you, I've been looking at the `getSegmentString` thing for years wanting to fix it, this pattern is used in experiments as well.

Even better would a be a single selector, would you mind trying this?

```typescript
// segments.selectors.ts
export const selectPaginationParams = createSelector(
  selectSkipSegments,
  selectTotalSegments,
  selectSearchKey,
  selectSortKey,
  selectSortAs,
  selectAreAllSegmentsFetched,
  selectSearchString
  (skip, total, searchKey, sortKey, sortAs, areAllFetched) => ({
    skip,
    total,
    searchKey,
    sortKey,
    sortAs,
    areAllFetched,
    searchString
  })
);

// segments.effects.ts can then be simplified even further with just one select
withLatestFrom(this.store$.pipe(select(selectPaginationParams))),
```




This also exists in experiments as well, but we don't need to do that now...

zackcl:
@danoswaltCL I resolved your comment.

danoswaltCL (approved):

PR #2401, opened 2025-04-16
global excludes list operation actions

danoswaltCL:
This resolves issues with the global exclude list operations not updating after adding the list (store wasn't being updated). It also fixes the import list modal load spinner in the global exclude page.

This borrows add/update/delete segment list actions in GlobalExclude state in order to add the segment changes back to the correct store, and refactors to avoid the redundancy between the two reducers.

Having a separate globalExclude entity state is starting to look less ideal, this feels like a strange thing to have to do but it will have to do the trick for now without larger refactor.

### DIFF
--- a/frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
+++ b/frontend/projects/upgrade/src/app/core/segments/store/segments.model.ts
@@ -249,7 +249,7 @@ export interface SegmentState extends EntityState<Segment> {
 }
 
 export interface GlobalSegmentState extends EntityState<Segment> {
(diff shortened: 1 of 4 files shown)


zackcl (approved):

bcb37 (changes requested):
Edits to the Global Exclude segments (besides list changes) are not reflected in the UI.
Also, it looks like editing the Global Exclude segments moves them to the regular segments menu (in fact the type is changed from 'global_exclude' to 'public'

bcb37 (approved):

PR #2403, opened 2025-04-17
import segments with validated lists

bcb37:
This fixes imports so that, in the case of a new-style segment, each list is validated as a full new segment.

### DIFF
--- a/backend/packages/Upgrade/src/api/controllers/validators/SegmentInputValidator.ts
+++ b/backend/packages/Upgrade/src/api/controllers/validators/SegmentInputValidator.ts
@@ -1,5 +1,6 @@
 import { Type } from 'class-transformer';
 import { IsArray, IsEnum, IsNotEmpty, IsOptional, IsString, IsUUID, ValidateNested } from 'class-validator';
+import { SegmentValidator } from 'src/api/DTO/ExperimentDTO';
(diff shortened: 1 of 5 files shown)


danoswaltCL:
will spend some quality time with this one tomorrow

danoswaltCL:
In UI, an "old" style imported list shows nothing for the type... which makes sense, dunno if it's worth thinking of something to put there if this is a use-case that will come up.

image

I wanted to see the values for the old-style list, but fascinatingly, when you click edit on an "old-style" list, it opens up an edit view that looks like the "add" modal, yet with the Edit title. And if you enter a type and fill out values, it saves it and replaces the segment as a "new-style" segment, I think.

Anyway, this is a bug in UI, we need to be able to edit the old-style list in some way. something we hadn't considered I guess. not sure what the expected behavior should be here, launching the old-style modal from here would be interesting to try...

image

danoswaltCL:
I went through several scenarios trying to break things, using the imports provided and otherwise, so far just the comments for the above thing is the only thing I noticed, but that's a UI thing. I'm seeing the correct behavior from validation/import as far as I can tell, I've run out of time though will continue on monday.

danoswaltCL (approved):

PR #2406, opened 2025-04-17
replace fetchall segment upsert behavior with listSegmentOptions store and selectors

danoswaltCL:
This was in response to comments on https://github.com/CarnegieLearningWeb/UpGrade/pull/2396 and this convo https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1744900743288969.

The idea is that although we now use a paginated call to gather proper 'new-style' segment data, we still have a use for getting all segment data for populating list segment selections. However, continuing to upsert these incomplete segments into the main entity store no longer makes sense.

This instead repurposes the fetch-all with a new store value `listSegmentOptions` that contains just id, name, and context of each segment, with a selector to get all by context. It will be called wherever a modal in the new or old experience needs a dropdown with these items to select a segment in that context.

For legacy usage:
1) this meant a small refactor of the "old" experiments modal for participants and the old segment edit modal, since it will still be accessible. These depended on fetch-all from the root page which was removed. Luckily it works fine just replacing the source and tweaking the timing of the fetch.

2) There are still 2 places that call fetch-all in the old manner, I left those in there because if the segments toggle is off technically these would still need that functionality, but they are not accessible in the new view and will be easy to remove for good once we are committed to the new UI.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/segments-legacy/components/modal/import-segment/import-segment.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/segments-legacy/components/modal/import-segment/import-segment.component.ts
@@ -45,7 +45,6 @@ export class ImportSegmentComponent {
         this.segmentDataService.importSegments(this.fileData)
       )) as importError[];
       this.showNotification(importResult);
-      this.segmentsService.fetchAllSegmentListOptions();
     } catch (error) {
       console.error('Error during segment import:', error);
     }
--- a/frontend/projects/upgrade/src/app/features/dashboard/segments-legacy/components/modal/import-segment/import-segment.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/segments-legacy/components/modal/import-segment/import-segment.component.ts
@@ -45,7 +45,7 @@ export class ImportSegmentComponent {
         this.segmentDataService.importSegments(this.fileData)
       )) as importError[];
       this.showNotification(importResult);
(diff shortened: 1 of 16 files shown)


frontend/projects/upgrade/src/app/features/dashboard/segments-legacy/components/modal/import-segment/import-segment.component.ts
```diff
@@ -45,7 +45,7 @@ export class ImportSegmentComponent {
         this.segmentDataService.importSegments(this.fileData)
       )) as importError[];
       this.showNotification(importResult);
-      this.segmentsService.fetchAllSegments(true);
+      this.segmentsService.fetchAllSegmentListOptions();
     } catch (error) {
       console.error('Error during segment import:', error);
     }
```
zackcl, line 51: Just curious, why do we need to fetch segments when the "Import Segment" button is clicked?

danoswaltCL: removed, not sure why it used to do fetchAllSegments either, but it works fine without any fetch

danoswaltCL:
@zackcl removed the import modal fetch, you're right it wasn't needed. also fixed a small thing for adding a new segment in the old modal. both items are things that won't be accessible in the new experience, but are good to have working just in case we have to toggle it off.

zackcl (approved):

PR #2413, opened 2025-04-18
fetch and use inclusion/exclusion data for single segments

bcb37:


frontend/projects/upgrade/src/app/core/segments/store/segments.reducer.ts
```diff
@@ -179,6 +198,8 @@ const globalReducer = createReducer(
     adapter.upsertMany(globalSegments, { ...state, isLoadingSegments: false })
   ),
   on(SegmentsActions.actionGetSegmentByIdSuccess, (state, { segment }) => {
+    // const segment = segments[0];
+
```
zackcl, line 202: Maybe forgot to remove this.

danoswaltCL (approved):

danoswaltCL:
I see the used by stuff when refresh on the details page, and the expected stuff in the store when i spy the redux tools. I think that was the test for this, maybe Zack can confirm, looks good to me.

PR #2421, opened 2025-04-21
Truncate Lengthy Tags on Root/Details Pages

### LINKED ISSUE #2420: A Lengthy Tag Stretches the Tags Column Width
If the defined tags have reasonable length, they are automatically displayed in multiple lines and doesn't really affect the column width.

Image

However, if one of the tags is very lengthy (e.g., over 100 characters), then it stretches the Tags column, affecting the whole table:

Image

The same happens on the Feature Flags root page:

Image

We should not allow the "Tags" column width being stretched by a lengthy tag, even if the lengthy tag is truncated. Feel free to make suggestions.

zackcl:
Resolves #2420

This PR truncates tags longer than 16 characters and displays the full tag string in a tooltip when truncated. This prevents long tags from affecting the column width of the root table and keeps the layout clean.

Screenshots after change:

Screenshot 2025-04-22 at 1 58 09 AM
Screenshot 2025-04-22 at 1 58 22 AM

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.html
@@ -79,20 +79,11 @@
         {{ FLAG_TRANSLATION_KEYS.TAGS | translate }}
       </th>
       <td mat-cell *matCellDef="let flag" class="tags-column ft-14-400 dense-2">
-        <mat-chip-listbox aria-label="Flag tags">
-          <mat-chip
-            *ngFor="let tag of flag.tags"
-            class="tag"
-            (click)="filterFeatureFlagByChips(tag, FeatureFlagSearchKey.TAG)"
-          >
-          <span
-            class="chip-label"
-            [matTooltip]="tag.length > 16 ? tag : null"
-            matTooltipPosition="above">
-              {{ tag | truncate : 16 }}
-          </span>
-          </mat-chip>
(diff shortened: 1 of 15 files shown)


danoswaltCL (approved):

danoswaltCL (changes requested):
works of course but we'll have to remember to add each time in future when using mat-chip... this is prob the official way we want to handle long tag text throughout so we can add a common-tag component to wrap this logic really simply, i tried this out real quick to convince myself it works

```
@Component({
  selector: 'app-common-tag',
  standalone: true,
  imports: [CommonModule, MatTooltipModule, SharedModule],
  template: `
    <span
      class="chip-label"
      (click)="onTagClick()"
      [matTooltip]="tag.length > 16 ? tag : null"
      matTooltipPosition="above"
    >
      {{ tag | truncate : 16 }}
    </span>
  `,
})
export class CommonTagComponent {
  @Input() tag = '';
  @Output() tagClick = new EventEmitter<string>();

  onTagClick(): void {
    this.tagClick.emit(this.tag);
  }
}
```

```
// tagClick is an optional callback since we have that use-case
<app-common-tag 
            [tag]="tag" 
            (tagClick)="searchByTag($event)">
          </app-common-tag>
```

zackcl:
@danoswaltCL I added `CommonTagComponent` and `CommonTagListComponent` similar to what you suggested, but included the `mat-chip` element in the component, not just the `span`.

I also updated `CommonTagsInputComponent` to use truncated tags. I personally think it's much cleaner to enter long UUIDs into the Values field when the tags truncate to 16 characters (we can modify the character limit if needed).

danoswaltCL (approved):
cool makes sense to me. if no styles are attached to the component it can just be omitted and scss file deleted, no use in an empty file.

zackcl:
>  if no styles are attached to the component it can just be omitted and scss file deleted, no use in an empty file.

@danoswaltCL We currently have empty `scss` files everywhere in our codebase. I personally think it's OK to keep them and sometimes it's easier for me to quickly apply styles to the existing files. Maybe we could consider cleaning these up later once the component designs are more finalized.

PR #2426, opened 2025-04-22
Refine Global Exclude Details Page Labels

### LINKED ISSUE #2397: Change Global Exclude list tab and card header title to Exclude Lists
The Design for Global Exclude tab was supposed to say ~~"Excluded Lists"~~ "Exclude Lists" instead of "Lists".

Image

zackcl:
Closes #2397

This PR refines the labels for tabs, buttons, menu items, and modals on the Global Exclude details page to differentiate them from the Segment details page.

## After Change:
Screenshot 2025-04-23 at 1 05 35 AM

## Before Change:
Screenshot 2025-04-23 at 1 10 02 AM


frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -260,9 +260,9 @@ export class DialogService {
     return this.openUpsertPrivateSegmentListModal(commonModalConfig);
   }
 
-  openAddListModal(appContext: string, segmentId: string) {
+  openAddListModal(appContext: string, segmentId: string, segmentType: SEGMENT_TYPE) {
```
danoswaltCL, line 263: you can just pass in the title directly instead of conditionalizing based on type

frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-details-page/segment-details-page-content/segment-lists-section-card/segment-lists-section-card.component.ts
```diff
@@ -65,16 +67,39 @@ export class SegmentListsSectionCardComponent {
...
+    this.title = config.title;
+    this.subtitle = config.subtitle;
+    this.buttonText = config.buttonText;
+    this.menuButtonItems[0].label = config.importLabel;
+    this.menuButtonItems[1].label = config.exportLabel;
   }
 
   onAddListClick(appContext: string, segmentId: string) {
-    this.dialogService.openAddListModal(appContext, segmentId);
+    this.dialogService.openAddListModal(appContext, segmentId, this.data.type);
```
danoswaltCL, line 102: for all of these, you can pass in a modal title directly instead of making the modal figure it out based passing in a type, you've got a convenient object `segmentTypeConfig` that is doing that kind of thing already.

zackcl:
@danoswaltCL Passing in the title directly would indeed remove the condition and make the code look cleaner, but we currently maintain all modal configurations in `common-dialog.service.ts`, and I wanted to preserve this consistency (e.g., when we need to change modal titles, we know to look in that file).

While the properties in `segmentTypeConfig` are specific to the `SegmentListsSectionCardComponent`, the modals themselves could be opened from multiple different components. I think it makes more sense to keep the modal configurations centralized in the dialog service as they are now, rather than having the calling component determine the specific parameters.

Let me know if you see it differently or have other considerations I might have missed.

zackcl:
I think i will also update these menu options to "Edit Global Exclude", "Duplicate Global Exclude", and "Export Global Exclude" if you all agree.

Screenshot 2025-04-23 at 10 21 32 PM

frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -397,40 +416,82 @@ export class DialogService {
     return this.dialog.open(DeleteSegmentModalComponent, config);
   }
 
-  openExportDesignModal(title: string, warning: string): MatDialogRef<CommonSimpleConfirmationModalComponent, boolean> {
+  openExportFeatureFlagDesignModal(): MatDialogRef<CommonSimpleConfirmationModalComponent, boolean> {
```
danoswaltCL, line 419: yeah I'm cool with an explicit separate method for each type of modal, i agree it's a good spot to be repetitive and keep the config options clearly defined here vs passing in the data.

zackcl:
I will work on #2432 once this is merged.

danoswaltCL (approved):

PR #2427, opened 2025-04-22
fix segments table UI pagination issues

danoswaltCL:
Sharing this to show some things if we want to bandaid this to mimic the experiments table behavior.

On sort or filter click, it will perform a call to pagination endpoint, with (fromStarting = true) to set the skip back to 0 before fetching.

This will mimick experiment list behavior so we're consistent, maybe that's good enough for now. But the pattern for fetching data on sort/filter change in experiments seems completely wrong/unintuitive to me also, so I'm not sure this is much of an improvement. some convo here: https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1745269308177459

### DIFF
--- a/frontend/projects/upgrade/src/app/core/segments/store/segments.actions.ts
+++ b/frontend/projects/upgrade/src/app/core/segments/store/segments.actions.ts
@@ -29,6 +29,7 @@ export const actionFetchSegmentsSuccess = createAction(
     experimentSegmentExclusion: experimentSegmentInclusionExclusionData[];
     featureFlagSegmentInclusion: featureFlagSegmentInclusionExclusionData[];
     featureFlagSegmentExclusion: featureFlagSegmentInclusionExclusionData[];
+    fromStarting?: boolean;
   }>()
 );
 
--- a/frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
+++ b/frontend/projects/upgrade/src/app/core/segments/store/segments.effects.ts
@@ -68,9 +68,7 @@ export class SegmentsEffects {
         }
         return this.segmentsDataService.fetchSegmentsPaginated(params).pipe(
           switchMap((data: any) => {
-            const actions = fromStarting ? [SegmentsActions.actionSetSkipSegments({ skipSegments: 0 })] : [];
             return [
-              ...actions,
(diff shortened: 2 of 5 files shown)


bcb37:
One subtle issue this still has is that, if you add a filter that would result in fetching segments that are not already loaded, the filter applies first in the UI, which empties the list and triggers the intersection observer. That results in a fetch with 'skip' incremented.

To reproduce:
- Add more than 20 segments (say 25) making sure that no more than 5 have names that start with 'z'
- Freshly load the segments page with sort set to 'name', so that there are no 'z'-named segments loaded yet
- type a 'z' in the search box

bcb37:
...But the pattern for fetching data on sort/filter change in experiments seems completely wrong/unintuitive to me 
Do you mean fetching on sort/filter is unintuitive, or the way we're doing it is unintuitive?

In order to support pagination you have to fetch on sort/filter. Imagine you have 1000 segments and you want to filter to just the ones that begin with 'Z'. It's unlikely you've already loaded all of those in the first fetch (which gets the first 20, with default sorting). You need to re-run the query with the filter applied, so it gets the first 20 with names that start with 'Z'. The same logic applies to sorting.

danoswaltCL:
understood, that's not what's happening. when I filter on a string that is not present in any of the entries, the pagination response is returning items that I would expect should be filtered out. You can type nonsense into the filters for anything and they all send back a response with a bunch of stuff.

image

image

image

I can reproduce in postman also, say I have a table with 2 total segments, one named "a" and one named "b", and send a request like: `{"skip":0,"take":2,"sortParams":{"key":"name","sortAs":"ASC"},"searchParams":{"key":"name","string":"a"}}`, I would expect to just get 1 item back, the item with a "b", but I get both. Is that expected? That's what I don't understand if so.

It's unintuitive that we upsert a filtered dataset into the store instead of replacing what's in there. There are bugs with the skip value in experiments related to this. It's super confusing to have to compare what the backend is sending vs. what the UI has stored vs. what the table actually shows because with a mix-and-match grab-bag of items loaded in, the UI has to take over the filtering duties. I just have no idea what is "supposed" to be happening with our pagination system.

bcb37 (approved):

PR #2434, opened 2025-04-23
Fix/2376 duplicate name and status

bcb37:
Fixes two bugs found QAing old style segment regressions:
- Duplicate was broken because new old-style segment were not getting a value for 'status': fixed by adding 'Unused' on create in the effect
- It's possible to have two segments with the same name by editing one of them: fix by checking for duplicates on edit as well as create.


danoswaltCL (approved):
ah, good catch(es), makes sense to me

PR #2441, opened 2025-04-24
segments: search only substring

bcb37:
- Switches from tokenizing, lexeme-based search to simple substring-based search for segments root page.
- Omit private segments when looking for duplicates on edit/add segment.
- Feature flag search: remove tokenization of search string - to support spaces and special characters.

Note: the change to simple substring-based search is not implemented for feature flags or experiments.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/SegmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/SegmentService.ts
@@ -214,7 +214,6 @@ export class SegmentService {
     sortParams?: ISegmentSortParams
   ): Promise<getSegmentsData> {
     logger.info({ message: `Find paginated segments` });
-    let customSearchString = '';
     let paginatedParentSubQuery = this.segmentRepository
       .createQueryBuilder()
       .subQuery()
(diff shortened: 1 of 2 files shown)


backend/packages/Upgrade/src/api/services/SegmentService.ts
```diff
@@ -1132,9 +1126,13 @@ export class SegmentService {
     // Check if the segment name already exists in the same context
     // If id is present, check only for other segments with the same name and context
     if (id) {
-      sameNameSegment = await this.segmentRepository.find({ where: { name, context, id: Not(id) } });
+      sameNameSegment = await this.segmentRepository.find({
+        where: { name, context, type: Not(SEGMENT_TYPE.PRIVATE), id: Not(id) },
```
danoswaltCL, line 1130: ok, this is because private segments can be named same thing right? makes sense.

bcb37: yep, and we don't to be able to create a regular segment with the same name as a global exclude one

zackcl (approved):

PR #2447, opened 2025-04-24
handle all bad segment imports

bcb37:
Handle the following incompatible segment imports:
-valid JSON without 'subSegments'
- empty JSON file
- malformed JSON


zackcl:
I tired it and works as expected!

zackcl (approved):

PR #2453, opened 2025-04-25
fix attribute case; don't create extra queries on mooclet exp. import

bcb37:
- corrects the casing 'currentPosteriors' -> 'current_posteriors'
- avoids creating duplicate queries for the reward metric on import

### DIFF
--- a/backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
@@ -207,11 +207,12 @@ export class MoocletExperimentService extends ExperimentService {
       });
       throw error;
     }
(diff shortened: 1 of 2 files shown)


backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -207,11 +207,12 @@ export class MoocletExperimentService extends ExperimentService {
       });
       throw error;
     }
+    if (!queries.some((query) => query?.metric?.key === rewardMetricKey)) {
```
danoswaltCL, line 210: got it, when importing, this will obviously not need to be re-added. makes sense to me.

danoswaltCL (approved):

PR #2473, opened 2025-05-07
Disable Import Button After Click to Prevent Multiple Requests

### LINKED ISSUE #2471: Import button not disabled, no spinner when pending large import
Import button not disabled, no spinner when pending large import, which allows the user to keep sending requests.

To reproduce, try importing this segment, which has about 15000 members.

Requirement: Bugfix for this release, the import button needs to be disabled after being clicked to prevent multiple requests and show that it's doing something.

[H1_segment_part1.json](https://github.com/user-attachments/files/20061779/H1_segment_part1.json)

zackcl:
Resolves #2471


zackcl:
To answer @danoswaltCL question, the "isLoadingImport$" is used when uploading and validating files, not when the files are being sent to the server for import. That's why we need a separate "isImporting$".

zackcl:
@danoswaltCL I think what you suggested makes more sense. I updated the code to reuse the existing `isLoadingImport$` for simplicity.

danoswaltCL:
ok cool yes great that's the direction I was thinking, or at least intending when I coded it up, glad that works.

danoswaltCL (approved):

PR #2481, opened 2025-05-08
[WIP] Make Tags Work Consistently

### LINKED ISSUE #2465: Experiment tag truncate/tooltip
Long tags don't truncate in experiments, we truncate and tool-tip it now in the new components. Maybe we can here also.

zackcl:
Resolves #2465

Still need to work on the experiment details page.

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-list/experiment-list.component.html
@@ -262,8 +262,12 @@
                 *ngIf="tagIndex < 2; else moreTagsTemplate"
                 (click)="filterExperimentByChips(tag, ExperimentSearchKey.TAG)"
               >
-                <span class="chip-label">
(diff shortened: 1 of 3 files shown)


bcb37 (commented):
This looks ok to me. A question though: The way it currently works with experiments (e.g. on the list page) the expansion is not reset when sorting or filtering is changed. With the new common component, If you click to expand the tags list, then click one of the tags that had been invisible, it will apply the filter to the list and re-hide the tag. Probably not a big deal but it could be confusing.

zackcl:
@bcb37 I applied the change to preserve the tag expansion state when sorting or filtering is changed (also when the section card expansion is changed).

danoswaltCL:
nice, makes sense to me. I looked at the standalone question, it's for sure just some import issue, I think you could try pulling the imports directly into experiments module, I think circular dependencies and missing imports will both produce that kind of an error. But as you said, it isn't that important right now.

danoswaltCL (approved):

bcb37 (approved):

PR #2486, opened 2025-05-13
replace `save()` calls with `upsert()` for users

bcb37:


danoswaltCL:
looks good to me, just a question.. if `.upsert` returns no object or undefined, it's clearly some kind of unexpected result that should be reported, but it would be pretty weird right? for it to return nothing without erroring itself? I guess I'm wondering it the error should just say that explicitly, "upsert unexpectedly returned empty object"...?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v5.ts
```diff
@@ -289,15 +289,22 @@ export class ExperimentClientController {
...
-    return { id, group };
+    if (!udateResult) {
+      request.logger.error({
+        details: 'update unexpectedly returned empty object',
+      });
+      throw new InternalServerError('set group membership failed');
+    }
+
+    return { id: experimentUserDoc.id, group: experimentUser.workingGroup };
   }
```
zackcl, line 308: Shouldn't the `setGroupMemberShip` function return `experimentUser.group` instead of `experimentUser.workingGroup`?

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v5.ts
```diff
@@ -289,15 +289,22 @@ export class ExperimentClientController {
     // append userDoc in logger
     request.logger.child({ userDoc: experimentUserDoc });
     request.logger.info({ message: 'Got the original user doc' });
-    const { id, group } = await this.experimentUserService.updateGroupMembership(
-      experimentUser.id,
+    const udateResult = await this.experimentUserService.updateGroupMembership(
+      experimentUserDoc.requestedUserId,
```
zackcl, line 293: I think this should be `updateResult`.

backend/packages/Upgrade/src/api/controllers/ExperimentClientController.v5.ts
```diff
@@ -289,15 +289,22 @@ export class ExperimentClientController {
...
     );
-    return { id, group };
+    if (!udateResult) {
+      request.logger.error({
+        details: 'update unexpectedly returned empty object',
+      });
+      throw new InternalServerError('set group membership failed');
+    }
+
+    return { id: experimentUserDoc.id, group: experimentUser.workingGroup };
```
zackcl, line 307: Shouldn't the `setGroupMemberShip` function return `experimentUser.group` instead of `experimentUser.workingGroup`?

zackcl (approved):

PR #2490, opened 2025-05-15
Fix error response code for /assign requests with unknown app contexts

### LINKED ISSUE #2480: Fix error response code for /assign and /featureflag requests with unknown app contexts
Currently, when making a request to the `/assign` or `/featureflag` endpoint with an unknown app context (one not defined in `CONTEXT_METADATA`), the API responds with a `500 Internal Server Error`. This is not the appropriate error code for this situation, as it suggests a server-side problem rather than an invalid client request. 

The `/assign` and `/featureflag` endpoints should properly validate the app context parameter and return a `400 Bad Request` error when an unknown context is provided. The response should include a clear error message indicating that the specified app context is unknown or invalid.

zackcl:
Resolves #2480

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentAssignmentService.ts
@@ -844,6 +844,13 @@ export class ExperimentAssignmentService {
     // Fetch the global exclude segment for the context
     const globalExcludeSegment = await this.segmentService.getGlobalExcludeSegmentByContext(context);
 
+    if (!globalExcludeSegment) {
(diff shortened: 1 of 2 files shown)


bcb37 (approved):
This makes the source of truth for app contexts for /assign the existence of a global exclude segment, which is kind of a way of moving that from the context metadata to the UpGrade db.

zackcl:
For future reference, this change also affects the `/featureflag` endpoint since the `getKeys()` function also calls the `checkUserOrGroupIsGloballyExcluded` function defined in `ExperimentAssignmentService.ts`, where we check the app context validity with `getGlobalExcludeSegmentByContext`.

I've updated the PR title to reflect this.

PR #2494, opened 2025-05-19
Fix the last row not being fully visible in Manage Roles table

### LINKED ISSUE #2474: Last row is not fully visible in Manage Roles table
6.1.0

**Describe the bug**
The last row in the Manage Roles table is not fully visible (only 20% visible)

**To Reproduce**
1. Go to Profile page
2. Define at least 4 users
3. Scroll down to the bottom row
4. See that the last row is not fully visible

**Expected behavior**
The last row should be fully visible

zackcl:
Resolves #2474

### Before fix:
Screenshot 2025-05-20 at 1 00 41 AM

### After fix:
Screenshot 2025-05-20 at 1 00 13 AM

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/profile/components/profile-info/profile-info.component.scss
+++ b/frontend/projects/upgrade/src/app/features/dashboard/profile/components/profile-info/profile-info.component.scss
@@ -86,7 +86,6 @@ $role-text-color: #0a7319;
 
     .users-list-container {
       margin-top: 20px;
-      padding-bottom: 35px;
       overflow: auto;
 
       .users-list {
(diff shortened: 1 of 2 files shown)


bcb37 (commented):
Is there a reason we're not having the flex components inherit the 'flex-grow' attribute?
If we add `flex-grow: inherit` to both  .mat-mdc-tab-group and .mat-mdc-tab-body-wrapper we don't have to add extra padding that will be hidden.

bcb37:
> @bcb37 Like how? Can you show me the change you want? You can feel free to make a commit too.

@zackcl I added a commit to show the change I was talking about. It doesn't seem like it messes anything else up, and I think it will help if we ever have any other tab bodies with extra height on included elements.

zackcl:
Thanks for the commit! I didn’t realize it was caused by the tab body. Good to know that fix.

danoswaltCL (approved):

PR #2495, opened 2025-05-19
Re-add disabling auth check in local

### LINKED ISSUE #2455: re-add disabling auth check in local
There was a change made last year to remove auth check in [this pr](https://github.com/CarnegieLearningWeb/UpGrade/pull/1457), this seemed hasty and I still can't see a justifiable reason why, seems like we broke a more important thing (ability to disable auth in dev configurations) in order to fix a not important thing (usernames not appearing in audit logs) and have just dealt with it for a year.

Let's please re-eavaluate that and reimplement disabling token-checks for the non-client API.

This was likely only an issue in QA or Staging, where auth-check may have been disabled at times for convenience but because it used profile info provided by the signed-in google user credential, we would see mystery people doing things in the logs. That makes sense to avoid and we should keep requiring auth tokens required in our deployed envs, that seems fine.

For local config though, we should have the option to disable this.

Requirements:

- An env var should be re-introduced, formerly it was called `AUTH_CHECK` but that was ambiguous seeing as there are 2 different kinds of auth (client API token or google user token), so maybe we can call it `GOOGLE_AUTH_TOKEN_REQUIRED`.

- Update the .env files, `false` for local dev, `true` for the rest.

- When `false`, it should not throw errors or introduce bugs that removing auth-check "fixed".

zackcl:
Resolves #2455

I've almost simply reverted the changes made in this PR: https://github.com/CarnegieLearningWeb/UpGrade/pull/1457/files (only the necessary ones).

@danoswaltCL I have a question regarding this requirement:

> Update the .env files, false for local dev, true for the rest.

I'm not sure if you meant updating the `environment.ts` files so that `authTokenRequired` is set to `false` on `environment.local.ts` and `true` on other files. The variable is used in the backend though. I would appreciate your clarification.


backend/packages/Upgrade/src/env.ts
```diff
@@ -83,6 +83,7 @@ export const env = {
     clientId: getOsEnvArray('GOOGLE_CLIENT_ID'),
     serviceAccountId: getOsEnvArray('GOOGLE_SERVICE_ACCOUNT_ID'),
     domainName: getOsEnvOptional('DOMAIN_NAME'),
+    authTokenRequired: toBool(getOsEnvOptional('GOOGLE_AUTH_TOKEN_REQUIRED')),
```
bcb37, line 86: Should this default to `true`? When we deploy this change, the variable won't be defined in the environments.

zackcl: I added a commit to update `app-infrastructure.yml` so that `GOOGLE_AUTH_TOKEN_REQUIRED` will be set to `true` when deployed.

bcb37: We're not using `projectBuilder.jenkinsfile` for anything anymore are we?

zackcl: Do we no longer use the file? Where do we get the SSM parameters from then? Is it not part of our repository?

zackcl: @danoswaltCL Please check the PR and let me know what you think.

backend/packages/Upgrade/src/auth/authorizationChecker.ts
```diff
@@ -17,15 +18,15 @@ export function authorizationChecker(): (action: Action, roles: any[]) => Promis
     const token = authService.parseBasicAuthFromRequest(action.request);
     if (token === undefined) {
       log.warn({ message: 'No token provided' });
-      return false;
+      return env.google.authTokenRequired ? false : true;
```
danoswaltCL, line 21: ternary not needed, can just do `return env.google.authTokenRequired`

danoswaltCL, line 21: ternary not required, you can just do `return env.google.authTokenRequired`

danoswaltCL (approved):

PR #2503, opened 2025-05-22
remove joins from queries and use plain ids

bcb37:
- Removes the joins on other tables from `findEnrollments()` and `findExclusions()` methods for individuals and groups.
- Changes references to nested objects' Ids to the plain foreign key ids on the entities.

This intention of this slight increase in complexity is an optimization of queries that are made potentially multiple times per /assign call (they're also made in /init and /mark calls).

### DIFF
--- a/backend/packages/Upgrade/src/api/models/GroupEnrollment.ts
+++ b/backend/packages/Upgrade/src/api/models/GroupEnrollment.ts
@@ -13,6 +13,9 @@ export class GroupEnrollment extends BaseModel {
   @ManyToOne(() => Experiment, { onDelete: 'CASCADE' })
   public experiment: Experiment;
 
+  @Column({ name: 'experimentId' })
+  experimentId?: string;
+
   @Index()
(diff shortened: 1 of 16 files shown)


bcb37:
> bleh I don't feel qualified exactly to be doing TypeORM reviews, I can say that I have looked this over and code-wise all is well, and I like the premise, so I'm inclined to approve but I wonder how best to put this to test to make sure it's all good in QA?
> 
> this wouldn't require any migration? are we adding columns to these models?

No new columns. We're just adding the literal foreign keys to the entity types so we don't have to join on those tables and then get their ids. It might have been possible to use the @JoinColumn decorator to achieve this, like we've done elsewhere, but that seems to mess up the binding between tables and entities if you add them after the fact. [At least one other person](https://dev.to/arnauddenoyelle/double-bind-the-foreign-key-to-avoid-unnecessary-join-in-typeorm-29bf) seems to have hit on this idea recently and used it successfully. 
The only thing I'm worried about for this pr is whether I've missed instances where the altered methods are used that rely on the joined object being present in the returned entity, but I think I found them all.

Issue #2382, opened 2025-04-14
Evaluate ci/cd and precommit checks


danoswaltCL:
Now that we're in ECS it's a good time to reevaluate our quality-control checks to make sure they're doing what we want. Some things are definitely not great:

- Currently ci/cd checks are not running tests / typecheck.

They did for the beanstalk builds, we prob just need to tell it to run the commands we want in Jenkinsfile. They should only run on relevant code changes (i.e. frontend changes don't trigger backend tests, etc).

- Currently our pre-commit hooks are always running tests on everything, not just the relevant code.

Lint checks are at least limited to relevant code though. VSCode settings for eslint/prettier (at least for me) are not quite consistent with the lint checks, so sometimes the checks will "fix" and commit linting issues uncaught by editor that were introduced by other devs for whatever reason and we end up with commits that have files with lint-only changes. Not a huge deal but in bigger PRs it's unhelpful for reviewing purposes to have to ignore those diffs.

PR #2508, opened 2025-05-28
add missing test checks to jenkins, remove commit tests

danoswaltCL:
We didn't really refine this or chat much about [this story](https://github.com/CarnegieLearningWeb/UpGrade/issues/2382) but we have failing tests on `dev` and all i wanted to do was put up the PR to bump version to 6.2.

This adds missing test scripts as PR checks for backend and frontend, as well as typecheck.

I also removed the tests from the precommit hook because they are redundant if we are doing them as a PR check, and they are quite annoying to always run. Rather than "fix" them to only run on relevant project changes, I just left the linter check, which silently fixes things and does run only on relevant changes, that seems ok but I'm open to what else may want to do.

@shpwe @kcalvo-cli let me know if this is the proper way to go about this in Jenkinsfile.


danoswaltCL:
hm, do i see the frontend test showing up, and one of my commits to test the backend failed but now I don't see the job showing up, may be missing something about how it works for backend.

PR #2515, opened 2025-06-02
Make tag truncation configurable in CommonTagsInputComponent

### LINKED ISSUE #2439: feature request: ways to quickly peak at user lists from exp or FF details page
while testing FF adding public segments as private lists, I have found it myself wishing it was easier to view the members of this list without having to leave the FF view. non-segment private list members can at least be viewed by clicking edit, but I want something even simpler and view-only.

It would be handy to have an "eyeball" action, which could open a modal to peak at the members of the list, or in a new tab, no more no less.

It would also be very handy to have an eyeball for the entire participation list, so I can get a single view of the union of all raw member values that are explicitly included, or all that will be excluded.

Another idea or additional feature which i keep wanting is the ability to click on the segment and be taken to the segment details view of that segment, with the ability to nav back to previous page, or just opening in a new tab.

Ultimately the purpose is to assure that I know that I'm using the correct list(s), or that the combination of my lists will work as expected to include/exclude, so another idea that is related would be like the preview user view, be able to type in an id and immediately get feedback on if the user id or group id would be included or excluded (whether the flag or experiment is on or off).

zackcl:
Resolves #2439

Changes:
- Make `truncationLength` configurable (default: 16 characters)
- Update private segment list modal to use 72-character truncation for UUID searchability
- Maintain default 16-character truncation for segment and feature flag modals
- Preserve tooltip functionality for truncated content

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/modals/upsert-feature-flag-modal/upsert-feature-flag-modal.component.html
@@ -69,7 +69,6 @@
       [inputType]="CommonTagInputType.TAGS"
       [label]="config.tagsLabel"
       [placeholder]="config.tagsPlaceholder"
-      [enableTruncation]="true"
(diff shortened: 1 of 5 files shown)


bcb37 (approved):
lg

PR #2523, opened 2025-06-05
Hide '1 Values' for Segment type in the Lists table

zackcl:
Resolves #2521


zackcl:
1. This also applies to the feature flag lists, since it updates a common component (`common-details-participant-list-table.component.html`).

2. I don't think they're related. This PR just adds an extra conditional check to `common-details-participant-list-table.component.html`.

bcb37 (approved):

PR #2538, opened 2025-06-16
Feature/2524 experiment redesign backend 1

bcb37:
Initial PR for backend support for the experiment UI refresh:
- Removes 'filteredTotal' from the frontend, since there's no longer a need to support a splash page
- Changes the relationships between experiments and and exclusions/inclusions from one-to-one to one-to-many
- Moves the call to update enrollments and exclusions from the segment service to the new 'addList' call in the experiment service.
- Uses the new 'addList' call to create exclusion/inclusion lists on experiment creation (for imports)
- Renames 'FEATURE_FLAG_LIST_FILTER_MODE' to 'LIST_FILTER_MODE

### DIFF
--- a/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
+++ b/backend/packages/Upgrade/src/api/DTO/ExperimentDTO.ts
@@ -307,6 +307,10 @@ class SegmentValidator {
   @IsString()
   @IsEnum(SEGMENT_TYPE)
   public type: SEGMENT_TYPE;
+
+  @IsOptional()
+  @IsString()
+  public listType?: string;
 }
 export class ParticipantsValidator {
   @IsNotEmpty()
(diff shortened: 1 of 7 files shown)


zackcl:
In this branch, if I search for experiments using a non-existent string, the splash screen shows up and can't be dismissed unless I clear local storage. Maybe it's fine to leave it like this until we switch to the new root page? Just wanted to clarify.

bcb37:
> In this branch, if I search for experiments using a non-existent string, the splash screen shows up and can't be dismissed unless I clear local storage. Maybe it's fine to leave it like this until we switch to the new root page? Just wanted to clarify.

Yeah, I think we decided we wouldn't keep the new branch back-compatible to the old UI.

zackcl (approved):

PR #2541, opened 2025-06-18
ephemeral and merged user group options for featureflags endpoint

danoswaltCL:
I'll provide a demo of this or something! Please take a look.


bcb37 (approved):

bcb37:
> @bcb37 before I merge this, there's still time to try stuff, it's not a fire-drill yet:
> 
> After we talked I was looking at what it would be like to add a different id column to the FF exposures table so we can avoid creating a stub user just to record that id for a unique exposure, which i agree is not great.
> 
> Would this be an accurate way of describing what we'd do?
> 
> * Add a column (like `ephemeralUserId`) and leave the rest as-is.
> * Update the exposure count query to consider both ids as equal when counting uniqueness.
> * Should not cause issues with migrating in prod as there's no data migration and no change in existing usage.
> 
> Is that basically it? If so I'm starting to think that might be a cleaner way to go about this and can test it out on monday.

That makes sense as long as we don't care about the data in UpGrade being accurate with respect to how many users from which groups were exposed. The ephemeral users would have been exposed or not possibly based on the temporary group information but that data would not be recoverable if we're just storing the id as a string in an extra field. I guess we'd be committing to this as a fairly permanent solution if we're doing a db migration to accommodate it.

danoswaltCL:
That's not a requirement I've heard of though, if it becomes one then we'd want to record the user's group as it was when the exposure was captured, not refer back to the experiment-users table.

Everything group-related in the experiment-user table is still transient even for stored users and can change at any moment. They are barely non-ephemeral as it is, we're always trying to update their information on-the-fly based on current session, and we know that this information is only truly relevant for that session.

I don't think it should be used for data analysis, and AFAICT I don't believe that it is. The only spot where we read group info from this table for non-assignment / flag eligibility purposes for an individual active user is related to working-group exclusions in exclude-if-reached group consistency experiments (debatable usage itself but a different rabbit-hole...)

But in any case, should I pursue this db change? We either need to change the exposures table or create a groupless user. We would create a groupless user because the context sending groups would be explicitly stating they don't need to be stored.

danoswaltCL:
went a different direction

PR #2543, opened 2025-06-23
add parent segments list to 'Used By' status of segments

bcb37:
- Adds an array of all segments that are parent segments of any of the segments corresponding to the supplied segment IDs to the return value from the backend  for any segments 'with status' call
- Uses that data in the frontend to determine 'Used' status and 'Used By' data
- Display of parent segment info in the 'Used By' tab follows that of experiments and feature flags, including a link to the parent segment's details page; 'Status' defaults to 'Active' for all segments.

### DIFF
--- a/backend/packages/Upgrade/src/api/repositories/SegmentRepository.ts
+++ b/backend/packages/Upgrade/src/api/repositories/SegmentRepository.ts
@@ -46,8 +46,21 @@ export class SegmentRepository extends Repository<Segment> {
       .where('segment.type=:type', { type })
       .getMany()
       .catch((errorMsg: any) => {
-        const errorMsgString = repositoryError('segmentRepository', 'getAllGlobalExcludeSegment', {}, errorMsg);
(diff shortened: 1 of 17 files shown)


bcb37:
> What if segment "A" is used by segment "B", and segment "B" is used by feature flag "C"?
> 
> Should we display both segment "B" and feature flag "C" in the "Used By" table for segment "A"? Or just segment "B" in this case?
> 
> I feel like it's better to display only the direct (one-level higher) parent—just segment "B" with its actual status (instead of "Active").
> 
> I believe we originally decided to display only the top-level non-segment parents (experiments or feature flags) in the "Used By" table, but it might be more helpful to show only direct parents so users know exactly where to remove the segment.
> 
> Maybe we should bring this up in the meeting.

I'm not sure I understand the question. We are still only showing direct parents - just one level - in the 'Used By' list, as specified. Can you give a specific example of what you're describing, maybe with a screenshot?

zackcl (approved):

PR #2545, opened 2025-06-23
ephemeral ff user groups and remove exposure id FK constraint

danoswaltCL:


backend/packages/Upgrade/src/database/migrations/1703000000000-dropExperimentUserFKFromExposures.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DropExperimentUserFKFromExposures1703000000000 implements MigrationInterface {
```
bcb37, line 3: Is there a reason we want this migration to seem to have been created on December 19, 2023 (1703000000000)? That puts it in a different place in the migrations list than it would be if it had been generated by running `npm run migration:generate --name=dropExperimentUserFKFromExposures`.  In particular, it breaks running migrations after a db drop because it attempts to alter a table that doesn't exist yet -- `feature_flag_exposures` was created March 27, 2024 (1711569269846).

danoswaltCL: uhhh good catch wtf. let me try that again

backend/packages/Upgrade/src/api/middlewares/UserCheckMiddleware.ts
```diff
@@ -38,4 +46,159 @@ export class UserCheckMiddleware {
...
+   * @param user_id - The user ID for the experiment user
+   * @param groupsForSession - The groups provided in the session
+   * @returns RequestedExperimentUser object with session groups
+   */
+  private createSessionUser(user_id: string, groupsForSession: Record<string, string[]>): RequestedExperimentUser {
+    const sessionUser = new RequestedExperimentUser();
+    sessionUser.id = user_id;
+    sessionUser.requestedUserId = user_id;
+    sessionUser.group = groupsForSession;
+    sessionUser.workingGroup = undefined;
```
bcb37, line 162: Users with no working group information will now always be excluded from feature flags in contexts where there's a global exclude defined with at least one group-type list. I think I recall that we were ok with that for now, and there's still stuff to be worked out w.r.t feature flags and global excludes?

danoswaltCL: yeah should be ok, esp. considering we know how this will be used initially, but good to remember for sure.

PR #2553, opened 2025-06-26
Implement experiment root section card table component

### LINKED ISSUE #2548: Experiments Root Section Card Table + data fetch
The `content` slot of the Experiments Root Section Card will contain a Mat-Table. There is no special "Common" component for this, but there are common utility functions that in the `common-table-helpers` service that should be reused here.

We already pretty much have the exact code we need if we reference the Feature Flag Root section card and table, the difference will just be the type of data and the last column (Enrollment).

So it can all be hooked up in this story to complete the table + search widget capabilities and the global filter. 

**Requirements:**

- Table headers = ['Name', 'Status', 'Updated at', 'App Context', 'Tags', 'Enrollment']
- Should fetch experiments from `experiments/paginated`
- Should show load-spinner while loading in the table space.
- If no experiments, should display a blank state.
- With experiments present, search functionality in widget should work as expected.
- Collapse should hide the table.

**Screenshots:**

Image

Image

Figma link: https://www.figma.com/design/SN58AVPDU5DMO5WP2COt5Y/UpGrade-UI-Latest-28--Refinement-?t=zQdZ1BRm5SLNqqhV-1

zackcl:
Resolves #2548

screenshot:
Screenshot 2025-06-26 at 6 43 16 PM

### DIFF
--- a/frontend/projects/upgrade/src/app/core/experiments/store/experiments.selectors.ts
+++ b/frontend/projects/upgrade/src/app/core/experiments/store/experiments.selectors.ts
@@ -58,7 +58,7 @@ export const selectSearchExperimentParams = createSelector(
   selectSearchString,
   (searchKey, searchString) => {
     if (!!searchKey && (!!searchString || searchString === '')) {
-      return { searchKey, searchString };
+      return { searchKey: searchKey === EXPERIMENT_SEARCH_KEY.STATUS ? 'state' : searchKey, searchString };
     }
     return null;
   }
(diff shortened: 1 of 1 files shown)


bcb37 (changes requested):
Just one issue. Otherwise looks good.

amurphy-cl:
I agree that it looks long, but we if we change it to just "Complete", we will need to change it to that elsewhere too. @zackcl any thoughts?

zackcl:
@danoswaltCL @amurphy-cl I like that idea! I will apply the change.

danoswaltCL (approved):

bcb37 (approved):

PR #2559, opened 2025-06-27
Use new 'repeated_enrollment' table for within-subject experiments

bcb37:
- adds new 'repeated_enrollment' table, with migration
- moves storage of repeated enrollments in within-subject experiments from 'monitored_decision_point' and 'monitored_decision_point_log' tables to 'repeated_enrollment'
- alters all db queries for enrollment and metrics in within-subject experiments to reflect the above simplification
- removes creation of new records in 'monitored_decision_point_log' on /mark


backend/packages/Upgrade/src/api/repositories/RepeatedEnrollmentRepository.ts
```diff
@@ -0,0 +1,39 @@
+import { RepeatedEnrollment } from '../models/RepeatedEnrollment';
+import { Repository } from 'typeorm';
+import { EntityRepository } from '../../typeorm-typedi-extensions';
+import { UpgradeLogger } from 'src/lib/logger/UpgradeLogger';
```
danoswaltCL, line 4: may need to be a relative import

PR #2609, opened 2025-08-01
Resolve segment options not showing in experiment participants include table

### LINKED ISSUE #2608: Segment options not shown in Experiment Include Table
**Version where the bug was found:**  
6.1.0 and 6.2.0 (occurs on QA, Staging and Prod)

**Describe the bug**  
When creating an experiment, in the "Participants" step's "Include" table, the "ID/Name" field shows no options when the "Type" is set to "Segment", even though a segment with the same app context exists.

**To Reproduce**  
1. Go to the **Segments** page  
2. Create a segment with a specific app context (e.g., `mathstream`)  
3. Go to the **Experiments** page  
4. Click the **Add Experiment** button  
5. Select the same **App Context** as the defined segment  
6. Go to the **Participants** step  
7. Add a member in the **Include** table  
8. Select **Segment** for the **Type** field  
9. Click the **ID/Name** field  
10. Observe that no options appear

**Expected behavior**  
The **ID/Name** field should display available segments that match the selected app context when "Segment" is selected as the type.

**Screenshot**  
Image

zackcl:
## Problem
Fixes issue #2608 where segment options were not displayed in the autocomplete dropdown for the "ID/Name" field when creating experiments and selecting "Segment" as the type in the participants include table, even when segments with matching app context existed.

## Root Cause
The issue occurred due to a timing problem between when segment data was loaded and when form control autocomplete observables were initialized. Specifically:

1. Segment options were only loaded once during `ngOnInit()` 
2. When the app context changed, segment data wasn't refreshed for the new context
3. Autocomplete observables were set up before segment data was available
4. No mechanism existed to refresh autocomplete when new segment data arrived

## Solution
Implemented a comprehensive fix to ensure segment autocomplete works reliably:

## Key Changes:
1. **Context-aware segment loading**: Added `setSegmentOptions()` call in `ngOnChanges()` to reload segments when context changes
2. **Centralized segment data management**: Created `updateSegmentData()` method to handle segment array updates consistently
3. **Enhanced subscription handling**: Added proper cleanup and refresh logic in `setSegmentOptions()` 
4. **Autocomplete refresh mechanism**: Ensured `bindParticipantsData()` is called after segment data loads to refresh autocomplete observables
5. **Defensive programming**: Added null checks to prevent errors during async loading
6. **Fixed indexing bug**: Corrected array index in `addMember1()` for proper autocomplete setup

## Technical Details:
- **`ngOnChanges()`**: Now calls `setSegmentOptions()` when context changes
- **`setSegmentOptions()`**: Enhanced with subscription cleanup, data updates, and autocomplete refresh
- **`updateSegmentData()`**: New centralized method for updating segment arrays and maps
- **`bindParticipantsData()`** & **`manageSegmentIdsControl()`**: Added defensive checks for form existence
- **`addMember1()`**: Fixed index calculation for newly added members
- **`ngOnDestroy()`**: Added safe unsubscription logic

Closes #2608

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/home/components/experiment-participants/experiment-participants.component.ts
@@ -19,7 +19,7 @@ import {
   ParticipantsMember,
 } from '../../../../../core/experiments/store/experiments.model';
 import { ExperimentService } from '../../../../../core/experiments/experiments.service';
-import { Segment, MemberTypes, ListSegmentOption } from '../../../../../core/segments/store/segments.model';
+import { MemberTypes, ListSegmentOption } from '../../../../../core/segments/store/segments.model';
 import { SegmentsService } from '../../../../../core/segments/segments.service';
 import { SEGMENT_TYPE, FILTER_MODE } from 'upgrade_types';
 import { INCLUSION_CRITERIA } from 'upgrade_types';
@@ -80,13 +80,7 @@ export class ExperimentParticipantsComponent implements OnInit {
 
   ngOnChanges() {
     if (this.currentContext) {
-      this.subSegmentIds = [];
-      if (this.allSegmentListOptions) {
-        this.allSegmentListOptions.forEach((segment) => {
-          this.subSegmentIds.push(segment.name);
-          this.segmentNameId.set(segment.name, segment.id);
-        });
-      }
+      this.setSegmentOptions();
       this.setMemberTypes();
     }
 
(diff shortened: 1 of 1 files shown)


bcb37 (dismissed):
This looks good to me. Just one question:
Do we still need to call `bindParticipantsData()` on line 181 from `ngOnInit()` if it's now also being called from `setSegmentOptions()`, which is called from `ngOnInit()` on line 114?

zackcl:
@bcb37 Good catch! I've removed the unnecessary call from `ngOnInit()`. Thanks!

bcb37 (approved):

PR #2638, opened 2025-09-03
add experiment list modals and functionality

bcb37:


zackcl:
This looks good to me. I think some of these should be defined in `en.json` though.

```
nameHint: 'experiments.upsert-include-list-modal.name-hint.text',
valuesLabel: 'experiments.upsert-list-modal.values-label.text',
valuesPlaceholder: 'experiments.upsert-list-modal.values-placeholder.text',
```

zackcl (approved):

PR #2646, opened 2025-09-09
add edit-condition-weights modal

bcb37:


frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/edit-condition-weights-modal/edit-condition-weights-modal.component.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  onWeightChange(): void {
+    // Force validation update
+    this.conditionsFormArray.updateValueAndValidity();
+  }
+
+  // Helper method to get total weight status for display
+  getTotalWeightStatus(): ValidationErrors {
+    return this.conditionsFormArray.errors as ValidationErrors;
```
danoswaltCL, line 278: remove "as ValidationErrors", shouldn't be needed, gives sonarqube warning

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/edit-condition-weights-modal/edit-condition-weights-modal.component.ts
```diff
@@ -0,0 +1,300 @@
...
+    // Return 'equal' only if weights are equally distributed AND total is valid
+    return areWeightsEquallyDistributed && isTotalValid ? 'equal' : 'custom';
+  }
+
+  private setupFormValidation(): void {
+    this.isPrimaryButtonDisabled$ = combineLatest([
+      this.conditionWeightForm.statusChanges.pipe(startWith(this.conditionWeightForm.status)),
+      this.conditionWeightForm.valueChanges.pipe(startWith(this.conditionWeightForm.value)),
+    ]).pipe(
+      map(([status, value]) => {
```
danoswaltCL, line 155: can remove unused "value"

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/edit-condition-weights-modal/edit-condition-weights-modal.component.html
```diff
@@ -0,0 +1,122 @@
...
+    <ng-container matColumnDef="condition">
+      <th mat-header-cell *matHeaderCellDef class="condition-column ft-14-600">
+        Condition
+      </th>
+      <td mat-cell *matCellDef="let condition" class="condition-column ft-14-400">
+        {{ condition.conditionCode }}
+      </td>
+      <!-- Summary row for condition column -->
+      <td mat-footer-cell *matFooterCellDef class="condition-column ft-14-400 summary-row">
+        @if (getTotalWeightStatus()?.totalWeightInvalid) {
```
danoswaltCL, line 53: this would be the only place we're using the newer angular @if syntax over `ngIf`. i'm in favor of this syntax generally but just noting it's different, we haven't really talked about using the newer syntax features.

danoswaltCL: i think migrating to the new syntax is something we should commit to doing when we do the next angular update, i see no harm in that, seems better, more like react/jsx. so if you want to leave this as is, the rest of the codebase can catch up later.

PR #2654, opened 2025-09-15
move jaxb.ws.rs to Jakarta and bump Jersey to 3

ddenniston:
This is needed in order to continue work on moving CTP to Jakarta. Since this dependency is used there, this also needs updated.


danoswaltCL (approved):

danoswaltCL:
manually, we have no publish job currently, but I can go ahead and do that.

ddenniston:
> manually, we have no publish job currently, but I can go ahead and do that.

ok, thanks.

Issue #2673, opened 2025-09-24
Experiment doesn't appear on the root page after first creation


zackcl:
**Version where bug was found:**
6.2.0

**Describe the bug**
When there's no experiments defined in the root page, we see the "Welcome! Let's start by creating a new experiment!" message. If I create a first experiment in this state, the experiment doesn't appear immediately after creation. Refreshing the page makes the experiments appear but this could create confusion (this is also a problem in the demo app tour experience).

**To Reproduce**
1. Go to the Experiments root page and ensure no experiments are defined
2. Click "Add Experiment" and create an experiment
3. See that the success snack bar appears but the created experiment is not shown on the root page

**Expected behavior**
The experiment should appear on the root page after first creation.

**Screenshots**
<img width="1406" height="803" alt="Image" src="https://github.com/user-attachments/assets/c6a60950-02eb-49fe-bcb9-830ef8cb31f1" />

PR #2676, opened 2025-09-25
ensure experiment count and navigation to detail page on upsert success

danoswaltCL:
#2673

### DIFF
--- a/frontend/projects/upgrade/src/app/core/experiments/store/experiments.reducer.ts
+++ b/frontend/projects/upgrade/src/app/core/experiments/store/experiments.reducer.ts
@@ -88,6 +88,9 @@ const reducer = createReducer(
   on(experimentsAction.actionUpsertExperiment, experimentsAction.actionGetExperimentById, (state) => ({
     ...state,
     isLoadingExperiment: true,
+    // if we don't have a totalExperiments count yet (no paginated call with > 1 has occured yet),
+    // set it to 1 because we are loading at least one experiment so that nav to root page will not show empty template
+    totalExperiments: !state.totalExperiments ? 1 : state.totalExperiments,
   })),
   on(experimentsAction.actionGetExperimentByIdSuccess, (state, { experiment }) =>
     adapter.upsertOne(experiment, { ...state, isLoadingExperiment: false })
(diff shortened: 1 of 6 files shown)


frontend/projects/upgrade/src/app/core/experiments/store/experiments.reducer.ts
```diff
@@ -90,7 +94,11 @@ const reducer = createReducer(
     isLoadingExperiment: true,
   })),
   on(experimentsAction.actionGetExperimentByIdSuccess, (state, { experiment }) =>
-    adapter.upsertOne(experiment, { ...state, isLoadingExperiment: false })
+    adapter.upsertOne(experiment, {
+      ...state,
+      isLoadingExperiment: false,
+      totalExperiments: state.totalExperiments + 1,
```
bcb37, line 100: This seems to result in the count being one more than what the experiment total is. If you're not adding an experiment, just navigating to the experiment detail page, `actionGetExperimentByIdSuccess` is triggered so that the experiment whose name you just clicked on (which would have already been counted as part of the total) gets added again to the total count. Maybe this is currently not an issue because the only thing `state.totalExperiments` is used for is to determine whether to show the `noExperimentTemplate`. But I think if we ever want to depend on the accuracy of `state.totalExperiments` it might be an issue.

frontend/projects/upgrade/src/app/features/dashboard/home/root/home.component.html
```diff
@@ -2,7 +2,7 @@
   <h1 class="title ft-32-700">{{ 'global.experiment.title' | translate }}</h1>
   <span class="subtitle ft-16-400" [innerHTML]="'home.experiment.text.subtitle' | translate"></span>
 
-  <ng-container *ngIf="isLoadingExperiments$ | async; else loadingExperimentState">
```
bcb37, line 5: The problem with using this observable here is that, rather than only re-generating the 'experiment-list' component once, on initial load, it does so every time the experiment list changes. One effect this has is that we lose the ability to enter more than one character at a time in the search field, which is probably not what we want.

bcb37 (changes requested):

bcb37 (changes requested):
It looks like just adding the `navigateToExperimentDetail` effect, without any of the other changes in this pr, fixes the problem without creating any new ones.

PR #2698, opened 2025-10-17
fix types and id association for analytics query

bcb37:


bcb37:
Closing the PR to rename and change base

PR #2713, opened 2025-11-04
Display the menu button on the details page to non-Reader roles

### LINKED ISSUE #2710: The menu button on the overview section card should be visible to non-Reader roles
**Description:**
Currently, the menu button on the Segment Details page’s overview section card is only visible to users with the ‘Admin’ role. This prevents users with the ‘Creator’ role from being able to delete a segment.

To fix this, the menu button should be hidden only from users with the ‘Reader’ role and visible to all other roles.

Similarly, on the Feature Flag Details page, the menu button on the overview section card should also be visible to non-Reader roles (it is currently visible to all roles).

zackcl:
Resolves #2710

### DIFF
--- a/frontend/projects/upgrade/src/app/core/auth/auth.service.ts
+++ b/frontend/projects/upgrade/src/app/core/auth/auth.service.ts
@@ -9,6 +9,7 @@ import {
   selectCurrentUser,
   selectGoogleCredential,
   selectUserIsAdmin,
+  selectUserIsReader,
 } from './store/auth.selectors';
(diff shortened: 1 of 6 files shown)


bcb37 (commented):
Now creators can edit segments, and readers can't export segments or feature flags.
However, user managers can still export both, which is probably fine?

zackcl:
I discussed this with @amurphy-cl, and we think it’s fine for user managers to see the menu button and export items.

bcb37 (approved):

PR #2721, opened 2025-11-10
handle payload crud on backend based on conditions and decision points

### LINKED ISSUE #2717: Move dynamic payload generation logic to backend
## Background:

Currently in the stepper, the frontend does a lot of work that the backend normally does, because the stepper design made it hard to not do the work in the frontend. you don't submit anything if you change condition, decision point, or payload info until the very end, so you have to dynamically update the table on the frontend, which means actually creating and manipulating faux payload objects on the frontend before submitting... that made this a kind of a mess to code up. 👎 

in the new design, we don't have to do any of that because we can now sanely generate the payload table data on backend as soon as we add/edit/delete decision points and conditions. that means some backend changes but these should be happy changes to make, the frontend won't have to 'think' about any of this, it'll just display what's there in the experiment. ✨ 

## Requirements:

No new endpoints 👍 , just adding additional actions taken on certain `/experiment` decision point and/or condition CRUD operations to handle the business logic of dynamically updating Payload entities.

- for all, when a payload object is created, the payload === condition name by default

- when a decision point is created, a payload object should be created for each condition, if conditions exist.
- when a decision point is deleted, all payloads with that parent decision point should also be deleted.
- when a site or target are modified, the backend will not need to do anything, no changes

- when a condition is created, a payload object should be created for each decision point, if decision points exist
- when a condition is deleted, all payloads with that parent condition should also be deleted.
- when a condition name is modified, and the previous condition name === the payload exactly, then the payload should also be updated to match the new condition name
- when a condition name is modified, and the previous condition name !== the payload exactly, then do not modify payload, leave it as is

bcb37:


danoswaltCL (approved):

danoswaltCL:
@bcb37 has some test failing but otherwise works perfectly. thought this would involve a lot more code, this takes care of business really concisely. nice.

PR #2738, opened 2025-11-12
remove editing enrollment/exclusion docs on experiment edits

### LINKED ISSUE #2626: Remove chnages to exclusion and enrollment on experiment exclusion changes
For [this issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/1404) functionality was added to remove users and groups from the enrollment tables when exclusions were added to experiments. This was never fully implemented, since subsegments were not considered - either in the case where an exclusion of type "segment" is added, or in the case where such an excluded subsegment is itself edited. After reevaluation, it was decided to remove this functionality. 

Specified new behavior:
- Once a student or group is in the enrollment table, they can not be removed by editing segments or experiments
- Similarly, the exclusions table should not be affected by changes made to segments or experiments
- /mark-ing after edits to experiment exclusions have been made should work as it currently does (i.e. by adding entries to the exclusions table)
- The data displayed in the UI and included in exports will contain all enrollments. The only effect of adding experiment exclusions will be to change the behavior of  future API calls.

Note:
- If a student or group is enrolled and then subsequently excluded by an edit to an experiment or segment, and they don't /mark again, the data will show them as enrolled with whatever codition they were assigned.

bcb37:


danoswaltCL:
remind me again if there are any implications for doing this? i'm trying to remember what all the discussion was, it won't affect running experiments right? we'll just potentially have more data in the exports showing user enrollments that would have previously been hidden?

@bcb37

danoswaltCL:
(per Ben, this pr isn't urgent and can wait on experiment design refresh stuff.)

bcb37:
> Collaborator

@danoswaltCL that's correct. It's just removing the logic that added and removed enrollments and exclusions when exclusion segments are added/updated. Also, group enrollments are no longer removed from the db on assign calls. This reflects our consensus that a) We want to maintain a record of all enrollments and exclusions, even is cases where there are entries on both tables for the same user/experiment; and b) Adding inclusions after an experiment has started is "too late" for students and groups that have already been excluded to receive conditions.

bcb37:
> (per Ben, this pr isn't urgent and can wait on experiment design refresh stuff.)

To be clear though, we do want this to go into the 6.3 release.

danoswaltCL (dismissed):

danoswaltCL (approved):

PR #2753, opened 2025-11-20
add trim-input to form fields

### LINKED ISSUE #2734: Ensure 'trim-input' directive is applied to inputs.
Since we didn't make this global for some reason, we need to always remember to apply this to inputs in the new experiment design. It doesn't appear this was added to all the places it should, so this is a small story to update.

bcb37:


zackcl (changes requested):
You need to import `SharedModule` in these components for the `appTrimInput` directive to work. You can refer to `UpsertFeatureFlagModalComponent` as an example.

zackcl (approved):

PR #2758, opened 2025-11-24
NO-TICKET tech/upgrade_client_lib: Add custom exports to support default entrypoint

landaetaluis:
- Added `./dist/<platform>`-like entrypoints for backward compatibility
- Also added platform-based paths to the default entrypoint

### DIFF
--- a/clientlibs/js/package.json
+++ b/clientlibs/js/package.json
@@ -1,13 +1,39 @@
 {
   "name": "upgrade_client_lib",
-  "version": "6.2.0",
+  "version": "6.2.1",
   "description": "Client library to communicate with the Upgrade server",
   "files": [
     "dist/*"
   ],
+  "exports": {
+    ".": {
+      "browser": "./dist/browser/index.js",
+      "browser-lite": "./dist/browser-lite/index.js",
+      "node": "./dist/node/index.js",
+      "node-lite": "./dist/node-lite/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./dist/browser": {
+      "default": "./dist/browser/index.js",
+      "types": "./dist/browser/index.d.ts"
+    },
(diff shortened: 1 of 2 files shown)


zackcl (dismissed):

clientlibs/js/package.json
```diff
@@ -1,13 +1,39 @@
 {
   "name": "upgrade_client_lib",
-  "version": "6.3.0",
+  "version": "6.4.0",
```
danoswaltCL:
```suggestion
  "version": "6.3.1",
```

landaetaluis: @danoswaltCL this is introducing the default entrypoint, which is an additive change, so according to semver it should be a minor bump.

danoswaltCL: well, there is no published 6.3.x version, latest prod API and client versions are on 6.2, so actually this doesn't need a version bump at all if it's against dev (this is not a repo where every change is automatically published).

however, if you need this immediately and it's not a breaking change, then what we usually do is keep the major-minor versions consistent between client and API because they are usually linked to the API's major-minor version, for better or worse. although exceptions are sometimes warranted.

So maybe this should be pointed at the release/6.2 branch and pip the patch version, whatever that currently is. I or someone else will then need to publish it manually, they never gave us a job for doing this automatically.

landaetaluis: It seems like you (UpGrade team) have your very own process definition for it, so I'm ok with whatever makes the most sense here.

> however, if you need this immediately and it's not a breaking change...

No breaking changes indeed and I'd definitely appreciate having this available as soon as possible.

landaetaluis: @danoswaltCL done. Thanks!

danoswaltCL (approved):

PR #2772, opened 2025-12-02
delete experiment lists: gather audit log data before it is deleted

bcb37:


zackcl (dismissed):

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -2016,6 +2002,20 @@ export class ExperimentService {
...
+    listFilterMode: LIST_FILTER_MODE
+  ): Promise<ExperimentSegmentInclusion[] | ExperimentSegmentExclusion[]> {
+    const repo =
+      listFilterMode === LIST_FILTER_MODE.INCLUSION
+        ? this.experimentSegmentInclusionRepository
+        : this.experimentSegmentExclusionRepository;
+
+    const records = await repo.getExistingSegments(segmentIds);
+
+    return records;
```
zackcl, line 2016: Should we check that `getExistingSegments(segmentIds)` returns every requested ID and throw if any are missing?

bcb37: This is just gathering data for the audit logs based on what was previously selected. We find the segments from the experiment list data initially, so there would have to have been records for include/exclude lists. If we get the list records again, then we'd have to maybe verify the segments' existence again, then it's an endless loop. The only scenario I can think of is one where the query to fetch the list records fails, in which case no audit log is created. But we're not always inside a transaction, so we'd report a failure because of the audit log record not being created, but the actual db mutations would have succeeded. Not sure if that's worth re-engineering it.

danoswaltCL (approved):

PR #2788, opened 2025-12-09
duplicate experiment modal and functionality

### LINKED ISSUE #2726: Duplicate Experiment Design (Overview Section Card)
In FF we allow changing context in the duplicate view, but that seems like an issue if context-specific metrics and lists are already attached.

Apparently in FF we allow changing the context and we clear out the lists without warning. Is that what we want to do for experiments also, clear out metric queries and lists automatically without warning?

@bcb37 @amurphy-cl @zackcl 

## Screenshots

When the modal is opened, everything should be filled with the current values just like the Edit Experiment modal. And the "Name" field should be focused. The "Create" button should initially be disabled until any change is made.

Image

When an experiment is duplicated, the Decision Points, Conditions, Payloads, Include/Excludes Lists, and Metrics should remain in the duplicated experiment.

When the App Context is changed, a warning message should be displayed.

Image

bcb37:


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1392,6 +1370,63 @@ export class ExperimentService {
         logger.error(error);
         throw error;
       }
+
+      // creating condition payload docs after getting conditionDocs and decisionPointDocs
+      if (!conditionPayloads) {
+        experiment = { ...experiment, conditionPayloads: [] };
+      }
+      const conditionPayloadDocToSave: Array<any> =
```
danoswaltCL: this can be `Array<ConditionPayload>` right?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1392,6 +1370,63 @@ export class ExperimentService {
...
+        (conditionPayloads &&
+          conditionPayloads.length > 0 &&
+          conditionPayloads.map((conditionPayload) => {
+            const parentCondition = conditionPayload.parentCondition
```
danoswaltCL: ternary isn't needed, the `if` side by itself is sufficient, will be `undefined` if `find` finds nothing, and if `conditionPayload.parentCondition` is undefined, `conditionIdMap.get` will return undefined also.

```typescript
const parentCondition = conditionDocs.find((doc) => doc.id ===  conditionIdMap.get(conditionPayload.parentCondition))
```

backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1392,6 +1370,63 @@ export class ExperimentService {
...
+                  (doc) => doc.id === conditionIdMap.get(conditionPayload.parentCondition)
+                )
+              : undefined;
+            if (conditionPayload.parentCondition && !parentCondition) {
+              throw new Error(
+                `Parent condition not found for condition payload: ${conditionPayload.id}`
+              );
+            }
+            const decisionPoint = conditionPayload.decisionPoint
+              ? decisionPointDocs.find(
```
danoswaltCL: same as above

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
```diff
@@ -559,7 +564,60 @@ export class UpsertExperimentModalComponent implements OnInit, OnDestroy {
...
+      groupType,
+      tags,
+    }: ExperimentFormData,
+    sourceExperiment: Experiment
+  ): void {
+    const stratificationFactorObj = stratificationFactor ? { stratificationFactorName: stratificationFactor } : null;
+    const experimentRequest: AddExperimentRequest = {
+      // Form data
+      name,
+      description: description || undefined, // @IsOptional - can be undefined
```
danoswaltCL: do these comments with decorator looking things do something? like `@IsOptional`?

backend/packages/Upgrade/src/api/services/ExperimentService.ts
bcb37: Partial, but yeah

backend/packages/Upgrade/src/api/services/ExperimentService.ts
bcb37: ok

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
bcb37: Not sure. I copied it from the other place the AddExperimentRequest type is used. If you want I'll research it and determine whether we need it in all the places it's already being used in this file.

danoswaltCL:
The submit button is greyed out initially, even though technically we ought to be able to duplicate an experiment with all the exact same form values, as the experiment name doesn't have to be unique. The constraint on changing from initial-values in the upsert-exp modal shouldn't apply to the DUPLICATE mode. 

btw we're doing something slightly different for the name and submit behavior each entity when duplicating... 

- for flags we remove the name of flag, forcing the user to write something new in before they can submit.
- when duplicating segments, we tack on `(COPY)` to the name, and they can submit immediately.
- here we're duplicating all the values, it's the same name... which is technically valid, BUT I can't submit immediately, as if this is an EDIT form where you have to make a change in order to be allowed to submit.

so the secondary thing is if we want some consistency here. probably not terribly important but food for thought. cc: @zackcl

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
danoswaltCL: it's fine, just feels like some AI commenting cruft that's creeping in from some earlier commit that now we're copy-pasting everywhere. hell maybe I'm the one who committed it originally, i'm guilty of it. can we just remove the comments if they don't mean anything to you and they don't mean anything to me?

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
```diff
@@ -559,7 +564,60 @@ export class UpsertExperimentModalComponent implements OnInit, OnDestroy {
...
+      tags,
+      state: EXPERIMENT_STATE.INACTIVE,
+      filterMode: sourceExperiment.filterMode || FILTER_MODE.EXCLUDE_ALL,
+
+      // Backend required fields with correct defaults
+      postExperimentRule: sourceExperiment.postExperimentRule || POST_EXPERIMENT_RULE.CONTINUE,
+      enrollmentCompleteCondition: undefined, // @IsOptional - can be undefined
+      startOn: undefined, // @IsOptional
+      endOn: undefined, // @IsOptional
+      revertTo: undefined, // @IsOptional
```
danoswaltCL: these `undefined` lines can be deleted

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
danoswaltCL: this can just be `description` by itself

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
```diff
@@ -559,7 +564,60 @@ export class UpsertExperimentModalComponent implements OnInit, OnDestroy {
...
+    const experimentRequest: AddExperimentRequest = {
+      // Form data
+      name,
+      description: description || undefined, // @IsOptional - can be undefined
+      context: [appContext],
+      type: experimentType,
+      assignmentUnit: unitOfAssignment,
+      consistencyRule: unitOfAssignment !== ASSIGNMENT_UNIT.WITHIN_SUBJECTS ? consistencyRule : undefined, // Conditional validation
+      conditionOrder: unitOfAssignment === ASSIGNMENT_UNIT.WITHIN_SUBJECTS ? conditionOrder : undefined, // Conditional validation
+      assignmentAlgorithm: assignmentAlgorithm || undefined, // @IsOptional
```
danoswaltCL: just `assignmentAlgorithm` will do

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/upsert-experiment-modal.component.ts
```diff
@@ -559,7 +564,60 @@ export class UpsertExperimentModalComponent implements OnInit, OnDestroy {
...
+      endOn: undefined, // @IsOptional
+      revertTo: undefined, // @IsOptional
+      conditions: sourceExperiment.conditions, // @IsNotEmpty @IsArray - must be empty array, not undefined
+      partitions: sourceExperiment.partitions, // @IsNotEmpty @IsArray - must be empty array, not undefined
+      factors: sourceExperiment.factors, // @IsOptional @IsArray - can be undefined
+      conditionPayloads: sourceExperiment.conditionPayloads, // @IsOptional @IsArray - can be undefined
+      queries: this.isContextChanged ? undefined : sourceExperiment.queries, // @IsOptional @IsArray - can be undefined
+      experimentSegmentInclusion: this.isContextChanged ? undefined : sourceExperiment.experimentSegmentInclusion, // @IsOptional @IsArray - can be undefined
+      experimentSegmentExclusion: this.isContextChanged ? undefined : sourceExperiment.experimentSegmentExclusion, // @IsOptional @IsArray - can be undefined
+      stateTimeLogs: undefined, // @IsOptional @IsArray - can be undefined
```
danoswaltCL: can be removed

backend/packages/Upgrade/test/integration/Experiment/factorial/FactorialEnrollment.ts
```diff
@@ -13,18 +13,13 @@ export default async function FactorialExperimentEnrollment(): Promise<void> {
 
   const experimentService = Container.get<ExperimentService>(ExperimentService);
   // experiment object
-  const experimentObject = firstFactorialExperiment;
+  const experimentObject = JSON.parse(JSON.stringify(firstFactorialExperiment));
```
danoswaltCL: i think `structuredClone` is the native way to do this trick nowadays.

bcb37: We've got 57 examples of JSON.parse(JSON.stringify(...)) and none yet of `structuredClone`. I'll see if changing them all doesn't break things. If it does I'll make a ticket to modernize this everywhere.

zackcl:
Yeah, I think we should make the duplicate modal behavior consistent across feature flags, segments, and experiments. The segment approach of auto-tacking on “(COPY)” feels the most intuitive since it allows immediate submit while still letting users rename if they want.

> The submit button is greyed out initially, even though technically we ought to be able to duplicate an experiment with all the exact same form values, as the experiment name doesn't have to be unique. The constraint on changing from initial-values in the upsert-exp modal shouldn't apply to the DUPLICATE mode.
> 
> btw we're doing something slightly different for the name and submit behavior each entity when duplicating...
> 
> * for flags we remove the name of flag, forcing the user to write something new in before they can submit.
> * when duplicating segments, we tack on `(COPY)` to the name, and they can submit immediately.
> * here we're duplicating all the values, it's the same name... which is technically valid, BUT I can't submit immediately, as if this is an EDIT form where you have to make a change in order to be allowed to submit.
> 
> so the secondary thing is if we want some consistency here. probably not terribly important but food for thought. cc: @zackcl

zackcl:
@bcb37 It seems like the duplicated experiment’s include/exclude list description is being filled with a UUID when the original list’s description is empty. The duplicated experiment’s include/exclude list description should remain empty if the original list’s description is empty.

Original experiment's exclude list:
Screenshot 2025-12-18 at 10 54 24 AM

Duplicated experiment's exclude list:
Screenshot 2025-12-18 at 10 55 57 AM

danoswaltCL (approved):

PR #2793, opened 2025-12-11
Feature/mooclet params modal and view

danoswaltCL:
- Adds a ts-configurable form in add/edit experiment modal
- Adds a bulleted list capability to the overview details component
- Implements bulleted list for moocletPolicyParameters when detected
- Mooclet Helper Service to delegate mooclet related logic as much as possible
- Adds Reward Feedback card under Enrollment Data in Data Tab, which is essentially the current_posteriors data

@amurphy-cl open to any wordsmithing of the language

image
image


frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/ts-configurable-policy-parameters-form/ts-configurable-policy-parameters-form.component.html
```diff
@@ -0,0 +1,102 @@
+<div class="ts-configurable-params-section">
+  <span class="section-label">
+    {{ 'home.new-experiment.design.ts-configurable-policy.section-label.text' | translate }}
+    <a
+      class="learn-more-link"
+      href="https://upgrade-platform.gitbook.io/upgrade-documentation/creating-an-adaptive-experiment/thompson-sampling-configurable"
```
zackcl, line 6: I’d suggest reducing the font size of "Learn more" by adding the `ft-12-400` class for better consistency and visual distinction.

Screenshot 2025-12-12 at 4 30 55 PM

frontend/projects/upgrade/src/assets/i18n/en.json
```diff
@@ -183,6 +183,27 @@
   "home.new-experiment.design.condition.column-header.weight.text": "Weight(%)",
   "home.new-experiment.design.condition.column-header.include.text": "INCLUDE",
   "home.new-experiment.design.mooclet-policy-parameters.header.text": "Mooclet Policy Parameters",
+  "home.new-experiment.design.ts-configurable-policy.section-label.text": "Thompson Sampling (Configurable) Policy Parameters",
+  "home.new-experiment.design.ts-configurable-policy.prior-success.label.text": "Prior: Success Count",
+  "home.new-experiment.design.ts-configurable-policy.prior-success.hint.text": "Prior success count for Beta distribution (default: 1)",
+  "home.new-experiment.design.ts-configurable-policy.prior-success.required-error.text": "Prior success is required",
+  "home.new-experiment.design.ts-configurable-policy.prior-success.min-error.text": "Prior success must be at least 1",
+  "home.new-experiment.design.ts-configurable-policy.prior-failure.label.text": "Prior: Failure Count",
+  "home.new-experiment.design.ts-configurable-policy.prior-failure.hint.text": "Prior failure count for Beta distribution (default: 1)",
```
zackcl, line 191: Should we rename "Prior: Success Count" and "Prior: Failure Count" to "Prior Success" and "Prior Failure" to match how they’re displayed on the details page?

Screenshot 2025-12-12 at 4 40 23 PM

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/ts-configurable-policy-parameters-form/ts-configurable-policy-parameters-form.component.html
```diff
@@ -0,0 +1,102 @@
...
+        max="1.0"
+        step="0.1"
+      />
+      <mat-hint class="form-hint ft-12-400">{{
+        'home.new-experiment.design.ts-configurable-policy.tspostdiff-thresh.hint.text' | translate
+      }}</mat-hint>
+      <mat-error *ngIf="policyForm.get('tspostdiff_thresh')?.hasError('required')"
+        >`
+        {{ 'home.new-experiment.design.ts-configurable-policy.tspostdiff-thresh.required-error.text' | translate }}
+      </mat-error>
```
zackcl: Remove ` after >

Screenshot 2025-12-12 at 4 46 12 PM

zackcl:
It might be worth setting a max limit for all numeric input fields for safety and displaying an error message when the value exceeds the limit (for example, the “TS Post Diff Threshold” currently doesn’t show an error).

Screenshot 2025-12-12 at 4 49 12 PM

zackcl:
When I create a "TS Configurable" experiment, switch the Assignment Algorithm to "Random", and then switch it back to "TS Configurable", all policy parameters become empty. I think they should be populated with the default values in this case.

Screenshot 2025-12-12 at 4 57 36 PM

frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-experiment-modal/ts-configurable-policy-parameters-form/ts-configurable-policy-parameters-form.component.html
zackcl: Consider removing these “X is required” error messages. We don’t seem to display this type of message elsewhere, and the red field state alone already communicates that the input is required.

danoswaltCL:
> When I create a "TS Configurable" experiment, switch the Assignment Algorithm to "Random", and then switch it back to "TS Configurable", all policy parameters become empty. I think they should be populated with the default values in this case.
> 
> Screenshot 2025-12-12 at 4 57 36 PM

hm yeah that was the intent i'll check that

zackcl:
It seems like the frontend test is failing.

danoswaltCL:
@zackcl tests are passing now. there's a backend bug but the UI stuff should be ok.

zackcl:
It looks like updating an experiment back to "TS Configurable" after switching it to "Random" causes a 500 error on `PUT /api/experiments/:id`.

Steps to reproduce:
1. Create an experiment with "TS Configurable".
2. Change the Assignment Algorithm to "Random".
3. Change it back to "TS Configurable".

zackcl:
@danoswaltCL It looks like this might be fixed now after pulling the latest commit. Can you confirm?

zackcl:
This might not be worth addressing right now, but I noticed an issue with the max limit validation.

The max limit error messages for the first three fields don’t appear until the fourth field reaches the max limit, whereas the min limit validation shows up immediately as expected. It seems like the fourth field (Uniform Threshold) is triggering the max limit validation for the previous three fields.

This isn’t critical, but I wanted to share it in case you’d like to address it.

Screenshot 2025-12-16 at 6 14 43 PM

Screenshot 2025-12-16 at 6 15 30 PM

Screenshot 2025-12-16 at 6 17 59 PM

Screenshot 2025-12-16 at 6 20 59 PM

danoswaltCL:
yes @zackcl the algorithm-switching errors are hopefully fixed, and i pushed up fix for the max validation clearing issue.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -1375,6 +1436,120 @@ export class MoocletExperimentService extends ExperimentService {
...
+          moocletRefId: moocletRefToDelete?.id,
+          experimentId: experiment.id,
+        });
+      }
+
+      // 2. Update experiment with new assignment algorithm, creating new mooclet resources when "isNowMooclet"
+      let message = `[Algorithm Change] Assignment algorithm changed from ${oldAlgorithm} to ${experiment.assignmentAlgorithm}.`;
+      message += wasMooclet && 'Will delete old Mooclet resources after successful update.';
+      message += isNowMooclet && 'Will create new Mooclet resources as part of update.';
+
```
zackcl, line 1484: Problem: `&&` operator doesn't work as expected for string concatenation. If `wasMooclet` is true, it evaluates to the string, but if false, it adds `false` to the message. Fix:

```
let message = `[Algorithm Change] Assignment algorithm changed from ${oldAlgorithm} to ${experiment.assignmentAlgorithm}.`;
if (wasMooclet) message += ' Will delete old Mooclet resources after successful update.';
if (isNowMooclet) message += ' Will create new Mooclet resources as part of update.';
```

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -164,13 +165,45 @@ export class MoocletExperimentService extends ExperimentService {
...
+  }
+
+  // public async syncUpdateTransitioningAwayFromMooclet(params: SyncEditParams): Promise<ExperimentDTO> {
+  //   return this.dataSource.transaction(async (manager) => {
+  //     if (params.moocletRefToDelete) {
+  //       await this.moocletExperimentRefRepository.delete(params.moocletRefToDelete.id);
+  //     }
+  //     return await this.updateUpgradeExperiment(manager, params);
+  //   });
+  // }
```
zackcl: Consider removing this commented code.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -164,13 +165,45 @@ export class MoocletExperimentService extends ExperimentService {
...
   }
 
+  public async syncUpdateWithMoocletAlgorithmTransition(params: SyncEditParams): Promise<ExperimentDTO> {
+    return this.dataSource.transaction(async (manager) => {
+      const updatedExperiment = await this.updateUpgradeExperiment(manager, params);
+
+      // when transitioning away from Mooclet, delete the old refs after successful experiment update
+      if (params.moocletRefToDelete) {
+        await this.moocletExperimentRefRepository.delete(params.moocletRefToDelete.id);
+      }
```
zackcl, line 186: Problem: The deletion uses `moocletExperimentRefRepository.delete()` directly instead of using the transaction manager. This means the deletion may not be part of the transaction and could cause:

- Orphaned refs if the transaction rolls back after deletion
- Loss of atomicity

Fix:
```
if (params.moocletRefToDelete) {
  await manager.delete(MoocletExperimentRef, params.moocletRefToDelete.id);
}
```

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
```diff
@@ -164,13 +165,45 @@ export class MoocletExperimentService extends ExperimentService {
...
+
+      // when transitioning to Mooclet, create the Mooclet resources
+      if (this.isMoocletExperiment(updatedExperiment.assignmentAlgorithm)) {
```
zackcl, line 189: Double Deletion Risk in Algorithm Change Flow

Location: `MoocletExperimentService.ts:1507-1535` The code deletes mooclet resources in two places:

1. Inside `syncUpdateWithMoocletAlgorithmTransition` transaction (line 185) - deletes UpGrade DB ref
2. Outside transaction in `handlePotentialMoocletAssignmentAlgorithmChange` (line 1515) - deletes external Mooclet API resources

Issue: If the external Mooclet deletion fails (line 1515), you've already deleted the UpGrade DB reference (line 185), making it impossible to retry cleanup later. This creates orphaned Mooclet API resources with no way to track them. Recommendation: Consider one of these approaches:
- Option A: Only delete the UpGrade ref inside the transaction, move ALL external deletions outside
- Option B: Keep a "soft delete" flag on refs to track failed cleanups
- Option C: Add a cleanup job to handle orphaned resources

zackcl:
I added a few review comments suggested by Claude. Feel free to resolve or ignore them.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
danoswaltCL: This one is okay, it isn't seeing that this function is only invoked in one place and should only ever be called in that place, and that function scope is wrapped in the transaction.

backend/packages/Upgrade/src/api/services/MoocletExperimentService.ts
danoswaltCL: orphaned resources aren't great but it won't hurt anything. we've accepted this scenario in other spots. I don't think i could stomach adding more complexity to handle all the cleanup variations. I think that's the big takeaway I have from this project in general, we underestimated how much extra work it is to keep two systems in sync, the tree of potential failure scenarios to guard against turns every simple crud operation into hundreds of extra lines of code

danoswaltCL:
@zackcl i resolved a couple of the claude suggestions and commented on the other two.

zackcl (approved):

zackcl:
@danoswaltCL I'm getting the "moocletPolicyParameters should not be null or undefined" error when trying to create an experiment with TS Configurable algorithm. I remember this used to work fine last time. Am I missing something? Or could it be related to the recent commit ([ed22847](https://github.com/CarnegieLearningWeb/UpGrade/pull/2793/commits/ed22847743c87ce1e93d71df706bca9ca61d5fe9)) you made?

danoswaltCL:
hm yes that looks like I missed something from the duplicate-mode merge yesterday, i'm on it

danoswaltCL:
@zackcl can you verify this is working again now? it was just a couple lines missed after a merge.

zackcl:
Yes, everything seems to be working fine now.

zackcl (approved):

PR #2801, opened 2025-12-11
Remove % from the weight values and restore focus after updating weights

### LINKED ISSUE #2796: Remove % from the weight values from Conditions table
Currently, the Conditions table looks like this:

Image

But it should be updated as the following (by removing % from weight values):

Image

Since the column header already indicates “Weight (%)”, repeating the percent symbol inside every row is redundant and adds unnecessary visual noise. Removing it makes the table cleaner and more consistent with the original design requirement in #2678.

### LINKED ISSUE #2800: Restore focus after clicking "Save" on the "Edit Condition Weights" modal
Clicking the "Save" button on the "Edit Condition Weights" modal currently highlights the edit icon/button on the Conditions table.

The edit icon/button should not be highlighted after the weights have been updated through the modal.

**Screenshot**
Image

zackcl:
Resolves #2796, resolves #2800

### DIFF
--- a/frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
+++ b/frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
@@ -528,6 +528,7 @@ export class DialogService {
       panelClass: ['experiment-modal', 'modal-shadow'],
       hasBackdrop: true,
       autoFocus: false,
+      restoreFocus: false,
       backdropClass: 'modal-backdrop',
       width: ModalSize.STANDARD,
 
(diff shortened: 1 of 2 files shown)


frontend/projects/upgrade/src/app/shared/services/common-dialog.service.ts
```diff
@@ -528,6 +528,7 @@ export class DialogService {
       panelClass: ['experiment-modal', 'modal-shadow'],
       hasBackdrop: true,
       autoFocus: false,
+      restoreFocus: false,
```
danoswaltCL, line 531: neat

danoswaltCL (approved):

PR #2802, opened 2025-12-12
Disallow defining a duplicate decision point in Decision Points table

### LINKED ISSUE #2797: Disallow defining a duplicate decision point in Decision Points table
Currently, it is allowed to add a decision point that already exists in Decision Points table. We should not allow this in the Add/Edit Decision Points modal similar to how we disallow duplicate condition name in Add/Edit Condition modal.

We should probably display the validation error message below the Site and Target fields like this (when invalid):

Image

zackcl:
Resolves #2797


danoswaltCL (approved):
looks good to me

PR #2813, opened 2026-01-06
add import experiment modal and functionality

### LINKED ISSUE #2723: Import Experiment Modal
This should operate the same as it always has.

## Screenshots:

Image

Image

Image

Image

bcb37:


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -98,7 +98,7 @@ import { ExperimentAuditLog } from '../models/ExperimentAuditLog';
 import { SegmentRepository } from '../repositories/SegmentRepository';
 import { NotFoundException } from '@nestjs/common/exceptions';
 
-const errorRemovePart = 'instance of ExperimentDTO has failed the validation:\n - ';
+const errorRemovePart = 'An instance of ExperimentDTO has failed the validation:\n - ';
```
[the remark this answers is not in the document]
bcb37: This has been verified. Leaving "An" in the message resulted in awkward messages like "An property state has failed the following constraints:...".

danoswaltCL:
maybe this is a pre-existing bug unrelated to this pr, but when importing the 'updated-at' date is still the old updated-at date:

image

bcb37:
> maybe this is a pre-existing bug unrelated to this pr, but when importing the 'updated-at' date is still the old updated-at date:
> 
> image

@danoswaltCL  It is a pre-existing behavior. Should we revisit the spec for this issue?

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-modal.component.ts
```diff
@@ -211,7 +211,7 @@ export class CommonImportModalComponent implements OnInit, OnDestroy {
 
   showNotification(importResult: ValidatedImportResponse[]) {
     const importSuccessFiles = importResult
-      .filter((data) => data.error == null || data.error.startsWith('warning'))
+      .filter((data) => data.error == null || data.error.toLowerCase().startsWith('warning'))
```
danoswaltCL: maybe a preexisting bug (and I'm allergic to `==` in javascript so I'm itching for a reason to change it), but I don't think the logic is quite right? 

could this be changed to `.filter((data) => !data.error || (typeof data.error === 'string' && data.error.toLowerCase().startsWith('warning')))`?

the previous and current logic would compile as "return TRUE if data.error is exactly `null` or `undefined`, otherwise FALSE." so it would be erroring on `.toLowerCase()` or previously `.startsWith()` if `data.error` is unexpectedly not a string.

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-modal.component.ts
```diff
@@ -224,7 +224,9 @@ export class CommonImportModalComponent implements OnInit, OnDestroy {
 
     this.notificationService.showSuccess(importSuccessMsg);
 
-    const importFailedFiles = importResult.filter((data) => data.error != null && !data.error.startsWith('warning'));
+    const importFailedFiles = importResult.filter(
```
danoswaltCL, line 229: same as above

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-type-adapters.ts
```diff
@@ -45,6 +45,31 @@ export class FeatureFlagImportAdapter implements ImportServiceAdapter {
   }
 }
 
+@Injectable({ providedIn: 'root' })
+export class ExperimentImportAdapter implements ImportServiceAdapter {
+  constructor(private experimentDataService: ExperimentDataService, private experimentService: ExperimentService) {}
+
+  validateFiles(files: any[]): Observable<ValidatedImportResponse[]> {
```
danoswaltCL: we do have some types defined for files, FeatureFlagFile and SegmentFile, this could be typed the same way probably instead of `any`.

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-modal.component.ts
bcb37: good catch

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-modal.component.html
```diff
@@ -73,12 +73,12 @@
                 <div class="ft-14-500" *ngIf="expandedRows.has(rowIndex)" class="compatibility-details-container">
                   <div *ngIf="element.compatibilityType === 'incompatible'">
                     <span *ngIf="config.params.incompatibleMessageKey">
-                      {{ config.params.incompatibleMessageKey | translate }}
+                      {{ element.error || (config.params.incompatibleMessageKey | translate) }}
```
danoswaltCL, line 76: i'm happy we're putting this in here, I couldn't remember what we decided on printing the real error but I think it's good to do even if the error is technical gibberish. @zackcl Do we have design specs for longer text error message like this? (I think this example message where I imported some random .json file looks decent as-is).

Image

zackcl: Yeah, I think this looks fine for now. I don’t think we ever defined strict design specs for long error text.
Only thing I noticed is the arrow icon should be down when expanded.

danoswaltCL:
> > maybe this is a pre-existing bug unrelated to this pr, but when importing the 'updated-at' date is still the old updated-at date:
> > image
> 
> @danoswaltCL It is a pre-existing behavior. Should we revisit the spec for this issue?

i created a separate bug issue

frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-import-modal/common-import-modal.component.html
danoswaltCL: ok, that's also a pre-existing issue (it happens in prod, somehow we've never notices), i'll make a bug ticket and we can move on, I think everything else in this PR was addressed so I'm approving.

bcb37: @zackcl the arrow icon doesn't work for segments or feature flag imports currently. I can fix it in this pr.

danoswaltCL (approved):

danoswaltCL (approved):

PR #2816, opened 2026-01-06
add metricKey to query upsert values

### LINKED ISSUE #2760: Metric key doesn't update from continuous to categorical type
**Version where the bug was found**
v6.3.0

**Describe the bug**
The metric does not properly update when changing its data type from "continuous" to "categorical."

**To Reproduce**
Steps to reproduce:
1. Go to https://apps.qa-cli.net/upgrade
2. Click the "Add Experiment" button.
3. Create an experiment using the "upgrade-internal" App Context.
4. In the "Metrics" step, add "exampleContinuousMetricKey" (use any "Statistic" and "Display Name").
5. After creating the experiment, go to the "Metrics" tab and confirm that "exampleContinuousMetricKey" has been added.
6. Click the "Edit" button and go to the "Metrics" step.
7. Update the metric to "exampleCategoricalMetricKey" (use any "Statistic", "Comparison", "Value", and "Display Name") and click "Save."
8. In the "Metrics" tab, you will see that the metric has been updated to "exampleCategoricalMetricKey" as expected.
9. Refresh the details page and return to the "Metrics" tab.
10. Observe that the metric has reverted back to "exampleContinuousMetricKey."

**Expected behavior**
The metric (`metricKey` column in the DB) should correctly update between "continuous" and "categorical" types, and persist after a page refresh.

bcb37:


danoswaltCL (approved):
looks good to me

PR #2823, opened 2026-01-08
Feature/2786 user managers can update segments

### LINKED ISSUE #2786: Allow User Manager role to modify segments and inclusion/exclusions
We need an UpGrade user role that aligns to a "customer service" type of use case. Specifically, one who can view experiments/flags, update segments, and include/exclude lists but can't start or stop experiments. This is of primary importance for the near short-term as we need to manually manage user exclusions in LHAI until they're able to be handled programatically. 

Rather than create another role, my suggestions is to simply add some functionality to the existing User Manager role. 

Not sure if it's possible to get this in for 6.3, but if not, maybe January can be a target?

bcb37:


zackcl:
I noticed that the `not-allowed` cursor icon doesn't appear when hovering over the disabled "Edit Weights" button. There was a mistake (scoping issue) in .scss so please replace the current `.weight-edit-column { ... }` with this:
```scss
    .weight-edit-column {
      width: 15%;
      text-align: left;

      .weight-edit-button {
        color: var(--dark-grey);

        &[disabled] {
          .mat-icon {
            opacity: 0.5;
          }
        }
      }
      &:has(.weight-edit-button[disabled]) {
        cursor: not-allowed;
      }
    }
```

Here's the diff for your referece:
```diff
             opacity: 0.5;
           }
         }
-
-        &:has(.weight-edit-button:disabled) {
-          cursor: not-allowed;
-        }
+      }
+      &:has(.weight-edit-button[disabled]) {
+        cursor: not-allowed;
       }
     }

```

Other than this, everything looks good and works as expected.

zackcl (approved):

PR #2853, opened 2026-01-15
Feature/separate libraries js client

danoswaltCL:
Changes:
- Merged `lite-browser` and lite-node into single `upgrade_client_lib_lite` package that works universally in both browser and Node.js environments

- Removed Client-source header and IS_BROWSER build flag (this is only code change, this was not used and was the only difference anymore between `browser-lite` and `node-lite` packages, so was pointless to make two packages).

Updated exports to maintain existing export paths for backwards compatibility. Browser and Node `lite` will point to same bundle. These bundles can now be published as their own libraries:

- upgrade_client_lib_full_browser (~489 KB, axios included)
- upgrade_client_lib_full_node (~684 KB, axios included)
- upgrade_client_lib_lite (~450 KB, provide custom httpclient)

The legacy `upgrade_client_lib` package will continue to work with all existing imports to keep backwards compatibility, there just won't be any difference between the `lite` node and browser bundles. All will be published as same version.

cc: @landaetaluis (I still don't see your name when i'm trying to pick reviewers)


lewisjmiller (dismissed):
From what I understand of this, it makes sense to me! Thank you!

clientlibs/js/packages/lite/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "description": "UpGrade client library with externalized axios - works in both browser and Node.js",
+  "main": "dist/lite/index.js",
+  "types": "dist/lite/index.d.ts",
+  "files": [
+    "dist/lite/**/*"
+  ],
+  "dependencies": {
+    "uuid": "^8.3.2"
+  },
+  "peerDependencies": {
```
kschaefe: Shouldn't this be an optional peer dependency?  The point of this code is to allow us _not_ to use `axios`, which we do not want to have on the browser.

clientlibs/js/packages/lite/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "name": "upgrade_client_lib_lite",
+  "version": "6.2.6",
+  "description": "UpGrade client library with externalized axios - works in both browser and Node.js",
+  "main": "dist/lite/index.js",
+  "types": "dist/lite/index.d.ts",
+  "files": [
+    "dist/lite/**/*"
+  ],
+  "dependencies": {
+    "uuid": "^8.3.2"
```
kschaefe, line 11: What are we using this for?  UUID generation?  Use `crypto.generateUUID()` instead.

This can be fixed in the next release, please.

kschaefe (changes requested):

clientlibs/js/packages/lite/package.json
danoswaltCL: yes sure.

clientlibs/js/packages/lite/package.json
danoswaltCL: @kschaefe was able to validate that axios shouldn't even need to be an optional dependency either in the lite build, made a small code change to ensure it never needs to be involved in these builds.

kschaefe (approved):
Seems fine.

clientlibs/js/packages/lite/package.json
kschaefe: Perfect.

lewisjmiller:
@danoswaltCL Is this good to go to be merged?

danoswaltCL:
@lewisjmiller I will soon and publish the proper lite library version, I wanted other team members to take a look but I think that we're good.

bcb37 (approved):

danoswaltCL:
@lewisjmiller ok, `upgrade_client_lib_lite` 6.2.6 is now available, should work exactly the same as importing the `lite` export from `upgrade_client_lib`. let me know!

PR #2865, opened 2026-01-20
implement no-inclusions warning for experiments

danoswaltCL:


bcb37 (commented):
This looks good. One thing that I think is from the last pr: The tooltip doesn't render '\n' so multiple warnings will all be on a single line separated by •.

danoswaltCL:
I updated this to borrow the multiline styling used in the start-button tooltip, took me a minute to understand what @zackcl 's magic was here to get the tooltip to render nicely. there's a case to be made for a common tooltip so we have same handling of tooltip stuff across the app, but i think a shared css class is good enough for now.

zackcl:
The frontend tests are failing.

danoswaltCL:
yeah bad merge, i should never use the github GUI to try and resolve merge conflicts, it should be fixed now.

frontend/projects/upgrade/src/assets/i18n/en.json
```diff
@@ -45,6 +45,7 @@
   "global.no-monitored-metrics-tabs.text": "No Monitored Metrics",
   "global.group-type-singular.text": "Group ({{groupType}})",
   "global.group-type-plural.text": "Groups ({{groupType}})",
+  "global.no-participants-included-warning.text": "No participants are currently included.",
   "app-header.title.login": "Login",
```
zackcl, line 49: Minor nit: can we drop the period here for consistency with other tooltip/status messages?

zackcl (approved):

PR #2874, opened 2026-01-21
fix order of payloads returned with experiments

### LINKED ISSUE #2845: Payloads table row order changes unexpectedly after experiment updates
The order of rows in the Payloads table changes randomly immediately after updating the experiment.

bcb37:


backend/packages/Upgrade/src/api/services/ExperimentService.ts
```diff
@@ -1065,6 +1065,14 @@ export class ExperimentService {
...
         });
 
+        // sort the payloads by decision point order and condition order
+        conditionPayloadDocToReturn.sort((a, b) => {
+          if (a.decisionPoint.order === b.decisionPoint.order) {
+            return a.parentCondition.order - b.parentCondition.order;
+          }
+          return a.decisionPoint.order - b.decisionPoint.order;
+        });
+
```
danoswaltCL, line 1075: both conditions are returning same thing, `a.decisionPoint.order - b.decisionPoint.order`

```suggestion
        // sort the payloads by decision point order and condition order
        conditionPayloadDocToReturn.sort((a, b) => {
          return a.decisionPoint.order - b.decisionPoint.order;
        });

```

bcb37: @danoswaltCL  I'm not sure I understand. If `decisionPoint.order` is the same for both, then we return sorted by `parentCondition.order`, otherwise we return sorted by `decisionPoint.order`.
Am I missing something?

danoswaltCL: no, i can't read good

danoswaltCL (approved):

PR #2876, opened 2026-01-21
Adjust vertical spacing between Enrollment and Metrics Data section cards

### LINKED ISSUE #2863: Adjust vertical spacing between Enrollment and Metrics Data section cards
The vertical spacing between Enrollment Data and Metrics Data section card is currently 64px. This is caused by the hidden Reward Feedback section card. This should be fixed so that the spacing is 32px which is consistent with other section cards spacing.

zackcl:
Resolves #2863

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-details-page-content.component.html
+++ b/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-details-page-content.component.html
@@ -53,7 +53,7 @@
       ></app-experiment-enrollment-data-section-card>
 
       <app-experiment-reward-feedback-section-card
-        *ngIf="activeTabIndex === 1"
+        *ngIf="activeTabIndex === 1 && (shouldShowRewardFeedback$ | async)"
         [isSectionCardExpanded]="isSectionCardExpanded"
         section-card
       ></app-experiment-reward-feedback-section-card>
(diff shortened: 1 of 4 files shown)


frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-details-page-content.component.html
```diff
@@ -53,7 +53,7 @@
       ></app-experiment-enrollment-data-section-card>
 
       <app-experiment-reward-feedback-section-card
-        *ngIf="activeTabIndex === 1"
+        *ngIf="activeTabIndex === 1 && (shouldShowRewardFeedback$ | async)"
```
danoswaltCL, line 56: what's the reason for moving the display control logic for the reward card?

zackcl: The `app-common-section-card-list` applies `row-gap: 32px` between all flex children. When the reward feedback card was conditionally hidden via an internal `*ngIf`, its empty host element remained in the DOM, causing two gaps (64px total) instead of one.

Moving the visibility logic to the parent's `*ngIf` ensures the host element is completely removed from the DOM when the card shouldn't be displayed, maintaining consistent 32px spacing between section cards.

danoswaltCL: aha, ok

danoswaltCL (approved):

Issue #2175, opened 2025-01-23
Bug: Delete sync on error should report mooclet deletion failure


danoswaltCL:
Per @bcb37's finding in QA: Deleting an experiment deletes all of its associated mooclet resources, if there are no problems communicating with the mooclet server.

If the deletion of the mooclet resources fails, the experiment is still deleted. However, there are still log messages generated that say 'Upgrade and Mooclet experiment resources deleted successfully'.

Expectation: The experiment should not be deleted if mooclet resources fail to delete, and the logs should say so.

bcb37:
QA: delete experiment fails when mooclet resources fail to be deleted; this is now reflected in the logs.

Issue #1930, opened 2024-09-12
mooclet backend: delete experiment sync


danoswaltCL:

bcb37:
QA: Deleting an experiment deletes all of its associated mooclet resources, if there are no problems communicating with the mooclet server.
If the deletion of the mooclet resources fails, the experiment is still deleted. However, there are still log messages generated that say 'Upgrade and Mooclet experiment resources deleted successfully'.

PR #2881, opened 2026-01-23
Fix/mooclet error messaging

danoswaltCL:
#2175 (spawned from #1930)

- Fixes the delete issue where it accidentally succeeds on deleting UpGrade experiment when the mooclet delete action fails, and logs wrong thing
- Standardizes Mooclet Experiment synchronization error types
- Adds a more complete set of unit tests for the Mooclet Experiment service.

Note! Also makes 2 necessary changes in the global ErrorHandlerMiddleware:

- `next(error)` was being called after res.json(error), this was causing an issue in some specific cases, and I came to find out that this is an incorrect thing to do if the middleware is the end of the line for a request, which the errorhandler is, the route handler will resolve and send the response, so next() is not right to use and just causes weird side effects sometimes.
- In cleaning this up and trying to figure out why certain error scenarios would send unexpected error responses, I also discovered that the wrong error object is being sent in this middleware! This file goes through the work of creating a specially formatted `experimentError` from the original error that it has received, but instead sends the original error in the response. That looks fine in a lot of cases but it's clear our globally handled errors are supposed to look different, and the difference is much nicer, much fewer ambiguous errors.


bcb37 (commented):
This results in some broken-seeming UI behavior when communication with the mooclet server fails: if you try to turn on the IncludeAll switch, for instance, the progress indicators for Include Lists, Exclude Lists and Metrics go infinite (though that could be something that would happen any time hitting that switch results in a 500)

One other odd thing: removing the `next()` call from the error middleware seems to result in the errors not being written to the command-line in the console, which I've found useful sometimes.

bcb37 (approved):

PR #2883, opened 2026-01-26
Optimize experiment pagination query by removing unused JOINs

### LINKED ISSUE #2859: Optimize experiment pagination query performance
Currently, the experiment pagination query is taking significantly longer than expected.

In my local UpGrade environment (using the `experiment-design-refresh branch`), I have about 20 experiments defined. I noticed that loading the experiment root page takes noticeably long, so I checked the browser’s developer console and found that the `/api/experiments/paginated` request takes over 3 seconds to complete.

The backend terminal also logs the following:
```
query is slow: SELECT "experiment"."createdAt" AS "experiment_createdAt", "experiment"."updatedAt" AS "experiment_updatedAt", "experiment"."versionNumber" AS "experiment_versionNumber", "experiment"."id" AS 
...(abbreviated)
execution time: 2815
```

For comparison, in the QA UpGrade environment (which uses the legacy experiment implementation), loading the experiment root page takes about 334ms, which is roughly 10x faster than in the refreshed app.

zackcl:
Resolves #2859

This PR removes unused JOINs in the `/api/experiments/paginated` endpoint to improve query performance. I also reverted the indexing changes as suggested. In my testing, this made the endpoint ~10x faster.

Since I’m less familiar with this area of the codebase, I’d appreciate a careful review to confirm the changes are safe and correct.

### DIFF
--- a/backend/packages/Upgrade/src/api/services/ExperimentService.ts
+++ b/backend/packages/Upgrade/src/api/services/ExperimentService.ts
@@ -184,28 +184,7 @@ export class ExperimentService {
 
     let queryBuilderToReturn = this.experimentRepository
       .createQueryBuilder('experiment')
-      .leftJoinAndSelect('experiment.conditions', 'conditions')
       .leftJoinAndSelect('experiment.partitions', 'partitions')
-      .leftJoinAndSelect('experiment.queries', 'queries')
-      .leftJoinAndSelect('experiment.factors', 'factors')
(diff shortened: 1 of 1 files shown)


bcb37 (approved):
From a backend standpoint, this looks fine to me. As long as there are no concerns about the frontend receiving a much simpler data structure than before, this is a good change.

PR #2896, opened 2026-01-27
Default to the Data tab for experiments that have started

### LINKED ISSUE #2892: Default to the Data tab for experiments that have started
The experiment details page currently opens on the Design tab by default.

Once an experiment has started, the primary user goal is to monitor enrollment and metrics, not to review the setup. For this reason, experiments that have started (Running, Paused, Completed) should default to the Data tab, while experiments that have never started (Inactive) or archived should continue to default to the Design tab.

The tab order and navigation remain unchanged. Only the initially selected tab varies based on experiment state.

---

Update: I closed the PR I opened since this seems like something we may want to discuss more.

zackcl:
Resolves #2892

I know this wasn’t planned for 6.3, but it’s something we discussed a long time ago and felt closer to a small UX refinement than a new feature. I think defaulting to the Data tab once an experiment has started better matches the primary user intent for this page.

I’ve tested it locally and it behaves as expected. The changes are relatively small and straightforward, but if this feels risky to include this late (or if you dislike this new behavior), I’m totally fine deferring it.

**Note**: the changes in `experiment-query-result.component.ts` only add a couple of null-safety guards and are not related to the UX change in this PR.


zackcl:
Since this seems like something we may want to discuss more, I’m happy to close this for now.
We can revisit it later if it feels right.

PR #2901, opened 2026-01-28
Merge latest release 6 2

danoswaltCL:


bcb37:
All but the last 3 commits are already in dev. Is there a way to do this more cleanly?

danoswaltCL:
@bcb37 well now with the dead-code clean-up and other merges this commit history is going to look even funnier, even though I'm positive only the last 2 client lib commits really matter.

I'm going to go ahead just cherry-pick these commits into dev this time around so it's clear what we're merging in.

PR #2917, opened 2026-01-30
Show snackbar when updating a flag's Include/Exclude Lists

### LINKED ISSUE #2824: Snackbar doesn't appear when updating a flag's Include/Exclude Lists
A snackbar notification should appear when a feature flag is updated.

zackcl:
Resolves #2824


danoswaltCL (approved):

danoswaltCL:
i wish the `notification.service.ts` did the translate so we could just pass the key and not have to bring the translate module and do it every spot... but that's for another day.

PR #2928, opened 2026-01-30
Prevent showing all metrics when experiment context has no defined metrics

### LINKED ISSUE #2921: Bug: Add Metric modal shows metrics from all contexts if a context has no metrics defined
6.3.0

**Describe the bug**
The Metric field in "Add Metric" modal shows metric options from all contexts if the experiment's context has no metrics defined.

**Expected behavior**
The Metric field shouldn't show any options to select if the experiment's context has no available metrics to use.

zackcl:
Resolves #2921

Changes:

Modified `filterMetricsByAssignmentContext()` to enforce strict context-based filtering:
- Returns empty array when context has no matching metrics (instead of showing all metrics)
- Returns empty array when context information is missing/loading (instead of showing all metrics)  
- Applies context filtering to both regular and within-subjects experiments

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-metric-modal/upsert-metric-modal.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-metric-modal/upsert-metric-modal.component.ts
@@ -772,14 +772,15 @@ export class UpsertMetricModalComponent implements OnInit, OnDestroy {
       filteredMetrics = metrics.filter((metric) => metric.children && metric.children.length > 0);
     }
 
-    // Then, apply context filtering if context is available
-    if (this.currentContext?.length) {
-      filteredMetrics = filteredMetrics.filter(
-        (metric) => metric.context && this.currentContext?.some((ctx) => metric.context.includes(ctx))
-      );
+    // Context is required - if not available, return empty array
+    if (!this.currentContext?.length) {
+      return [];
     }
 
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-metric-modal/upsert-metric-modal.component.ts
```diff
@@ -765,21 +765,22 @@ export class UpsertMetricModalComponent implements OnInit, OnDestroy {
...
     }
 
-    if (this.currentAssignmentUnit && this.currentContext?.length) {
-      const contextFilteredMetrics = metrics.filter(
-        (metric) => metric.context && this.currentContext?.some((ctx) => metric.context.includes(ctx))
-      );
-      if (contextFilteredMetrics.length > 0) {
-        return contextFilteredMetrics;
-      }
+    // Context is required - if not available, return empty array
```
danoswaltCL: nitpick but can we move this up in the function so our return-early conditions are clear and grouped together? or marry them together?

```
if (!metrics?.length || !this.currentContext?.length) {
      return [];
    }
```

zackcl: Done! Thanks for the suggestion.

danoswaltCL (approved):

PR #2931, opened 2026-02-02
Fix stratifications table not loading on page refresh

### LINKED ISSUE #2930: Refreshing the Participants page clears Stratifications table
6.3.0

**Describe the bug**
Refreshing the Participants page clears Stratifications table

zackcl:
Resolves #2930

## Problem
Stratifications table was empty after refreshing the Participants page while the Preview Users tab is temporarily disabled.

## Solution
- Changed `ngOnInit` to call `selectedTabChange({index: 1})` to match current tab structure
- Added sequential data loading: fetch experiment names first, then stratifications after names are loaded (otherwise, an error occurs) using `filter` and `take(1)`

## Changes
- `experiment-users-root.component.ts`: Sequential fetch logic for stratifications tab
- Child component remains unchanged (subscribe-only pattern)

## Revert Instructions
When re-enabling Preview Users tab:
1. Uncomment tab in HTML
2. Change `index: 1` back to `index: 0` in ngOnInit

### DIFF
--- a/frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
+++ b/frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
@@ -2,6 +2,7 @@ import { Component } from '@angular/core';
 import { ExperimentService } from '../../../../core/experiments/experiments.service';
 import { PreviewUsersService } from '../../../../core/preview-users/preview-users.service';
 import { StratificationFactorsService } from '../../../../core/stratification-factors/stratification-factors.service';
+import { filter, take } from 'rxjs/operators';
 
 @Component({
   selector: 'app-user-root',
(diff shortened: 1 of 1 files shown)


frontend/projects/upgrade/src/app/features/dashboard/experiment-users/experiment-users-root/experiment-users-root.component.ts
```diff
@@ -17,15 +18,25 @@ export class ExperimentUsersRootComponent {
...
+    this.selectedTabChange({ index: 1 });
   }
 
   selectedTabChange(event) {
     if (event.index === 0) {
       this.experimentService.fetchAllExperimentNames();
       this.previewUsersService.fetchPreviewUsers(true);
     } else if (event.index === 1) {
-      this.stratificationFactorsService.fetchStratificationFactors(true);
+      this.experimentService.fetchAllExperimentNames();
```
danoswaltCL, line 31: hm well the data flow for this page and its tabs is a bit convoluted isn't it. some things are loaded here for no reason I can see, some things are loaded in other places...

really either this parent component should pass the data down into the child component or the child component should do this work itself and wait...

but i pulled this down to see if that would be simple to do and it's kind of a mess in that component. so i suppose this is sensible, I don't know if it's worth troubling too much with these older components we don't use a lot I guess.

danoswaltCL (approved):

PR #2980, opened 2026-02-26
tabbed url, refresh stays on tab, refactor tabbed footer logic across components

danoswaltCL:
Adds `&tab=` to url when selecting a "common-tabbed-footer"-based tab.

Refresh will use the query param to open on that tab.

There was a lot of variance in how different components did tabbing, I found that I could tuck all tabbing logic into the common-tabbed-footer itself to standardize it.

PLUS because the feature components no longer need to manage tabbing I could eliminate a bunch of obsolete code 👍 👍 👍 👍 👍 👍


bcb37 (commented):
This seems to work. Are we putting off adding this to the 'Profile' and 'Metrics' tabs on the profile page?

danoswaltCL:
meh, should we? It wasn't mentioned so I didn't bother for the older infrequently visited components, but I probably should have at least asked. i can ask in stand-up.

danoswaltCL:
I'l just add it for those also, why not, will update

danoswaltCL:
@bcb37 alright this will now work on the profile page tabs also, ready for re-review

bcb37 (approved):

Issue #2955, opened 2026-02-09
mooclet: directly query reward counts, instead of 'posteriors'


danoswaltCL:
requires a little research, the `posteriors` object on the `policyParameters` is how we've been getting the reward count, but it re-renders on the _next_ "assign" call to mooclet instead of when a reward is actually logged, which has been awkward.

Instead we can try to get an immediate count by querying the Mooclet API directly (probably via the `values` table). Hopefully this can be accomplished without creating an endpoint in the mooclet code, but even that may be okay.

PR #2994, opened 2026-03-04
Feature/rewards summary direct query

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/issues/2955

this adds functionality to directly query the mooclet "value" table to count rewards instead of using the currentPosteriors, which is always stale and a little confusing in the UI. 

The display remains the same, and we are able to use existing infrastructure to do so without making changes to the mooclet code. 👍


packages/backend/src/api/services/MoocletRewardsService.ts
```diff
@@ -204,6 +213,74 @@ export class MoocletRewardsService {
...
+
+    return await this.moocletDataService.getRewardsForExperiment(requestBody, logger);
+  }
+
+  public async createExperimentRewardsSummary(
+    moocletExperimentRef: MoocletExperimentRef,
+    moocletRewardsResponse: MoocletPaginatedResponse<MoocletValueResponseDetails>,
+    logger: UpgradeLogger
+  ): Promise<ExperimentRewardsSummary> {
+    const rewards: MoocletValueResponseDetails[] = moocletRewardsResponse.results;
```
bcb37: Based on the type, this is a paginated response? Could it ever be possible that the results in `moocletRewardsResponse.results` will be an incomplete data set of only the first [page_length] results?

bcb37 (commented):
This looks good and seems to work. Just one question.

packages/backend/src/api/services/MoocletRewardsService.ts
danoswaltCL: i'll re-look at this, I was thinking I could configure the mooclet server to ignore page max limitations but that doesn't seem to work the way i hoped. i guess will need to have it iterate until all are fetched.

danoswaltCL: i will push up a commit to iterate through if there are more results than the max page size. Also, the max page size is configurable in Django so I'm going to make it pretty big (it's at 100 currently, 10000 seems good, why not. I checked and nothing else is likely to be affected by this as we wouldn't need limiting by pages for anything we're doing).

danoswaltCL:
@bcb37 I've updated to iterate.

packages/backend/src/api/services/MoocletDataService.ts
```diff
@@ -236,13 +236,23 @@ export class MoocletDataService {
 
   public async getRewardsForExperiment(
     requestBody: MoocletRewardCountRequestBody,
-    logger: UpgradeLogger
+    logger: UpgradeLogger,
+    nextPageUrl?: string
   ): Promise<MoocletPaginatedResponse<MoocletValueResponseDetails>> {
-    const endpoint = `/value?mooclet=${requestBody.moocletId}&variable__name=${requestBody.variableName}`;
+    let url = '';
```
bcb37: Ignorable nitpick: This could be a constant, since its value is fully determined by the `nextPageUrl` parameter.
`const url = nextPageUrl ?? this.apiUrl + endpoint`

danoswaltCL: like this, yeah?

```
const url = nextPageUrl ?? `${this.apiUrl}/value?mooclet=${requestBody.moocletId}&variable__name=${requestBody.variableName}`
```

danoswaltCL: actually `||` is probably better than `??` also now that i look at it, which is even nitpickier (switches on all falsey values, not just null or undefined)

```
const url = nextPageUrl || `${this.apiUrl}/value?mooclet=${requestBody.moocletId}&variable__name=${requestBody.variableName}`
```

bcb37 (dismissed):
This looks good.

PR #2996, opened 2026-03-04
disable the weights editing and display n/a for mooclet experiments c…

danoswaltCL:
adds N/A for weights in the conditions sections card (was already there for the condition enrollments data tab)

adds a tooltip in both places we show condition weights when N/A because mooclet

image

image

image


bcb37 (dismissed):
This looks ok. 'N/A' is probably meaningful enough without having to explain that the mooclet algorithm is handling condition assignment?

danoswaltCL:
fyi @bcb37 @zackcl i updated this one since it was approved, added tooltip as discussed earlier.

bcb37 (approved):

Issue #2979, opened 2026-02-25
Adding IDs in include/exclude list reorders them while adding


amurphy-cl:
UI reorders values as they're being added to an include/exclude list. 

Go to include list or exclude list. Add list, select type, begin to add strings separated by commas. If you save and open the list again, and/or delete a list member from the middle, the list reorders in the UI to be alphabetical rather than the order the user originally entered the values. 

Reordering will also happen when the user switches focus e.g. moves to a different browser tab while the modal is open and then moves back (e.g. when copying schoolids from a google sheet into the modal). 

<img width="654" height="548" alt="Image" src="https://github.com/user-attachments/assets/7e3039cd-ce94-4c75-aee5-bc4f24215b54" />

PR #3001, opened 2026-03-05
remove unneccessary and bug-inducing extra sorting in list tags

### LINKED ISSUE #2989: Deleting value from include or exclude list is broken
When using the add or edit list modal, and selecting the X to delete a value, the value is not deleted AND the list is re-ordered so it's not clear whether anything is deleted, or the wrong value is deleted, or nothing at all is happening except for reordering. 

* After testing, it looks like when the value is on the last line or possibly the first line, the value will get deleted, but if the value is in the middle, it is reordered. Probably requires more testing to get to the bottom of the issue.

danoswaltCL:
fixes #2989 and #2979


bcb37 (commented):
For long lists I can see why we'd want to sort them when looking at a list that's already been created (either via the tooltip or the edit modal). What was the bug?
Oh, I see on the issue.

bcb37 (approved):

PR #3017, opened 2026-03-11
convert to new ng template syntax

danoswaltCL:
ran this command: `ng generate @angular/core:control-flow --path=proje…cts/upgrade/src/app`


danoswaltCL:
i'll close this one, too many PRs coming fast and furious, this command is takes 2 seconds to run so let's do it at the very end after everything else is in.

PR #3020, opened 2026-03-12
add state time logs download action

bcb37:


packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
```diff
@@ -197,6 +210,43 @@ export class ExperimentOverviewDetailsSectionCardComponent implements OnInit, On
...
 
+  downloadStateChangeLogs(name: string, stateTimeLogs: ExperimentStateTimeLog[]) {
+    const formattedLogs = stateTimeLogs
+      .map((log) => ({
+        timeLog: new Date(log.timeLog).toLocaleString(),
+        fromState: log.fromState,
+        toState: log.toState,
+      }))
+      .sort((a, b) => new Date(a.timeLog).getTime() - new Date(b.timeLog).getTime()); // Sort logs by time
+    this.downloadLogsAsCSV(formattedLogs, `${name}_state_change_logs_${new Date().toISOString()}`);
```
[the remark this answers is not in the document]
bcb37: Browser sanitation is fine in this case, since we don't need consistency in filenames.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
```diff
@@ -197,6 +209,42 @@ export class ExperimentOverviewDetailsSectionCardComponent implements OnInit, On
     );
   }
 
+  downloadStateChangeLogs(name: string, stateTimeLogs: ExperimentStateTimeLog[]) {
+    const sortedLogs = [...stateTimeLogs].sort((a, b) => new Date(a.timeLog).getTime() - new Date(b.timeLog).getTime());
```
danoswaltCL: do we need to sort the logs? would they ever really arrive out of order?

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
```diff
@@ -197,6 +209,42 @@ export class ExperimentOverviewDetailsSectionCardComponent implements OnInit, On
...
+
+  private downloadLogsAsCSV(values: Partial<ExperimentStateTimeLog>[], fileName: string): void {
+    if (!values || values.length === 0) {
+      return;
+    }
+
+    this.commonExportHelpersService.downloadValuesAsCSV(this.objectsToCsvRows(values), fileName);
+  }
+
+  private objectsToCsvRows(values: object[]): string[] {
```
danoswaltCL: `object` is a type to avoid, like `any`, it doesn't work as expected... any time any ol' unknown "object" is valid `Record<string, unknown>` is what I've heard is the best typesafe thing.

however... we do know what the type is here, it's`ExperimentStateTimeLog[]`, this isn't any ol' unknown object, this is doing a lot of work that IMO can be safely simplified.

I think something like this essentially does the same thing? The one difference is that it would allow unexpected values, safely wrapped in quotes to make them strings. IMO that would be preferable to mysterious blank cells via coercing to "", plus it's really clear what's going on.

```typescript
downloadStateChangeLogs(name: string, stateTimeLogs: ExperimentStateTimeLog[]) {
    if (!stateTimeLogs.length) return; // could throw an explicit error, this would be weird? 

    const headerRow = 'timeLog,fromState,toState';
    const sortedLogs = [...stateTimeLogs].sort((a, b) => new Date(a.timeLog).getTime() - new Date(b.timeLog).getTime()); // if sorting is really needed, if not this gets even simpler.
    const dataRows = sortedLogs.map(({ timeLog, fromState, toState }) => {
      const formattedTime = new Date(timeLog).toLocaleString();
      return `"${formattedTime}","${fromState}","${toState}"`;
    });
    const csvRows = [headerRow, ...dataRows];

    this.commonExportHelpersService.downloadValuesAsCSV(
      csvRows,
      `${name}_state_change_logs_${new Date().toISOString()}`
    );
  }
```

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
```diff
@@ -179,6 +188,9 @@ export class ExperimentOverviewDetailsSectionCardComponent implements OnInit, On
       case EXPERIMENT_DETAILS_PAGE_ACTIONS.EMAIL_DATA:
         this.openConfirmEmailDataModal(experiment.id, experiment.name);
         break;
+      case EXPERIMENT_DETAILS_PAGE_ACTIONS.EXPORT_STATE_CHANGE_LOGS:
+        this.downloadStateChangeLogs(experiment.name, experiment.stateTimeLogs || []);
```
danoswaltCL, line 192: for what it's worth, downloading the design or emailing data menu functions open a confirm modal. why? i don't know why, but it's what every other menu item in the new design does, they all at least open an are-you-sure, so for consistency we may want to do that here also.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
bcb37: yess we need to sort them because `timelog` is not a deafult sorting field. they showed up out of order a couple times while i was testing

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
bcb37: @amurphy-cl confirmed in the meeting yesterday that we don't want a modal

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-overview-details-section-card/experiment-overview-details-section-card.component.ts
bcb37: I was going to change that to a `Record<>` type. Missed it. Good catch. Will do now

bcb37: Also: a) I wanted to keep the `objectsToCsvRows` generic so that it could potentially handle any object, and so it's not brittle with respect to possible changes in the log format (or if we want to add other log fields to the download); and b) I think it makes sense to separate the method that performs the conversion from the one that does the downloading.

danoswaltCL: ok if it's really for a generic object then can it be a utility of the export helper service? i think that's what i'm snagged on, in this scoped component context we'll be confident of the type so I didn't know why we'd sweat over the trickerations of converting a generic one.

bcb37: That's probably unnecessary. We can use yours.

danoswaltCL (approved):

PR #3021, opened 2026-03-13
remove files accidentally missed in post-restructure merge

danoswaltCL:
re-remove the old frontend directory and the artifacts from merging workplace restructure into existing branch that had new components that the workplace restructure didn't know about.


bcb37 (approved):
This gets rid of the frontend directory and also `experiment-log-section-card-component`. I noticed that after this change, the UI flickers and jump when you switch to the logs tab, while before it just scrolled to the top (which the other tabs don't do) Is this intended behavior?

danoswaltCL:
@bcb37 Well there are two `experiment-log-section-card-component`s on dev at the moment because one is real and in the right place, and the component being deleted _here_ is a ghost copy, it is was an artifact from merging the restructure into this pre-existing branch, it viewed these as new components in new files because they didn't exist in the restructure branch, which caused it to recreate a nearly-empty `frontend` directory where the old one used to be as if it was new in this branch, with just these two items in it.

Which is a long way to say that whatever you're observing couldn't be caused by this particular change, these files are not part of the build.

PR #3031, opened 2026-03-17
yarn for precommit hook

danoswaltCL:
noticed linter wasn't running on precommit

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,7 +1,7 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+yarn lint-staged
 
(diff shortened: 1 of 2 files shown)


bcb37 (commented):
Odd: the npx command lints on pre-commit for me, but this yarn one doesn't

danoswaltCL:
so if you are on dev, and you make a change to both a frontend file and a backend file that the linter should catch, you don't encounter any issues? it detects both changes and autofixes them?

bcb37:
> so if you are on dev, and you make a change to both a frontend file and a backend file that the linter should catch, you don't encounter any issues? it detects both changes and autofixes them?

Yeah, just tested it again. The yarn works but command doesn't error on an empty commit.

PR #3041, opened 2026-03-20
remove UI support for legacy segments

bcb37:


danoswaltCL:
yep this looks right, Claude and i found a few more things that can be removed too, these ngrx actions and effects and things can go, they are only used by the legacy components:

actionfetchAllSegments (action)
actionFetchSegmentsSuccessLegacyGetAll (action)
fetchAllSegments$ (effect)
fetchAllSegments() method on the service
Corresponding reducer handler for actionFetchSegmentsSuccessLegacyGetAll
Tests for fetchAllSegments$ in effects.spec.ts and reducer.spec.tsf

danoswaltCL (approved):

PR #3044, opened 2026-03-25
Fix/normalize experiment shapes for audit log diffs

danoswaltCL:
backend: "normalizes" the difference between old and new experiment objects when creating a diff for audit logs so we see only the data we care about changing in the update log diffs.

frontend: turns the diff into a side-by-side comparison

this diff-creating business is really ugly stuff in the backend, and this adds to it, but the root cause is much deeper and I'm not getting into that right now.


bcb37 (approved):
lgtm

PR #3050, opened 2026-03-27
Fix/audit logs page as common timeline component

danoswaltCL:
updates logs page to a version with the new common page layout and timeline section card with global logs configuration

refactors the separate log-timeline configurations a bit to keep things tidy

deletes the old logs page, which was a lot of stuff.

image


danoswaltCL:
@bcb37 found some changes that need made, i put in draft mode, will put back in review mode when ready

danoswaltCL:
@bcb37 ok it's back open for review

bcb37 (commented):
The code reorganization seems ok, from what I can tell. But did we want to remove the name of the experiment or feature flag on the main logs page? It used to tell which experiment or feature flag had been created/updated and the name was a link to the details page. Now it says, e.g. "Successfully updated experiment", without telling which experiment, but you can still look at the change diff of the unnamed experiment that had been updated.

danoswaltCL:
no good call, i missed that, definitely want to keep that part, will update

danoswaltCL:
@bcb37 this is ready, links to experiments and flags restored in the main logs page

packages/frontend/projects/upgrade/src/app/shared-standalone-component-lib/components/common-audit-log-timeline/common-audit-log-timeline-config.model.ts
```diff
@@ -89,6 +89,27 @@ export interface AuditLogTimelineConfig {
...
+  /**
+   * Extracts the entity name (e.g., experiment name, feature flag name) from log data.
+   * When provided, the name is displayed after the action message in the log entry.
+   * On context-specific timelines (e.g., the experiment detail page) this should return null
+   * since the entity name is already implied by the page context.
+   *
+   * @param logData - The log's data object
+   * @returns The entity name string, or null if not applicable
+   */
+  getEntityName?: (logData: any) => string | null;
```
bcb37, line 102: The log data has been typed `any`, but we know at least that it's an object, the backend response is actually typed `object`, which I think we're trying to avoid. It's been that way for a while, maybe it's ok for now.

bcb37 (approved):

PR #3055, opened 2026-04-02
update dev to version 6.5

danoswaltCL:


packages/backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "upgrade-backend",
-    "version": "6.3.0",
```
bcb37, line 3: We'll have to be careful to keep this when we merge the hotfixes back from 6.4, since this version never got set to 6.4 on dev.

bcb37 (approved):
We'll have to be careful to keep this when w merge the hotfixes from 6.4 back, since the backend version never got set to 6.4 in dev.

Issue #3056, opened 2026-04-06
mooclet bug on /assign causes error instead of assigning null


danoswaltCL:
This is pre-existing in production. If there is a mooclet-related error while doing /assign, we are not gracefully handling that error, it is causing /assign to return as an error altogether.

Expected: `/assign` will not error when fetching assignments, unless 404 user not found. if it can't get a particular experiment's assignment, it should warn and return `null` for the assignment and move on to finding other experiment assignments.

Actual: `/assign` can fail during a mooclet call, which throws an error for /assign up to the client which effectively blocks ALL assignments where a user is eligible for multiple experiments within the given context. This was observed when a mooclet experiment that had bad data in staging.

The fix should be to log a loud warning when this happens and why, but it shouldn't allow 'assign' to error, it should treat same as "default".

PR #3061, opened 2026-04-07
log mooclet assignment errors but do not throw, treat as default

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/issues/3056

This adds logic to properly log a clear error but not throw it if the mooclet server is unreachable, or if there data corruption finding mooclet-related resources.

This patches up a couple of spots to ensure that we will treat this error consistently and return "undefined" for this decision point, leaving it out of the /assign return response.

To test the error, create an experiment and then delete the condition maps / experiment refs and try to /assign for a user, with at least one other eligible experiment running that should return a valid assignment. That should error currently in the release branch, but on this branch it will return valid experiments, and an error should be seen in the logs.

### DIFF
--- a/packages/backend/src/api/services/ExperimentAssignmentService.ts
+++ b/packages/backend/src/api/services/ExperimentAssignmentService.ts
@@ -2000,7 +2000,7 @@ export class ExperimentAssignmentService {
         message: 'Error getting condition from Mooclet proxy; experiment will return no condition for this user',
         experimentId: experiment.id,
         userId,
-        err,
+        error: err,
       });
       return undefined;
     }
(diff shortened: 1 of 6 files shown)


bcb37 (commented):
This works based on the testing criteria specified. It will prevent client errors on /assign calls when the mooclet server is unreachable or the mooclet data in the UpGrade database doesn't correspond to what's on the mooclet server. It looks like that's the only place in the client code where the mooclet data service in called.
Aside from the co-pilot suggestions, this looks ok to me.

danoswaltCL:
@bcb37 ready for re-review, i just implemented some of the copilot suggestions

bcb37 (approved):
looks good

PR #3083, opened 2026-04-22
add feature flag exposure data tab component

### LINKED ISSUE #3032: Enable Data tab for feature flags
Currently the Data tab in feature flags is greyed out. We should be able to see the number of exposure for a given feature flag without navigating back to the list page. 

An MVP Exposure Data card can be essentially a copy of the Enrollment Data card. It should read Exposure Data and track exposures (aka assigned flags, not _experienced_ flags) by date. 

Image

bcb37:


danoswaltCL:
looks good, my only suggestion is to create a helper service for the duplicate stuff, i created a suggestion commit while playing around with this that would do that, just centralizes some of the functions so they can be shared: https://github.com/CarnegieLearningWeb/UpGrade/commit/f74df3753e1e98ecfba43e9d2149b0107935b33d

bcb37:
> looks good, my only suggestion is to create a helper service for the duplicate stuff, i created a suggestion commit while playing around with this that would do that, just centralizes some of the functions so they can be shared: [f74df37](https://github.com/CarnegieLearningWeb/UpGrade/commit/f74df3753e1e98ecfba43e9d2149b0107935b33d)

Cool, looks good. I'll add it in. The frontend code was all Claude, so I can continue the hands-off approach.

danoswaltCL (approved):

PR #3091, opened 2026-04-29
CICD-532 Use trusted publishing for upgrade_client_lib instead of token

mfugate1:
I set up the publishing account on npmjs.com to allow trusted publishing from the clientlibs-js-publish.yml workflow. I'm sticking with GitHub Actions because there isn't a way for me to use trusted publishing with Jenkins.


mfugate1:
@danoswaltCL It would be nice to test this, but I'm not sure we want to publish a new package version just for that.

danoswaltCL (approved):
works for me, we'll have an opportunity to publish a new version soon enough, will try it then.

PR #3104, opened 2026-05-07
remove data manipulation of table lists from frontend

### LINKED ISSUE #3102: Move all sorting and filtering to the backend
We've set up sorting and filtering for experiment, feature flag, and segments lists to be done on the backend. But it's still being done redundantly in the frontend, which is unnecessary and could result in unpredictable behavior.

- For all sorting and filtering interactions on the three root pages (experiments, feature flags, and segments), make sure that: 
- a) a backend API call is made with the new sort/filter parameters
- b) the list display is replaced by the data from that call
- c) no frontend manipulation of the data happens

- Pagination should continue to work smoothly, without any reordering of the data
- Archived experiments should still be filtered out of results unless the 'Status' filter is selected
- All adding/editing/deleting/importing should continue to work

bcb37:


packages/frontend/projects/upgrade/src/app/features/dashboard/segments/pages/segment-root-page/segment-root-page-content/segment-root-section-card/segment-root-section-card-table/segment-root-section-card-table.component.ts
```diff
@@ -84,11 +85,7 @@ export class SegmentRootSectionCardTableComponent implements OnInit {
...
 
     this.observer = new IntersectionObserver((entries) => {
-      // if the list is short, this will trigger again on page load because the bottom trigger is within range already
-      // if the size of the filtered set is less than the number of segments to take though, fetching more will not be needed
-      const isFilteredSetLessThanTake = this.dataSource$?.filteredData?.length < NUMBER_OF_SEGMENTS;
-
-      if (entries[0].isIntersecting && !isFilteredSetLessThanTake) {
+      if (entries[0].isIntersecting) {
         this.fetchSegmentsOnScroll();
       }
```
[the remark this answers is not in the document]
bcb37: This is not reproducible. It seems that `switchMap` prevents more than one fetch from completing.

packages/frontend/projects/upgrade/src/app/core/segments/store/segments.reducer.ts
```diff
@@ -86,9 +83,10 @@ const reducer = createReducer(
     SegmentsActions.actionAddSegmentSuccess,
     (state) => ({ ...state, isLoadingSegments: false })
   ),
-  on(SegmentsActions.actionUpsertSegmentSuccess, (state, { segment }) =>
-    adapter.upsertOne(segment, { ...state, isLoadingSegments: false })
-  ),
+  on(SegmentsActions.actionUpsertSegmentSuccess, (state, { segment }) => ({
+    ...state,
+    isLoadingSegments: false,
+  })),
```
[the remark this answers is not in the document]
bcb37: Not relevant because editing and list display are separate now.

danoswaltCL:
double checking, is it right for it show archived when status is first selected (or when no status typed yet)? It looks like it does that in QA also, so may not be related to this change.

image

packages/frontend/projects/upgrade/src/app/core/segments/store/segments.reducer.ts
```diff
@@ -186,84 +199,99 @@ const globalReducer = createReducer(
...
-  ),
+  on(SegmentsActions.actionFetchGlobalSegmentsSuccess, (state, { globalSegments }) => ({
+    ...state,
+    segments: globalSegments,
+    isLoadingSegments: false,
+  })),
   on(SegmentsActions.actionGetSegmentByIdSuccess, (state, { segment }) => {
     if (segment.type === SEGMENT_TYPE.GLOBAL_EXCLUDE) {
-      return adapter.upsertOne(segment, { ...state, isLoadingSegments: false });
+      return { ...state, isLoadingSegments: false };
```
danoswaltCL: both return paths are the same now actually, looks like it can just be `return { ...state, isLoadingSegments: false }`

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
```diff
@@ -232,24 +243,24 @@ const reducer = createReducer(
...
     const { featureFlag } = listResponse;
-    const existingFlag = state.entities[featureFlag?.id];
 
-    if (existingFlag) {
-      const updatedExclusions =
-        existingFlag.featureFlagSegmentExclusion?.map((exclusion) =>
-          exclusion.segment.id === listResponse.segment.id ? listResponse : exclusion
-        ) ?? [];
+    // Update selectedFlag if it matches the modified flag
+    const updatedSelectedFlag =
```
danoswaltCL, line 248: && state.selectedFlag not needed

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
```diff
@@ -232,24 +243,24 @@ const reducer = createReducer(
...
+      state.selectedFlag?.id === featureFlag?.id && state.selectedFlag
```
danoswaltCL, line 249: can this and the experiment equivalent be refactored a little to avoid a chain of operators? claude suggests something like:

```typescript
if (state.selectedFlag?.id !== featureFlag?.id) {
  return { ...state, isLoadingUpsertPrivateSegmentList: false };
}
const updatedInclusions = state.selectedFlag.featureFlagSegmentInclusion.map((item) =>
  item.segment.id === listResponse.segment.id ? listResponse : item
);
return {
  ...state,
  selectedFlag: { ...state.selectedFlag, featureFlagSegmentInclusion: updatedInclusions },
  isLoadingUpsertPrivateSegmentList: false,
};
```

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-root-page/experiment-root-page-content/experiment-root-section-card/experiment-root-section-card-table/experiment-root-section-card-table.component.ts
```diff
@@ -61,7 +61,7 @@ export class ExperimentRootSectionCardTableComponent implements OnInit {
   constructor(private readonly experimentService: ExperimentService) {}
 
   ngOnInit() {
```
danoswaltCL: we can just remove these lifecycle hooks if we're not using them, they don't have to be there

packages/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
```diff
@@ -59,15 +59,15 @@ export class FeatureFlagRootSectionCardTableComponent implements OnInit {
   constructor(private featureFlagsService: FeatureFlagsService) {}
 
   ngOnInit() {
```
danoswaltCL: can be removed

packages/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
```diff
@@ -145,19 +137,21 @@ export class FeatureFlagRootSectionCardTableComponent implements OnInit {
...
     if (event.direction) {
+      // Make backend call with new sort parameters
       this.featureFlagsService.setSortingType(event.direction.toUpperCase());
       this.featureFlagsService.setSortKey(event.active);
+      this.featureFlagsService.fetchFeatureFlags(true); // true = reset pagination
     } else {
       // When sorting is cleared, revert to default sorting
       this.featureFlagsService.setSortingType(null);
       this.featureFlagsService.setSortKey(null);
+      this.featureFlagsService.fetchFeatureFlags(true); // true = reset pagination
```
danoswaltCL, line 141: what if we updated this fetch pattern globally to explain the mysterious "true" param of these fetches from `this.featureFlagsService.fetchFeatureFlags(true)` to `this.featureFlagsService.fetchFeatureFlags({ fromStart: true })` so its obvious and we don't need the comment? doesn't have to be in this pr, i can put something up if you agree.

packages/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card.component.html
```diff
@@ -31,7 +31,7 @@
     } @else {
     <div class="table-container">
       <app-feature-flag-root-section-card-table
-        [dataSource$]="dataSource$ | async"
+        [featureFlags$]="featureFlags$"
```
danoswaltCL, line 34: it's not async anymore? or was this pipe always not needed?

danoswaltCL:
this is fascinating. i wouldn't have thought to remove the 'adapter' things, i never really thought about why we needed it but yeah i guess if this works and we always refetch... i guess they didn't really do anything useful...

my comments are mostly nitty things, i think everything looks to be working but I'd like to play around a little more tomorrow to see if i can break it :)

bcb37:
> double checking, is it right for it show archived when status is first selected (or when no status typed yet)? It looks like it does that in QA also, so may not be related to this change.
> 
> image

Yeah, that's the current (desired) behavior.

bcb37:
> this is fascinating. i wouldn't have thought to remove the 'adapter' things, i never really thought about why we needed it but yeah i guess if this works and we always refetch... i guess they didn't really do anything useful...
> 
> my comments are mostly nitty things, i think everything looks to be working but I'd like to play around a little more tomorrow to see if i can break it :)

Yeah, It's a separation of concerns thing. The adapters, and the 'entities' they worked with, were there to make sure that edits to the experiments, etc. were reflected in the lists being maintained in the state. I found that the adapter/entity pattern can control the order of lists in ways that are not so obvious. But the lists should always reflect what we're getting from the backend. Since edits/adds/deletes are always performed on the details page, and we refetch when returning to the lists, it made sense to rely only on the backend for the state of the lists, but continue using the UI-based state updates only for the details pages.

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
bcb37: The above suggestion doesn't do quite the same thing; for instance, it will try and fail to dereference `state.selectedFlag.featureFlagSegmentInclusion` if both `state.selectedFlag?.id` and `featureFlag?.id` evaluate to null or undefined. There are many places where this reducer idiom is used. Would we want them all to be changed? Maybe a refactor ticket?

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
bcb37: Woudn't it evaluate to true if they're both undefined?

packages/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card-table/feature-flag-root-section-card-table.component.ts
bcb37: Makes sense to me. Maybe not for this pr

packages/frontend/projects/upgrade/src/app/features/dashboard/feature-flags/pages/feature-flag-root-page/feature-flag-root-page-content/feature-flag-root-section-card/feature-flag-root-section-card.component.html
bcb37: The data doesn't need to be unwrapped in the card component anymore, so the observable is just passed through and the 'async' was moved to the table component.

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
danoswaltCL: ok, my brain read it backwards i guess.

packages/frontend/projects/upgrade/src/app/core/feature-flags/store/feature-flags.reducer.ts
danoswaltCL: if you don't want to that's fine i'll move on

danoswaltCL (approved):

Issue #491, opened 2022-08-03
Refactor term 'Partition' out of the code


danoswaltCL:
This should be updated to mean decision-point, there are 200 places in the code in various ways where `partition` is used, which will be confusing if not taken care of. Will make sure to do in small chunks to keep from code conflicts, esp with renaming files, etc.

zackcl:
@danoswaltCL I've added the Acceptance Criteria to your post. Please check if that makes sense to you.

danoswaltCL:
Sure, acceptance criteria for tech debt is a little awkward because usually an independent "tester" can't validate it, has to be the developers.

zackcl:
I need to do this.

zackcl:
Since this will make changes to so many files, I will work on this once our current QA tasks (including SRS) are mostly done.

PR #3111, opened 2026-05-11
allow target to be empty; update displays

### LINKED ISSUE #2513: Make target optional
We discussed making the target optional so that it could be used only when it makes sense.

Further clarifications needed for the changes in the UI and the backend.

UI changes
- Add Decision Point Modal: Site would be default field, with button or arrow dropdown to add optional Target
- Decision Point table - remove target column, only show when target is defined in the Add Decision Point modal
- Data tab - remove Target when shown 

Backend changes
- Change types 
- Target may still be a required field in the DB, possibly replace with empty string when unused? May be easier than defining a rule for decision points that use targets and those that don't use targets
- Refactoring (we still have DPs called "Partition" in several places)

We need a design for "stacking" the site and target in the displays - like on the payload card. Site, with target in parentheses. Apply this design everywhere we display DP with site & target. Truncate or wrap?

Image

Image

bcb37:


packages/backend/src/api/services/AnalyticsService.ts
```diff
@@ -356,7 +356,7 @@ export class AnalyticsService {
           DesignType: row.designType,
           AlgorithmType: row.algorithmType,
           Stratification: stratification,
-          Site: row.site,
+          ['Site (Decision Point)']: row.site,
```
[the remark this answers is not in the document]
bcb37: I wasn't sure about this. Will ask.

packages/backend/src/api/services/AnalyticsService.ts
```diff
@@ -342,6 +342,8 @@ export class AnalyticsService {
         const excludeIfReached = row.excludeIfReached ? 'TRUE' : 'FALSE';
         const stratification =
           row.stratification && row.stratificationValue ? `${row.stratification}: ${row.stratificationValue}` : 'NA';
+        const target = row.target ? ` (${row.target})` : '';
+        const decisionPoint = row.site ? row.site + target : 'NA';
```
danoswaltCL: if site isn't defined, that would be weird here right? i think we'd want to throw an error or something, a decisionPoint can't be 'NA'

packages/frontend/projects/upgrade/src/app/core/experiments/store/experiments.model.ts
```diff
@@ -54,7 +54,7 @@ export interface ExperimentConditionFilterOptions {
 export interface ExperimentPartitionFilterOptions {
   id: string;
   point: string;
-  twoCharacterId: string;
+  target?: string;
```
danoswaltCL: but it's not really optional as far as types go, it should be guaranteed to be a string when it comes from the backend, whether its an empty string or not. If it can be undefined then we're still gonna have to do extraneous type-checking.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-decision-points-section-card/experiment-decision-points-table/experiment-decision-points-table.component.ts
```diff
@@ -40,15 +40,22 @@ export class ExperimentDecisionPointsTableComponent {
...
   DECISION_POINT_TRANSLATION_KEYS = {
-    SITE: 'experiments.details.decision-points.site.text',
-    TARGET: 'experiments.details.decision-points.target.text',
+    DECISION_POINT: 'experiments.details.decision-points.decision-point.text',
     EXCLUDE_IF_REACHED: 'experiments.details.decision-points.exclude-if-reached.text',
     ACTIONS: 'experiments.details.decision-points.actions.text',
   };
 
+  getDecisionPoint(decisionPoint: ExperimentDecisionPoint): string {
+    if (!decisionPoint) {
```
danoswaltCL: how can decisionPoint be undefined? that would be an error?

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-decision-points-section-card/experiment-decision-points-table/experiment-decision-points-table.component.ts
```diff
@@ -40,15 +40,22 @@ export class ExperimentDecisionPointsTableComponent {
...
+      return '';
+    }
+    const targetDisplay = decisionPoint.target ? ` (${decisionPoint.target})` : '';
```
danoswaltCL: this formatting is implemented multiple times in separate components, maybe a shared helper method for all of these?

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-enrollment-data-section-card/enrollment-condition-table/enrollment-condition-table.component.ts
```diff
@@ -73,11 +72,9 @@ export class EnrollmentConditionTableComponent implements OnChanges, OnInit, OnD
     });
   }
 
-  getPartitionData(partitionId: string, key: string) {
-    return this.experiment.partitions.reduce(
-      (acc, partition) => (partition.id === partitionId ? (acc = partition[key]) : acc),
-      null
-    );
+  getPartitionData(partitionId: string) {
```
danoswaltCL: can we update to "DecisionPoint" instead of keeping Partition around?

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-enrollment-data-section-card/enrollment-over-time/enrollment-over-time.component.ts
```diff
@@ -310,6 +310,10 @@ export class EnrollmentOverTimeComponent implements OnChanges, OnInit, OnDestroy
     );
   }
 
+  getPartitionDisplayText(partition: ExperimentPartitionFilterOptions): string {
```
danoswaltCL: can we update to "DecisionPoint" here and in some other places instead of keeping Partition around? I know it says partition all over this file but it's already inconsistent throughout the frontend since `experiment.partitions` is present where we otherwise speak of decisionPoints, I don't want to add new code using the word partition if it can be avoided.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-payloads-section-card/experiment-payloads-table/experiment-payloads-table.component.ts
```diff
@@ -51,16 +51,13 @@ export class ExperimentPayloadsTableComponent {
...
+  getDecisionPoint(decisionPointId: string): string {
     const decisionPoint = this.partitions.find((dp) => dp.id === decisionPointId);
-    return decisionPoint?.site || '';
-  }
-
-  // Get target name from decision point ID
-  getTarget(decisionPointId: string): string {
-    const decisionPoint = this.partitions.find((dp) => dp.id === decisionPointId);
-    return decisionPoint?.target || '';
+    if (!decisionPoint) {
```
danoswaltCL: here also, if there's no decision point that's an error right? what will returning an empty string do?

danoswaltCL:
Decision point formatted `site (target)` looks good. Should this say "Target (optional)" as the placeholder?

image

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-decision-point-modal/upsert-decision-point-modal.component.ts
```diff
@@ -190,15 +190,15 @@ export class UpsertDecisionPointModalComponent implements OnInit, OnDestroy {
     // Check if this decision point already exists
     const isDuplicate = currentExperiment.partitions.some((decisionPoint) => {
       const isSameSite = decisionPoint.site?.trim() === site;
```
danoswaltCL: isn't it guaranteed that this is a string?

danoswaltCL (dismissed):
is there any downside in making the target column non-nullable and defaulting to empty string? So we truly have an airtight expectation that site and target are both strings? if either one is undefined then there's a problem somewhere else, i think we should be throwing an error in those cases.

packages/backend/src/api/services/AnalyticsService.ts
bcb37: Yeah, I guess it can't

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-enrollment-data-section-card/enrollment-condition-table/enrollment-condition-table.component.ts
bcb37: I'll change what I added that uses that. There's an [issue](https://github.com/CarnegieLearningWeb/UpGrade/issues/491) for the wider refactor.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/modals/upsert-decision-point-modal/upsert-decision-point-modal.component.ts
bcb37: yeah

bcb37:
> is there any downside in making the target column non-nullable and defaulting to empty string? So we truly have an airtight expectation that site and target are both strings? if either one is undefined then there's a problem somewhere else, i think we should be throwing an error in those cases.

It seems like it would be safe to do that schema migration for us; I verified that there's no null targets on any of the CL deployments. However, it's possible to have one, for instance if you import an experiment with any of the targets removed (just verified that on QA). For now, we can make sure that the API always sends a string, with class-transformer and normalization guards for cases where the types are inferred directly from db queries. I'll add those.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-enrollment-data-section-card/enrollment-condition-table/enrollment-condition-table.component.ts
danoswaltCL: ha yeah i know that's an ancient one that the old team sort of did it but never finished the job.

danoswaltCL:
so no to the db migration? last time i bring it up, i promise 😬. last pitch: the only place we're allowing an unhelpful null value is in the db itself, and the only complexity of this change is the extra code to ideal with that possibility. I don't see any obstacle to safely migrating, could easily backfill "" where null.

it seems more complicated to preserve it than remove, i guess is what i'm saying.

bcb37:
> so no to the db migration? last time i bring it up, i promise 😬. last pitch: the only place we're allowing an unhelpful null value is in the db itself, and the only complexity of this change is the extra code to ideal with that possibility. I don't see any obstacle to safely migrating, could easily backfill "" where null.
> 
> it seems more complicated to preserve it than remove, i guess is what i'm saying.

Ok, I've come around. I'll do the migration. I was actually on the fence for a while myself.

danoswaltCL:
cool, ty ty

bcb37:
> cool, ty ty

We'll have to note this as a breaking change for the release, but I've verified that all 3 CL environments will be fine, since there are no null targets.

danoswaltCL (approved):

PR #3117, opened 2026-05-14
add decision point while paused, pendingActivation property added

danoswaltCL:
- Adds `pendingActivation` property to decision points
- Migration updates existing decision points based on experiment state (INACTIVE get set to true, all else true)
- Backend updates `pendingActivation` DPs to `false` when an experiment is set to RUNNING
- Imported experiment decision points will be set to `pendingActivation=true`
- Adds UI checks for disabled form based on `pendingActivation`, allows adding decision point while paused.


packages/backend/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -883,6 +883,10 @@ export class ExperimentAssignmentService {
     return experimentsToAssign;
   }
 
+  private getActivePartitions(experiment: Experiment): DecisionPoint[] {
+    return experiment.partitions.filter((dp) => !dp.pendingActivation);
+  }
```
bcb37, line 888: Do we want to be adding new methods that use the old 'partitions' terminology?

danoswaltCL: :) no, we don't

packages/backend/src/api/services/ExperimentService.ts
```diff
@@ -908,7 +912,7 @@ export class ExperimentService {
                 })
               );
               decisionPoint.id = decisionPoint.id || crypto.randomUUID();
-              return { ...decisionPoint, experiment: experimentDoc };
+              return { ...decisionPoint, experiment: experimentDoc, pendingActivation: true };
             })) ||
```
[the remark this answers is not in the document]
danoswaltCL: this is worth adding, i'm updating and doing some tests

packages/backend/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -883,6 +883,10 @@ export class ExperimentAssignmentService {
     return experimentsToAssign;
   }
 
+  private getActivePartitions(experiment: Experiment): DecisionPoint[] {
+    return experiment.partitions.filter((dp) => !dp.pendingActivation);
```
bcb37: This is removing experiments from pools even when they are in preview mode and the user is a preview user, which is inconsistent with the test a few lines lower in `mapDecisionPoints`. That's why the preview user integration tests weren't passing. If this is fixed we can add them back in and support the feature.

packages/backend/src/api/services/ExperimentService.ts
```diff
@@ -1347,6 +1348,9 @@ export class ExperimentService {
           : [];
 
       // creating decision point docs
+      // DPs are pending unless the experiment starts directly in an enrolling state
+      const internalCreationState = EXPERIMENT_STATE_INTERNAL_NAME_OVERRIDES[experiment.state] || experiment.state;
+      const startsPending = internalCreationState !== EXPERIMENT_STATE.ENROLLING;
```
bcb37, line 1372: If an experiment is created directly in COMPLETED ('cancelled'), pendingActivation will be true, which is inconsistent. Even though editing is disallowed in the frontend for 'cancelled' experiments, we may rely in the future on pendingActivation only being true when the dps are actually pending activation.

bcb37 (changes requested):
A couple more things.

packages/backend/src/api/services/ExperimentAssignmentService.ts
danoswaltCL: i thought we were done with preview mode

packages/backend/src/api/services/ExperimentService.ts
danoswaltCL: if an experiment is created in any state but enrolling, the decision point is pending activation, because it has never been activated in an enrolling state. that seems consistent to me.

packages/backend/src/api/services/ExperimentAssignmentService.ts
bcb37: I knew we were hiding it. should we make an issue to remove it completely?

danoswaltCL: yeah, but in meantime i fixed this up as suggested, ready for re-review

bcb37 (approved):

PR #3151, opened 2026-06-05
Pin Postgres image version in Docker Compose files

### LINKED ISSUE #3150: Update Docker Compose for Postgres 18
The Docker Compose files currently use the floating `postgres:alpine` image tag, which can resolve to newer Postgres major versions without notice.

Since QA and staging are using Postgres 18, the local and demo Docker Compose files should also pin to `postgres:18-alpine`.

Postgres 18 Docker images also use a different recommended data directory layout, so the volume mount should be updated from `/var/lib/postgresql/data` to `/var/lib/postgresql`.

zackcl:
Resolves #3150

Pins the Postgres image to `postgres:18-alpine` and updates the Docker volume mount to `/var/lib/postgresql`, which matches the recommended layout for Postgres 18 Docker images.

### DIFF
--- a/demo.singleContainerApp-docker-compose.yml
+++ b/demo.singleContainerApp-docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - "postgres"
 
   postgres:
-    image: postgres:alpine
+    image: postgres:16-alpine
     restart: always
     environment:
       POSTGRES_USER: postgres
--- a/demo.singleContainerApp-docker-compose.yml
+++ b/demo.singleContainerApp-docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - "postgres"
 
   postgres:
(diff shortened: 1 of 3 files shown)


bcb37:
The version of postgres that's now deployed on QA and staging (and soon prod) is 18, so that should probably be the one we use as a default.

zackcl:
Thanks, that makes sense. I updated the PR to use `postgres:18-alpine` and changed the volume mount to `/var/lib/postgresql` for the Postgres 18 Docker image layout.

danoswaltCL:
no 16 was right, it's version 16.13 that we're using in RDS

danoswaltCL:
although hey if 18 works fine, i don't know if that's a bad thing

danoswaltCL (approved):

Issue #3157, opened 2026-06-08
bug: `createBaseQueryBuilder` is a performance dragon that must be slayed ⚔️ 🐉


danoswaltCL:
So i think i've found a monster...while going spelunking this way and that trying to track down why [this experiment](https://apps.qa-cli.net/upgrade/#/home/detail/24f2a66b-4366-46b0-addc-c18498c480a3) is not loading correctly in QA, I found that `findOneExperiment` called in `/experiment/:id` was to blame because within it we call `createBaseQueryBuilder` which does one query with exponentially more expensive joins the more decision points, conditions, and segments are involved.

Then looking around I see that we also use `createBaseQueryBuilder` in csv export and /batch-assign`.... it's not quite as bad per experiment fetch because it is only for one decision point which eliminates one factor, but still not good and it's for every valid experiment in the context.

Claude helped me break this down into something concrete I can understand: The formula is essentially as shown below, this would be the # of rows the query ends up creating.
- D = number of decision points
- C = number of conditions
- I and E = number of segments in inclusion list and exclusion lists one row deep (aka not every nested segment in this particular call, just the root segments to be resolved, but that's not much of a comfort based on discussions we've had about `getSegmentById` issues. Let's leave that out of this but assume it's making things slower also).

when cache is missed and hits the db:
- /experiment/:id will be (C² × D² × I × E) times slower than it should be
- data-export and /batch-assign) will be (C² × D × I × E) times slower than it should be (for every experiment fetched)

so for express mode, we would be getting (3^2 * 46^2) = 19044 rows at least, multiplied by roughly number of segments involved also, which is almost entirely thrown away by the logic.

in my testing, I am trying to split up these into paralell calls, it is making a gigantic difference, it becomes as quick as just (C x D) (or roughly I x E), which is (138 rows). Everything goes so much faster, including running tests.

danoswaltCL:
https://github.com/CarnegieLearningWeb/UpGrade/compare/fix/experiment-by-id-slow-sql?expand=1

pure claude answer to how to make this better for `/experiment:id` for example of what I think we need to do in general, need similar for the /assign call also.

danoswaltCL:
and some sql with explanation here, based on an experiment in local with 46 dps and 3 conditions: 

[biggun.json](https://github.com/user-attachments/files/28723003/biggun.json)

```sql
-- biggun experiment:
-- 0ea4b511-bc5a-48d3-b43e-9a3b11a14a4d
[...]

PR #3166, opened 2026-06-09
split up monolithic createBaseQueryBuilder experiment query

danoswaltCL:
#3157

`createBaseQueryBuilder` broken up into individual calls, same as other calls. This affects `/experiments/:id`, csv data export, and batch-assign. 

after creating individual queries, we end up with identical query builder implementations in several places in ExperimentRepository, so these were refactored in general as private entity builder functions. one of the implementations was only different in that a query to leftJoin stratification data was apparently missing, so i took the liberty of adding that.

also also, got rid of the word partition where i could in the process.

btw `getEnrollingExperimentsForContextAndDecisionPoint` was dead code, i'm guess a missed merge conflict resolution artifact.


danoswaltCL:
@copilot please make the recommended label updates.

danoswaltCL:
@bcb37 this is ready for review, i addressed copilot's comments.

bcb37 (approved):

Issue #3144, opened 2026-06-02
Feature flag optimizations chapter one


danoswaltCL:
#3143 do 1a, 1b, and 4a from the research spike plan that claude and I spun up.

bcb37:
No regressions found. Load testing should verify performance improvement.

PR #3169, opened 2026-06-10
Fix/cache wrap function fixup

danoswaltCL:
this is 4b from https://github.com/CarnegieLearningWeb/UpGrade/issues/3144.

this is a lower-order thing and can go in 6.6, just putting it up now cuz i have it


danoswaltCL:
i'll just close this for now, it doesn't seem as important now if we're mostly bypassing the `resolveSegments` flow.

PR #3178, opened 2026-06-11
Add Decision Point to experiment filter dropdown

### LINKED ISSUE #3159: Decision Point Filter in Experiment Root
As a user I want to be able to compare experiments where there are decision points that might be shared. Or as a user I simply want to find an experiment by decision point.

Add an explicit `DECISION POINT` filter into the dropdown to make it super clear that you can do so, substring combobox will filter just on decision point site and target strings.

This currently does technically happen within the "all" search but you kinda gotta know, and even there you'd get extra stuff if the site or target being looked up contained the same substring as other experiment property strings, so this would only look at site and target.

bcb37:


danoswaltCL:
i'm fine with plunking this into 6.5 personally

danoswaltCL (approved):

PR #3195, opened 2026-06-25
remove twoCharacterId from codebase

### LINKED ISSUE #585: Remove twoCharacterId
**Describe the bug**
We've decided to get rid of the `twoCharacterId` property from the experiment data structure as they are not being used.

So a decision point (partition) and a condition shouldn't have the `twoCharacterId` field.

### LINKED ISSUE #1244: Remove twoCharacterId from database
linked issue #585. When the twoCharacterId is removed from the experiment structure, we should also remove it from the database.

bcb37:


danoswaltCL (approved):
looks good to me, good riddance

Issue #3191, opened 2026-06-23
precomputed segments


danoswaltCL:
## Background

Segment resolution is the bottleneck in the `/assign` and `/featureflag` endpoints. Every request must fetch all related segments for each flag and experiment, then recursively flatten them into inclusion/exclusion sets on-the-fly. This is expensive: segment reads are constant, segment writes are rare, and a cache miss on any large segment set can cause significant DB load.

The core inefficiency: we repeat expensive flattening work on every read, when that work only needs to happen after a write.

## Proposal

Move segment flattening to write-time. When a segment's membership changes, compute and persist flat ID lists immediately. Reads at assignment time become simple Set lookups against pre-built arrays — no recursion, no large joins. This applies equally to feature flags and experiments.

---

### Storage

1. Precomputed flat arrays (`inclusionIds`, `exclusionIds`) are stored per flag and per experiment. This may be implemented as two separate tables.
2. Each row is keyed by its parent flag or experiment ID (FK with `onDelete: CASCADE`). A row is created when a flag or experiment is created and deleted automatically when it is deleted.
3. The flat arrays contain all IDs (individual IDs, group IDs, etc.) drawn from the full segment graph — including members of nested sub-segments, recursively resolved.

### Triggering recomputation

4. **On any segment list add, update, or remove on a flag or experiment:** recompute that flag's or experiment's precomputed row.
5. **On any segment-level change** (members added/updated/removed on a shared segment): recompute all flags and experiments that reference that segment or any of its ancestors.
6. **On segment deletion:** collect the set of affected flag and experiment IDs *before* the delete (join table records will be gone afterward), then recompute those rows *after* the delete completes.
7. **If a segment is not referenced by any flag or experiment:** there is nothing to recompute — return early.
8. CRUD operations on lists should do the recompute in the background, i.e. a UI user should get their response for success and not need to wait for the recompute part.

### Startup backfill

9. On server startup, any flag or experiment missing a precomputed row should be backfilled. Flags and experiments that already have a row should be skipped (no-op).

---

## Assignment Algorithm

Replace the current on-the-fly segment resolution with the following. This algorithm applies identically to both feature flags (`/featureflag`) and experiments (`/assign`).

**For each flag or experiment being evaluated:**

### INCLUDE_ALL
> Everyone is included unless explicitly excluded.

1. Fetch the precomputed exclusion list for this flag or experiment.
2. Scan it against all of the user's contextual IDs (individual + all group IDs).
[...]

PR #3198, opened 2026-06-26
Wip/precomputed segments for flags

danoswaltCL:
requirements largely found here: https://github.com/CarnegieLearningWeb/UpGrade/issues/3191

note that this is just for the flags at this moment, as its somewhat more straightforward, the main mechanics will be the same for both.

This creates a new table for precomputed segments, which will get updated on segment list changes.
`getKeys()` will now do very little to snag flags.

I have this going to the old flow still in the event of some error finding the precomputed segment, which hopefully would be really hard to do.


packages/backend/src/api/services/FeatureFlagService.ts
```diff
@@ -859,23 +912,88 @@ export class FeatureFlagService {
...
-      const excludeIds = getEnabledSegmentIds(flag.featureFlagSegmentExclusion);
-      let includeIds = [];
+      const exclusionSet = new Set(computed.exclusionIds);
+      const inclusionSet = new Set(computed.inclusionIds);
+
+      // Individual exclusion always wins
+      if (exclusionSet.has(experimentUser.id)) return false;
+
+      // Individual inclusion bypasses group checks
+      if (inclusionSet.has(experimentUser.id)) return true;
```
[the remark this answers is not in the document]
danoswaltCL: this is actually the pre-existing logic, it contradicts the spec i wrote in https://github.com/CarnegieLearningWeb/UpGrade/issues/3191 because the pre-existing behavior surprisingly does do this: https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1782498104248719

packages/types/src/Experiment/enums.ts
```diff
@@ -337,9 +336,9 @@ export enum SUPPORTED_CALIPER_EVENTS {
 export enum CACHE_PREFIX {
   EXPERIMENT_KEY_PREFIX = 'validExperiments-',
   SEGMENT_KEY_PREFIX = 'segments-',
-  GLOBAL_EXCLUDE_SEGMENT_KEY_PREFIX = 'globalExcludeSegment-',
   MARK_KEY_PREFIX = 'markExperiments-',
   FEATURE_FLAG_KEY_PREFIX = 'featureFlags-',
+  PRECOMPUTED_SEGMENT_KEY_PREFIX = 'precomputedSegments-',
```
bcb37: This, and other places that have the 'precomputedSegments' naming should be changed if there's going to be a separate entity for experiment precomputed segments, separate from feature flag ones. Include 'featureFlag' in the name to distinguish it and make it clear that it's an entity at the same level.

packages/backend/src/api/models/FeatureFlagPrecomputedSegment.ts
```diff
@@ -0,0 +1,19 @@
...
+import { FeatureFlag } from './FeatureFlag';
+
+@Entity()
+export class PrecomputedSegment extends BaseModel {
+  @PrimaryColumn('uuid')
+  public featureFlagId: string;
+
+  @ManyToOne(() => FeatureFlag, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'featureFlagId' })
+  public featureFlag: FeatureFlag;
```
bcb37, line 12: This entity should have "featureFlag" in the name if it's specific to feature flag-related segments.

packages/backend/src/api/models/FeatureFlagPrecomputedSegment.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @Column('text', { array: true, default: '{}' })
+  public inclusionIds: string[];
```
bcb37, line 15: These id lists are not distinguished by groupId and userId. If there's ever a case where a groupId and a userId are identical, the name clash could result in unexpected behavior.

danoswaltCL: will handle in follow-up w/ experiment lists, as this is an edge-case

danoswaltCL:
@copilot please review [8193311](https://github.com/CarnegieLearningWeb/UpGrade/pull/3198/commits/8193311f7cfc57f7a46edf261a8015fee8e0169f), which addresses a couple comments about grouptype information

packages/backend/src/api/services/FeatureFlagService.ts
zackcl, line 510: When the context changes, we delete the existing flag lists here, but I don’t see the precomputed row getting recomputed or cleared after the transaction commits. Could this leave `feature_flag_precomputed_segment` with stale inclusion/exclusion IDs from the deleted lists?

packages/backend/src/api/services/FeatureFlagService.ts
```diff
@@ -648,15 +683,26 @@ export class FeatureFlagService {
...
-      // Create a new transaction if no entity manager is provided
-      return await this.dataSource.transaction(async (manager) => {
+      result = await this.dataSource.transaction(async (manager) => {
         return await executeTransaction(manager);
       });
+
+      // Recompute precomputed sets for each affected flag after the transaction commits
+      const affectedFlagIds = [...new Set(listsInput.map((l) => l.id))];
+      await Promise.all(
+        affectedFlagIds.map((flagId) => this.featureFlagPrecomputedSegmentService.recomputeForFlag(flagId, logger))
```
zackcl: Should this post-commit recompute be fire-and-forget instead of awaited? The issue calls out that list write responses shouldn’t wait on recomputation, and this same pattern appears in `deleteList`, `updateList`, `importFeatureFlags`, and `importFeatureFlagLists`.

packages/backend/src/api/services/FeatureFlagPrecomputedSegmentService.ts
```diff
@@ -0,0 +1,200 @@
...
+      .where('segment.id IN (:...ids)', { ids: unresolved })
+      .getMany();
+
+    const ids: string[] = [];
+    const subSegmentIds: string[] = [];
+
+    for (const segment of segments) {
+      segment.individualForSegment.forEach((ind) => ids.push(ind.userId));
+      segment.groupForSegment.forEach((grp) => ids.push(grp.groupId));
+      segment.subSegments.forEach((sub) => {
```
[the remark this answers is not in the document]
danoswaltCL: applied something like this

packages/backend/src/api/services/FeatureFlagService.ts
```diff
@@ -859,23 +912,88 @@ export class FeatureFlagService {
...
+    experimentUser: ExperimentUser,
+    context: string,
+    logger: UpgradeLogger
+  ): Promise<Pick<FeatureFlag, 'id' | 'key' | 'filterMode'>[]> {
+    const flagIds = featureFlags.map((f) => f.id);
+    const precomputedMap = await this.featureFlagPrecomputedSegmentService.getPrecomputedSets(flagIds);
+
+    // Flatten all group IDs from the user's group map (type is ignored per design decision)
+    const userGroupIds: string[] = experimentUser.group ? Object.values(experimentUser.group).flat() : [];
+
```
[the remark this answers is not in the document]
danoswaltCL: applied something like this

packages/backend/src/api/services/FeatureFlagService.ts
danoswaltCL: yeah, i had thought copilot was misreading things but it looks like in some spots it's inconsistent. the recompute isn't really very slow in even the worst spots but i think i've got an idea for a small change to make this easy to keep consistent.

danoswaltCL: i've applied a change that will make sure each change is consistent in how it does recompute.

packages/backend/src/api/services/FeatureFlagService.ts
danoswaltCL: this path should be included now

packages/backend/src/api/services/FeatureFlagService.ts
zackcl: Thanks, the helper looks good for update/delete/addList. I may be missing an intentional exception here, but I still see the import paths awaiting recompute: `importFeatureFlags` and `importFeatureFlagLists` both call `await recomputeForFlag(...)` after commit. Should these use the same fire-and-forget scheduling path too, so import responses also don’t wait on recomputation?

danoswaltCL: it is actually intentional here, just because it's a bulk action vs single operations, where since literally all the resources are being updated and created, waiting feels natural here. and it would somewhat complicate composing this helper function. there's a note called out in the claude.md about this, but a clearer comment here makes sense so i pushed that up.

danoswaltCL:
@zackcl i pushed up merge conflict resolution w/ the other branch, this is ready for re-review. see my other comment about why i chose to let import be awaited vs the other ones, it's just a little more complex of an operation, so more would go wrong if the precompute didn't work right when adding multiple lists in bulk, and would be less likely to be able 'self-correct' and recover during other edits.

If that's okay then this is good to go.

zackcl (dismissed):

zackcl (dismissed):

PR #3232, opened 2026-07-14
Add experiment usage counts to the Decision Points table

### LINKED ISSUE #3230: Show competing experiment counts for decision points
## Description
On the experiment details page, decision points are shown in the Decision Points table as `site (target)`.

Add a **Used By** column after **Decision Point** to show how many experiments use each decision point.

## Expected behavior
- The Decision Points table includes a **Used By** column after **Decision Point**.
- Each row displays the number of experiments using that decision point.
- The count uses singular and plural labels as appropriate, e.g. `1 Experiment` or `2 Experiments`.
- The displayed count matches the number of experiments returned by the existing Decision Point filter for the same `site (target)` value.

## Notes
- The count is informational only and does not need to be clickable.

zackcl:
Resolves #3230

## Dependency

This PR is based on #3228 and should be reviewed after that PR. Once #3228 is merged, this PR will be retargeted to `dev`.

## Changes

- Add a `Used By` column to the Decision Points table.
- Display the number of non-archived experiments using each site-target pair.
- Include usage counts in experiment detail and update responses.
- Preserve current counts after decision point and experiment state updates.
- Add backend and frontend unit coverage for usage counting and display behavior.


packages/backend/src/api/repositories/DecisionPointRepository.ts
```diff
@@ -106,6 +112,40 @@ export class DecisionPointRepository extends Repository<DecisionPoint> {
...
+    return await repository
+      .createQueryBuilder('sourceDecisionPoint')
+      .select('sourceDecisionPoint.id', 'decisionPointId')
+      .addSelect('COUNT(DISTINCT usedByExperiment.id)::int', 'usedByCount')
+      .leftJoin(
+        DecisionPoint,
+        'usedByDecisionPoint',
+        `LOWER(usedByDecisionPoint.site) = LOWER(sourceDecisionPoint.site)
+          AND LOWER(COALESCE(usedByDecisionPoint.target, '')) = LOWER(COALESCE(sourceDecisionPoint.target, ''))`
+      )
```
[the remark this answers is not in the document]
zackcl: The usage count intentionally matches the same site-target pair, not every result returned by the partial-text search.

A targetless decision point such as `lesson-stream` is different from `lesson-stream (question-hint)` for assignment and pooling. Treating an empty target as a wildcard would incorrectly count different decision points, so the current exact pair matching is intentional.

packages/frontend/projects/upgrade/src/app/features/dashboard/experiments/pages/experiment-details-page/experiment-details-page-content/experiment-decision-points-section-card/experiment-decision-points-table/experiment-decision-points-table.component.ts
```diff
@@ -82,6 +83,20 @@ export class ExperimentDecisionPointsTableComponent {
     return formatDecisionPointDisplay(decisionPoint);
   }
 
+  getUsedByCountText(decisionPoint: ExperimentDecisionPoint): string {
+    const count = decisionPoint.usedByCount;
+    if (count === undefined || count === null) {
```
danoswaltCL: this leaves open other types if the API response or upstream code changes and allows something unexpected to pass in, such as any string. if we explicitly say it has to be a number, then i think that covers it completely.

`if (typeof count !== 'number') {
  return '';
}`

zackcl: Good point, updated the type check as suggested.

danoswaltCL (approved):

PR #3242, opened 2026-07-23
allow for separate cache ttl per data-type category

danoswaltCL:
creates 3 explicit ttl configurations, which are optional and can be left off:

CACHING_TTL_EXPERIMENTS
CACHING_TTL_FEATURE_FLAGS
CACHING_TTL_SEGMENTS

`CACHING_TTL` is still a global fallback if the explicit configs aren't set. this also keeps it back compatible, nothing will break. `CACHING_ENABLED` and `CACHING_MAX_KEYS` unchanged.


danoswaltCL:
i don't know why the frontend test check failed. looking...

bcb37 (approved):
This looks good to me. As long as we're not worried about setting any of the env vars to 0 resulting in the default hard-coded value of 900, as copilot pointed out. But that is a pre-existing behavior.

PR #3246, opened 2026-07-23
rotate all condition lists in an experiment on mark

### LINKED ISSUE #3239: Update client libraries to rotate conditions by experiment
In https://github.com/CarnegieLearningWeb/UpGrade/issues/3196 we changed to API to cycle through the condition stack by experiment. The client libraries still maintain a separate stack for each decision point. We should make it consistent with the new API behavior. It should:

- maintain one condition stack per experiment
- rotate the stack when any decision point in the experiment is hit
- only make an assign call when there's no cached condition stack for the experiment

bcb37:


danoswaltCL (approved):
java 🙃 . i trust you that you've got this working.

PR #3251, opened 2026-07-24
Wip/precomputed experiment segments

danoswaltCL:
this completes the work started last release with precomputed segments

- creates precomputed-experiment-segments table
- backfills on startup if experiments don't have a row in that table (should only need to happen first time the code version runs)
- tries to mirror the precomputed-feature-flag code paths as closely as possible
- creates shared services and utilities where it's exactly the same
- clearly notes where the implementation diverge due to differences in legacy code services and the fact that experiment private lists are a bit different from feature-flag private lists (they don't need enabling events)
- global exclude lists are NOT compiled with the precomputed exclude lists, left that as is


packages/backend/src/api/repositories/ExperimentRepository.ts
```diff
@@ -437,8 +436,11 @@ export class ExperimentRepository extends Repository<Experiment> {
           await repository.query(`TRUNCATE ${entity.tableName} CASCADE;`);
```
danoswaltCL, line 436: i need to check on this, this seems like an unscripted claude change that i didn't notice...

danoswaltCL:
@copilot please re-review after fixes, and please focus effort on parity between experiment and feature flag implementations. The requirements in the end are the same, so these should be as DRY as they can be to avoid duplicate code, without abstracting or refactoring existing code too much. The notable differences are that experiment private lists do not use list-level 'enable' flags, so that is a simplifier, but experiments do need to factor in global-excludes, as these should remain handled separately the precomputed segments. Some code differences that perform very similar tasks are likely due to legacy codebase; it is not worth a giant refactor to get the code working exactly the same, as long as we are decently DRY and that a human reading this can follow the same general pattern for precomputed segment behavior through the two code paths. In general also: the efficiency of the read paths is more important than efficiency of the write paths.

packages/backend/src/api/repositories/ExperimentRepository.ts
danoswaltCL: ok, while that is true, it seems unrelated but that was the problem, it was some antipattern that was getting in the way of a clean change, and going around it wouldn't been harder than fixing it. this is only something called in the demo app, and the behavior didn't change just a refactored to avoid a potential db race issue that was apparently revealed by this change while trying to unit tests. longs story short, it looks fine.

bcb37:
@danoswaltCL  Is there any reason global excludes couldn't be folded into this too? I can't remember why we evaluate those separately, but I think we should be able to calculate them along with the others, just a matter of adding the segment of type 'global_exclude' with the same context as the experiment to the exclude lists.

danoswaltCL:
the global exclude list is still a bit of a different player in the include-exclude logic:
- it's consulted first and we return early before fetching other data
- adding it to the precomputed segments would mean concatenating potentially the same list of hundreds or thousands of ids to every exclusion list in the context
- technically the experiment-level-exclusion lists have a special carve-out for when the filter is exclude_all, and a user's group is on the exclusion list, BUT the individual is explicitly included in an include list, then they are actually included (!). it's not documented anywhere that this is the case, so i have a half-written article for gitbook on inclusion-exclusion logic that i haven't put up yet, but [Ritter seems to want it in there](https://carnegielearning.slack.com/archives/C055CGCRP6Z/p1783697563156089) so I left that be for precomputed-segments. global-excludes DO universally override that logic, so I've just left global-excludes as-is.

bcb37 (approved):

Issue #3267, opened 2026-08-05
UpGrade client: clear flag cache on groups for session update


danoswaltCL:
https://carnegielearning.atlassian.net/browse/UG-170

PR #3270, opened 2026-08-06
clear flags when changing groups mid-session

danoswaltCL:
#3267 this will unblock some work clc is trying to do, the ts client lib has a method for changing the groups that get sent with feature-flags, where a different set of groups could potentially fetch different flags. it had a bug though in that it doesn't clear previously fetched flags, so it always returned the flags for the original set of groups used. this now just clears the previous flags so that we'll successfully refetch.

### DIFF
--- a/clientlibs/js/src/DataService/DataService.spec.ts
+++ b/clientlibs/js/src/DataService/DataService.spec.ts
@@ -118,6 +118,16 @@ describe('DataService', () => {
     });
   });
 
+  describe('#clearFeatureFlags', () => {
+    it('should clear previously set feature flags', () => {
+      dataService.setFeatureFlags(['testFlagKey']);
+
+      dataService.clearFeatureFlags();
+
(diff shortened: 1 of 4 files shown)


bcb37 (commented):
1. Are we not caching per user/groups? Shouldn't it be the default behavior to refetch flags when user info has changed?
2. Do we need to make this change in the Java client lib too?

danoswaltCL:
1. yes it should clear cache on feature-flag user info change, which it is not doing currently, that's what this PR fixes. this is specific to the feature-flags option that allows for sending groups directly, because it's the only spot where we expose a public method for directly changing user info after construction, and thus potentially after already performing a fetch with a different config.

2. the java client to date has not been updated for the feature-flags option to send groups, there's probably a tech debt story there to investigate library parity but not for this ticket.

bcb37 (approved):

PR #3272, opened 2026-08-07
split experiment/dp and monitored dp queries

bcb37:
times for the the two methods `updateEnrollmentExclusionDocumentsAndCheckEndingCriteria` and `getCachedExperiments` improved a bit in perftrace

### DIFF
--- a/packages/backend/src/api/services/ExperimentAssignmentService.ts
+++ b/packages/backend/src/api/services/ExperimentAssignmentService.ts
@@ -70,7 +70,7 @@ import { CacheService } from './CacheService';
 import { UserStratificationFactorRepository } from '../repositories/UserStratificationRepository';
 import { UserStratificationFactor } from '../models/UserStratificationFactor';
 import { RequestedExperimentUser } from '../controllers/validators/ExperimentUserValidator';
-import { In } from 'typeorm';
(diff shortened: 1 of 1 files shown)


packages/backend/src/api/services/ExperimentAssignmentService.ts
```diff
@@ -134,31 +134,61 @@ export class ExperimentAssignmentService {
...
-            },
+      const experiments = await this.experimentRepository.find({
+        where: { id: In(experimentIds) },
+        relationLoadStrategy: 'query',
+        relations: {
+          conditions: {
+            conditionPayloads: true,
+          },
+          partitions: true,
+          experimentSegmentInclusion: {
```
danoswaltCL, line 173: i wonder if we even need the segments data? the inclusion/exclusion logic should just now use the precomputed lists.

danoswaltCL (approved):

packages/backend/src/api/services/ExperimentAssignmentService.ts
danoswaltCL: approving though, i'm not 100% sure, would be good to follow up on though as this would really balloon up with large re-hydrated segment lists, but no very large lists exist in prod at this very moment.

PR #3280, opened 2026-08-11
use refreshThreshold in cache-manager

danoswaltCL:


danoswaltCL:
@bcb37 Found a bug, I forgot that "precomputed segments" does use the terrible "wrapFunction" method by passing in the list of "valid" experiment or flag ids.

The "wrapFunction" was already bad because it is all-or-nothing, as-in it re-refetches everything in the supplied id list if a single id on that list isn't cached currently.

It also never records when we asked the db for a record and the db said nothing exists, so it will continue to treat that id as a cache-miss even though we confirmed already that there's nothing there to find.

Now that we want to use `refreshThreshold`, we have to overhaul this function anyway, because it doesn't actually use `cache.wrap`, which is the special sauce that allows cache-manager to manage refreshing the key the way we want.

So, I rewrote it so that it uses cache.wrap and actually handles each key individually, including recording when a key is confirmed to not exist in the db.

bcb37 (approved):

PR #3286, opened 2026-08-12
Add list value filtering across experiments, feature flags, and segments

### LINKED ISSUE #3284: Support filtering experiments, feature flags, and segments by list values
## Description

Related Slack post: https://carnegielearning.slack.com/archives/CTD5UG8R2/p1785963721834729

Users currently cannot determine which experiments, feature flags, or segments contain a specific individual or group ID in their lists.

For example, searching for `capublicre7eb2qn` should return the `ENABLE_NCE_MAIN` feature flag because the value belongs to its `CA and TX Community Review Schools` exclude list.

Image

Image

## Expected Behavior

- Add a `List Value` filter to the Experiments, Feature Flags, and Segments list pages.
- Include list-value matching in the existing `All` search on each page.
- Match both individual IDs and group IDs, such as school IDs.
- Return experiments and feature flags containing the value in an attached include or exclude list.
- Return segments containing the value in one of their lists.

zackcl:
Resolves #3284

## Changes

- Add a `List Value` filter to the Experiments, Feature Flags, and Segments pages.
- Include list values in each page's `All` search.
- Support case-insensitive partial matching for individual and group IDs in direct and nested lists.
- Respect `Include All` behavior while searching applicable inclusion and exclusion lists.
- Add unit and integration coverage for list-value filtering.

### DIFF
--- a/packages/backend/src/api/services/listValueSearchHelpers.ts
+++ b/packages/backend/src/api/services/listValueSearchHelpers.ts
@@ -1,0 +1,78 @@
+import { FILTER_MODE } from 'upgrade_types';
+
+export const LIST_VALUE_SEARCH_PATTERN_PARAMETER = 'listValueSearchPattern';
+
+export type ListValueSearchTarget = 'experiment' | 'featureFlag' | 'segment';
+
+const matchingSegmentIdsQuery = `
+  WITH RECURSIVE "directListValueSegments" AS (
+    SELECT "segmentId" AS "id"
+    FROM "individual_for_segment"
+    WHERE "userId" ILIKE :${LIST_VALUE_SEARCH_PATTERN_PARAMETER} ESCAPE '\\'
+    UNION
+    SELECT "segmentId" AS "id"
+    FROM "group_for_segment"
+    WHERE "groupId" ILIKE :${LIST_VALUE_SEARCH_PATTERN_PARAMETER} ESCAPE '\\'
+  ),
+  "matchingListValueSegments" AS (
+    SELECT "id"
+    FROM "directListValueSegments"
+    UNION
+    SELECT "segmentRelation"."parentSegmentId" AS "id"
+    FROM "segment_for_segment" "segmentRelation"
+    INNER JOIN "matchingListValueSegments" "matchedSegment"
+      ON "segmentRelation"."childSegmentId" = "matchedSegment"."id"
+  )
+  SELECT "id"
+  FROM "matchingListValueSegments"
+`;
+
+const ownerSearchConfig = {
+  experiment: {
+    ownerAlias: 'experiment',
+    ownerIdColumn: 'experimentId',
+    inclusionTable: 'experiment_segment_inclusion',
+    exclusionTable: 'experiment_segment_exclusion',
+  },
+  featureFlag: {
+    ownerAlias: 'feature_flag',
+    ownerIdColumn: 'featureFlagId',
+    inclusionTable: 'feature_flag_segment_inclusion',
+    exclusionTable: 'feature_flag_segment_exclusion',
+  },
+} as const;
+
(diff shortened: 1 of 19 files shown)


packages/backend/src/api/services/listValueSearchHelpers.ts
```diff
@@ -0,0 +1,78 @@
...
+  return `EXISTS (
+    SELECT 1
+    FROM (
+      SELECT "segmentId", "${config.ownerIdColumn}" AS "ownerId",
+        TRUE AS "isInclusion"
+      FROM "${config.inclusionTable}"
+      UNION ALL
+      SELECT "segmentId", "${config.ownerIdColumn}" AS "ownerId",
+        FALSE AS "isInclusion"
+      FROM "${config.exclusionTable}"
```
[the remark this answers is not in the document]
zackcl: This is intentional. List Value search is for discovering attached list configuration, so feature flag lists remain searchable regardless of their enabled state. Assignment evaluation still ignores disabled lists, and this behavior is covered by unit and integration tests.

danoswaltCL:
@zackcl i tried running this in my local but i get 400 Bad Request errors for the /paginated call when i try to use the List Value filter and type text in

zackcl:
> @zackcl i tried running this in my local but i get 400 Bad Request errors for the /paginated call when i try to use the List Value filter and type text in

@danoswaltCL I couldn't reproduce this on the PR branch. All three paginated endpoints return 200 when the request uses `searchParams.key: "listValue"`.

I can reproduce the same 400 if the backend is still running the previous code, or if the request sends `"List Value"` instead of `"listValue"`.

Could you confirm that both the frontend and backend are running from this PR branch and that the backend was restarted or rebuilt after switching branches? If it still happens, could you share:

- Which page and `/paginated` endpoint you tested
- The request payload from the Network tab
- The full 400 response body
- The search text you entered

packages/backend/src/api/services/listValueSearchHelpers.ts
```diff
@@ -0,0 +1,83 @@
+import { FILTER_MODE } from 'upgrade_types';
+
+export const LIST_VALUE_SEARCH_PATTERN_PARAMETER = 'listValueSearchPattern';
+
+export type ListValueSearchTarget = 'experiment' | 'featureFlag' | 'segment';
+
+const matchingSegmentIdsCte = `WITH RECURSIVE "directListValueSegments" AS (
```
danoswaltCL, line 7: what does "Cte" mean?

zackcl: CTE stands for Common Table Expression. This one recursively finds directly matching segments and their parent segments.

danoswaltCL:
(no errors anymore, i blame my local)

i threw a bunch of things at this and it worked as expected each time. 100k list performed fine, weirdly nested segments where the id appeared at multiple levels give the expected response, where in the segments view it lists both segments, that makes sense i think.

one awkward thing is that it hides lists when include all is on. that makes total sense, the awkward thing is that we hide the lists in the first place, so it makes sense we make them disappear from the search results when the filter_mode of a thing is set to INCLUDE_ALL. They will also suddenly appear if when the filter mode is off... 

the one scenario I can think of where this might cause an issue would be a user is searching for an id in order to delete or edit a typo for that id globally, they could potentially miss it in the hidden private lists of INCLUDE_ALL flags/exps that were attached and used to control inclusion in include_lists before being turned on for all. maybe not important enough to do deal, but it could come up.

this looks good to me

danoswaltCL (approved):

zackcl:
I think we should get Ben Graff’s feedback before merging, since this was based on his suggestion. I’ll record and share a short video shortly (sorry for the delay) and convert the PR to draft for now.

PR #3293, opened 2026-08-19
add whitelist to validator pipeline

bcb37:


zackcl (dismissed):

kschaefe (commented):
This seems right to me, but will let a person more knowledgable of UpGrade approve.

danoswaltCL:
cool, so this strips out anything not explicitly not listed in the validators before it runs the validation, and doesn't throw a bad request error, its 200 ok? how would this be externally verified, did we have these nonsense values mirrored back in responses?

bcb37:
> cool, so this strips out anything not explicitly not listed in the validators before it runs the validation, and doesn't throw a bad request error, its 200 ok? how would this be externally verified, did we have these nonsense values mirrored back in responses?

@danoswaltCL  [Karl answered that question when I asked it on the Jira ticket](https://carnegielearning.atlassian.net/browse/UG-172?focusedCommentId=977195)

danoswaltCL (approved):

PR #3296, opened 2026-08-22
Prevent Cartesian row growth in experiment list queries

### LINKED ISSUE #3295: Prevent experiment queries from exhausting memory with large inclusion and exclusion lists
## Problem

Experiment queries currently load inclusion and exclusion list relations in the same joined query.

When an experiment has values in both lists, the database result can grow multiplicatively. For example, an experiment with approximately 2,300 inclusion values and 2,300 exclusion values can produce more than 5 million intermediate rows before TypeORM reconstructs the entities.

This can cause excessive memory usage and eventually terminate the backend with a Node.js heap out-of-memory error.

The problem was discovered while testing the List Details work in #3255, but it is an existing backend query issue and is not specific to that UI redesign.

## Expected behavior

Experiments containing large inclusion and exclusion lists should load without producing a Cartesian multiplication of their list-member rows or exhausting the backend heap.

The existing API response structure and list membership behavior should remain unchanged.

## Proposed approach

- Load experiment inclusion and exclusion segment relations in separate queries
- Merge the results by experiment ID before returning them
- Apply the same approach to all experiment repository paths that currently load both relations, including experiment details and assignment/preview queries
- Preserve the existing API contract
- Do not require a database migration

## Acceptance criteria

- Inclusion and exclusion list relations are not joined together in a way that produces `N × M` intermediate rows
- Returned experiments still contain both inclusion and exclusion list data
- Experiment details and assignment-related queries preserve their existing behavior
- Repository tests cover merging independently loaded inclusion and exclusion data
- An experiment with thousands of values in both lists can be loaded without exhausting the Node.js heap

zackcl:
Resolves #3295

## Summary

- Loads experiment inclusion and exclusion segment relations in separate database queries
- Merges the independently loaded relations by experiment ID while preserving the existing response structure
- Applies the split to experiment details, list, assignment, and preview query paths
- Adds regression coverage for independently loaded inclusion and exclusion data

## Problem

The existing experiment segment query joins inclusion and exclusion list members together. When both lists contain many values, the intermediate result grows multiplicatively as `N × M`.

For example, approximately 2,300 inclusion values and 2,300 exclusion values can produce more than 5 million intermediate rows. Materializing those rows through TypeORM can exhaust the Node.js heap and terminate the backend process.

## Implementation

The combined segment query has been replaced with separate inclusion and exclusion queries. Their results are merged by experiment ID before being combined with the remaining experiment data.

This preserves the existing API response and list membership behavior while avoiding Cartesian growth. No API contract or database schema changes are required.

## Testing

- Backend typecheck
- ESLint
- Experiment repository unit tests
- Complete backend unit test suite: 1,004 passing

### DIFF
--- a/packages/backend/src/api/repositories/ExperimentRepository.ts
+++ b/packages/backend/src/api/repositories/ExperimentRepository.ts
@@ -24,13 +24,15 @@ export class ExperimentRepository extends Repository<Experiment> {
       .addOrderBy('queries.order', 'ASC', 'NULLS LAST')
       .addOrderBy('queries.createdAt', 'ASC');
 
-    const experimentSegment = this.buildSegmentQuery();
+    const experimentInclusionSegmentQuery = this.buildInclusionSegmentQuery();
+    const experimentExclusionSegmentQuery = this.buildExclusionSegmentQuery();
 
     const [
       experimentConditionLevelPayloadData,
       experimentFactorPartitionLevelPayloadData,
       experimentMetricData,
-      experimentSegmentData,
+      experimentInclusionSegmentData,
+      experimentExclusionSegmentData,
     ] = await Promise.all([
       experimentConditionLevelPayloadQuery.getMany().catch((errorMsg: any) => {
         const errorMsgString = repositoryError(
@@ -59,16 +61,27 @@ export class ExperimentRepository extends Repository<Experiment> {
         );
         throw errorMsgString;
       }),
-      experimentSegment.getMany().catch((errorMsg: any) => {
+      experimentInclusionSegmentQuery.getMany().catch((errorMsg: any) => {
         const errorMsgString = repositoryError(
           'ExperimentRepository',
-          'findAllExperiments-experimentSegmentData',
+          'findAllExperiments-experimentInclusionSegmentData',
+          {},
+          errorMsg
+        );
+        throw errorMsgString;
+      }),
+      experimentExclusionSegmentQuery.getMany().catch((errorMsg: any) => {
+        const errorMsgString = repositoryError(
+          'ExperimentRepository',
+          'findAllExperiments-experimentExclusionSegmentData',
           {},
           errorMsg
         );
         throw errorMsgString;
       }),
     ]);
+
+    const experimentSegmentData = this.mergeSegmentData(experimentInclusionSegmentData, experimentExclusionSegmentData);
 
     const experimentData = experimentConditionLevelPayloadData.map((data) => {
       const data2 = experimentFactorPartitionLevelPayloadData.find((i) => i.id === data.id);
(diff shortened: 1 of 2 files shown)


bcb37:
For the two methods in the assignment path: `getValidExperiments()` and `getValidExperimentsForContextAndDecisionPoint()`, we may not need to fetch the entire hydrated segment data anymore, since we're getting the experiment precomputed segments (with fallback) using only the segment IDs in for both regular and batch assignments.

danoswaltCL:
Yeah is one of the things that we kicked the can on to refactor later once other things were resolved... which they are now, so fyi:

These paths still do need this change:
`findAllExperiments` GET /experiments
`findOneExperiment`  GET /experiments/:id, POST /experiments/state  PUT /experiments/:id

But I'm pretty sure these paths no longer even use the segment data being gathered, so it really shouldn't be a part of what is queried by these paths:
`getValidExperiments` (/assign)
`getValidExperimentsWithPreview`: (/assign for preview users)
`getValidExperimentsForContextAndDecisionPoint` (batch-assign)

No need optimizing what isn't needed... so we can either merge this and take care of that as a follow-up, or, if it's easy to just rearrange what is called here a little, maybe we can roll that into this?

bcb37:
> Yeah is one of the things that we kicked the can on to refactor later once other things were resolved... which they are now, so fyi:
> 
> These paths still do need this change: `findAllExperiments` GET /experiments `findOneExperiment` GET /experiments/:id, POST /experiments/state PUT /experiments/:id
> 
> But I'm pretty sure these paths no longer even use the segment data being gathered, so it really shouldn't be a part of what is queried by these paths: `getValidExperiments` (/assign) `getValidExperimentsWithPreview`: (/assign for preview users) `getValidExperimentsForContextAndDecisionPoint` (batch-assign)
> 
> No need optimizing what isn't needed... so we can either merge this and take care of that as a follow-up, or, if it's easy to just rearrange what is called here a little, maybe we can roll that into this?

@danoswaltCL @zackcl I have a relatively simple change that avoids getting the full segment data for the assign paths. I can add it as a commit here or as a separate pr.

zackcl:
@bcb37 Sure, feel free to add a commit or create a new PR.

bcb37:
> @bcb37 Sure, feel free to add a commit or create a new PR.

I'll make a separate pr. I've got a slightly more aggressive refactor that will get rid of the joins entirely on the assign path. It will not affect the queries separation in this pr for the non-assign methods.

packages/backend/src/api/repositories/ExperimentRepository.ts
```diff
@@ -502,51 +604,67 @@ export class ExperimentRepository extends Repository<Experiment> {
...
+          );
+          throw errorMsgString;
+        }),
+      ]);
 
     if (!conditionLevelPayloadData) {
       return undefined;
     }
 
+    const [segmentData] = this.mergeSegmentData(
```
danoswaltCL: is there a point to calling mergeSegmentData like this when it's only one experiment? there's nothing to process, it kinda goes in and comes right back out, you can handle the negative cases in the return object:

```
return {
      ...conditionLevelPayloadData,
      ...factorDecisionPointPayloadData,
      ...metricData,
      experimentSegmentInclusion: inclusionData?.experimentSegmentInclusion ?? [],
      experimentSegmentExclusion: exclusionData?.experimentSegmentExclusion ?? [],
    };
```

packages/backend/src/api/repositories/ExperimentRepository.ts
```diff
@@ -468,21 +543,48 @@ export class ExperimentRepository extends Repository<Experiment> {
...
+    return this.createQueryBuilder('experiment')
+      .select('experiment.id')
       .leftJoinAndSelect('experiment.experimentSegmentExclusion', 'experimentSegmentExclusion')
       .leftJoinAndSelect('experimentSegmentExclusion.segment', 'segmentExclusion')
       .leftJoinAndSelect('segmentExclusion.individualForSegment', 'individualForSegmentExclusion')
       .leftJoinAndSelect('segmentExclusion.groupForSegment', 'groupForSegmentExclusion')
       .leftJoinAndSelect('segmentExclusion.subSegments', 'subSegmentExclusion');
   }
 
+  private mergeSegmentData(inclusionData: Experiment[], exclusionData: Experiment[]): Experiment[] {
```
danoswaltCL: alright, I had to really sit with this for a while to try and understand it:

1. it would be very helpful to have some actual types, as obviously nothing here is an `Experiment[]`, so maybe we can pin something like this at the top of this file for all these kinds of partials when TypeORM is returning pieces of Experiments as if they are Experiments...

```
type SegmentInclusionFragment = Pick<Experiment, 'id' | 'experimentSegmentInclusion'>;
type SegmentExclusionFragment = Pick<Experiment, 'id' | 'experimentSegmentExclusion'>;
type SegmentFragment = SegmentInclusionFragment & SegmentExclusionFragment;
```

2. I would love it if we could deal with this without the confusing cascading spread operators and ternaries? This seems to do the same thing, covers the negative cases, and the returned object is clearly readable as the "SegmentFragment" of an Experiment

```
private mergeSegmentData(
    inclusionData: SegmentInclusionFragment[],
    exclusionData: SegmentExclusionFragment[]
  ): SegmentFragment[] {
    const inclusionById = new Map(inclusionData.map((experiment) => [experiment.id, experiment]));
    const exclusionById = new Map(exclusionData.map((experiment) => [experiment.id, experiment]));
    const experimentIds = new Set([...inclusionById.keys(), ...exclusionById.keys()]);

    return [...experimentIds].map((id) => ({
      id,
      experimentSegmentInclusion: inclusionById.get(id)?.experimentSegmentInclusion ?? [],
      experimentSegmentExclusion: exclusionById.get(id)?.experimentSegmentExclusion ?? [],
    }));
  }
```

3. If it's true that `findOne` doesn't actually need this method, and it's true that no other callers of this method will exist after @bcb37 's refactor to move /assign path stuff away from these queries, it may only really be needed by findAllExperiments.

packages/backend/src/api/repositories/ExperimentRepository.ts
zackcl: Good call. I removed the `mergeSegmentData` call from `findOneExperiment` and now merge the segment results directly with empty-array fallbacks. Addressed in bba4fc6d0c1ce1cf6d25c93dec3fecb53caa94b9.

packages/backend/src/api/repositories/ExperimentRepository.ts
zackcl: Agreed. I added the fragment types and simplified `mergeSegmentData` to return the explicit segment shape. `findOneExperiment` no longer uses it; the remaining assignment-path callers can be removed in @bcb37’s follow-up. Addressed in fbd183e904b8e0146b347f2d03b59a512c4b2ea7.

danoswaltCL (approved):
cool looks good ty for fixing this

PR #3298, opened 2026-08-26
disable deleting metrics that are in use

### LINKED ISSUE #2922: Prevent deletion of metrics that are in use by experiments
We should clearly indicate in the UI that metrics currently used by any experiment cannot be deleted.

At the moment, deletion fails silently. Instead, the delete button should be disabled (greyed out) and show a tooltip explaining the reason, for example: "This metric is currently being used by an experiment"

bcb37:


packages/frontend/projects/upgrade/src/app/core/experiments/store/experiments.effects.ts
```diff
@@ -288,9 +293,12 @@ export class ExperimentEffects {
         this.experimentDataService.deleteExperiment(experimentId).pipe(
           switchMap(() => {
             this.notificationService.showSuccess(this.translate.instant('global.delete-experiments.message.text'));
+            // Deleting an experiment also deletes its queries, which can change whether a
+            // catalog metric still has a query associated with it.
             return [
               experimentAction.actionDeleteExperimentSuccess({ experimentId }),
               experimentAction.actionFetchAllDecisionPoints(),
+              analysisActions.actionFetchMetrics(),
             ];
```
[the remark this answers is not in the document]
bcb37: The test's .pipe(take(2), pairwise()) only takes the first two emitted actions and pairs them, then unsubscribes. It never looks at a third emission at all, so the extra actionFetchMetrics() dispatch is simply not observed/asserted — it doesn't cause a mismatch. That's exactly why the spec still passes despite the additional action being dispatched.

packages/backend/src/api/repositories/QueryRepository.ts
```diff
@@ -66,4 +66,17 @@ export class QueryRepository extends Repository<Query> {
...
+      .innerJoin('query.metric', 'metric')
+      .select('DISTINCT metric.key', 'metricKey')
+      .getRawMany()
+      .catch((errorMsg: any) => {
+        const errorMsgString = repositoryError('QueryRepository', 'getMetricKeysWithQueries', {}, errorMsg);
+        throw errorMsgString;
+      });
+
+    return queryResult.map((row: { metricKey: string }) => row.metricKey);
+  }
```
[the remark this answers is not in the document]
bcb37: I've confirmed that this would result in the metrics data disappearing from the data tab and from exports, which we probably don't want.

zackcl:
It would be better to show the tooltip on the right instead of the bottom.

Screenshot 2026-08-31 at 11 39 58 AM

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
```diff
@@ -57,8 +57,16 @@
             <mat-tree-node *matTreeNodeDef="let node">
               <mat-icon class="leaf-node-icon">fiber_manual_record</mat-icon>
               {{ node.key }}
-              @if (permissions.metrics.delete && keyEditMode) {
-              <mat-icon class="delete-icon" (click)="deleteNode(node, index)"> delete_outline </mat-icon>
+              @if (permissions.metrics.delete && keyEditMode && node.isTopLevel) {
+              <mat-icon
```
zackcl, line 61: Restricting the delete action to `node.isTopLevel` changes the existing behavior by removing it from every nested metric node. For example, if one leaf under a grouped metric is in use, the root is disabled, but unused sibling branches and leaves can no longer be deleted through the UI either.

Could we preserve deletion on nested nodes and set `hasQuery` on each node based on whether that node or any descendant contains a referenced metric? That would keep referenced leaves and their ancestors disabled while allowing unused sibling nodes to be deleted.

zackcl (changes requested):

packages/backend/src/api/services/MetricService.ts
```diff
@@ -50,6 +56,14 @@ export class MetricService {
...
   public async deleteMetric(key: string, logger: UpgradeLogger): Promise<IMetricUnit[]> {
     logger.info({ message: `Delete metric by key ${key}` });
+    const experimentsUsingMetric = await this.queryRepository.getExperimentsUsingMetricKey(key, METRICS_JOIN_TEXT);
+    if (experimentsUsingMetric.length) {
+      const experimentNames = experimentsUsingMetric.map(({ name }) => name).join(', ');
+      throw new HttpError(
+        409,
+        `Metric key ${key} cannot be deleted because it is used by the following experiment(s): ${experimentNames}`
+      );
+    }
```
zackcl, line 66: This adds the expected API-level guard. One concern is that this lookup and the deletion below are separate database operations, while the existing `query.metric` foreign key uses `ON DELETE CASCADE`. If another request adds a query after this check but before the delete, that query could still be cascade-deleted along with the metric.

Could we enforce this invariant at the database level, for example with `RESTRICT`/`NO ACTION` and translate the foreign-key violation to a 409, or otherwise make the check and deletion atomic? If that is outside the scope of this PR, it may be worth tracking as a follow-up.

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
bcb37: @zackcl Deletion on nested nodes doesn't seem to work. Are you sure it ever did? I wasn't able to find the spec that would have determined the behavior it was supposed to have.

packages/backend/src/api/services/MetricService.ts
bcb37: I think that is outside the scope of this work. We're currently not expecting these delete API requests to be coming from anywhere but the frontend, which shouldn't ever request a delete for an in-use metric.

zackcl: Understood on the scope. To clarify, the frontend would not be requesting deletion of a metric that is already known to be in use. The race I had in mind would be:

1. Metric `M` is unused, so user A's frontend allows the delete request.
2. The backend lookup in this method finds no experiment query referencing `M`.
3. Before the subsequent delete executes, user B saves an experiment query referencing `M` from another frontend session, and that transaction commits.
4. User A's delete then continues, and `ON DELETE CASCADE` deletes user B's newly committed query along with `M`.

So the frontend guard correctly handles metrics that are already known to be in use, but it does not fully eliminate this narrow concurrent state change. I agree that this is a low-probability edge case and am fine with tracking the database-level guarantee as a follow-up.

bcb37: Oh I see, yeah. That is a low-probability edge case that could require a database migration. Still in agreement that it merits a follow-up, given that it's also a pre-existing possible  issue and wouldn't be a regression introduced by this pr.

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
zackcl: You're right. I tested this on `dev`, and nested deletion is currently broken. It throws an error at `key.join(...)` before sending a DELETE request.

I assumed the visible delete buttons meant nested deletion was supported, but I had not verified it. Should deletion be limited to top-level metrics? If so, the `isTopLevel` change makes sense and I'll resolve this comment. If nested deletion is intended, we should probably track the existing frontend bug separately.

packages/backend/src/api/services/MetricService.ts
zackcl: Sounds good. Thanks for confirming. I'll resolve this thread.

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.html
bcb37: > Should deletion be limited to top-level metrics? 

I'm not sure. We could, but it would be simpler not to.

bcb37: Update: restored support for nested deletion; it works with the new restriction

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
```diff
@@ -125,6 +135,9 @@ export class MetricsComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   deleteNode(nodeToBeDeleted: LazyLoadingMetric, index: number) {
+    if (nodeToBeDeleted.hasQuery) {
+      return;
+    }
     const data = {
       children: [this.allMetrics.data[index]],
```
zackcl: I found an issue when deleting a metric from filtered results. If I search for `masteryWorkspace` and try to delete a nested metric, `findParents()` returns `undefined`, so `key.join(...)` throws and no DELETE request is sent.

It looks like `index` refers to the row's position in the filtered results, while `allMetrics.data` is still the unfiltered array. Could we pass the current `element` into `deleteNode()` instead of looking it up by index?

Image

zackcl (changes requested):

packages/frontend/projects/upgrade/src/app/features/dashboard/profile/components/metrics/metrics.component.ts
zackcl, line 68: Deleting a metric while the table is filtered resets the displayed results, even though the search input still contains the filter. I reproduced this by filtering for `masteryWorkspace` and deleting a nested `timeSeconds`; `totalTimeSeconds` reappears until the filter is entered again.

It looks like recreating the `MatTableDataSource` here clears its `filter`. Could we reapply `this.searchValue` after rebuilding the data source, perhaps by moving `this.applyFilter(this.searchValue)` into this subscription?

Image

Image

zackcl (approved):

PR #3312, opened 2026-09-04
remove class-validator and rejigger types loading to shrink bundles

danoswaltCL:
- removing the need for UpgradeClient.BINARY_REWARD_VALUE as an import from upgrade_types allows for rejiggering the building of types so we can slice out `class-validator` from our bundle, which hugely reduces the bundle size (the `lite` version used by cli-services will go from 400kb to 36kb, it was really that bad)

- also removes a double-bundling of tslib which was apparently happening :shrug:

- also quicktest now has a good built-in way of testing the `lite` bundle

### DIFF
--- a/clientlibs/js/src/types/enums.ts
+++ b/clientlibs/js/src/types/enums.ts
@@ -5,4 +5,9 @@ export namespace UpGradeClientEnums {
     POST = 'POST',
     PATCH = 'PATCH',
   }
+
+  export enum BINARY_REWARD_VALUE {
+    SUCCESS = 'SUCCESS',
+    FAILURE = 'FAILURE',
+  }
 }
--- a/clientlibs/js/package.json
+++ b/clientlibs/js/package.json
@@ -50,8 +50,8 @@
     "docs:markdown": "typedoc --options typedoc.json",
     "test": "jest",
     "test:coverage": "jest --coverage",
-    "quicktest": "([ -d dist ] || yarn build) && ts-node quickTest.ts",
-    "quicktest:lite": "([ -d dist ] || yarn build) && ts-node quickTest.ts lite"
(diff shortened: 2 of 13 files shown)


clientlibs/js/src/types/enums.ts
```diff
@@ -5,4 +5,9 @@ export namespace UpGradeClientEnums {
     POST = 'POST',
     PATCH = 'PATCH',
   }
+
+  export enum BINARY_REWARD_VALUE {
```
bcb37: Is there a reason we're defining these separately in upgrade_types and for the clientlib? We don't want the values to be able to diverge, like others that we import from upgrade_types (e.g. PAYLOAD_TYPE, MARKED_DECISION_POINT_STATUS, etc.).

bcb37:
It seems unnecessary to define an enum in two places, which we'd have to keep in sync manually. The `BinaryRewardAllowedValue` enum is inert and only pulls `class-validator` along with it because it's defined in a file that imports `reflect-metadata`. If we move the inert exports to a separate file and import them where they're needed in `Mooclet/index.ts`, they can be imported directly using the deep path, like the other enums from `Experiment/...` 
Maybe it doesn't matter that much if we're getting rid of the mooclet stuff, but if we're keeping the binary rewards enum it might be a good idea to move that now, so we're not defining and maintaining it in two places going forward.

danoswaltCL:
i didn't think about the fact that the mooclet-removal already does the major work here in removing class-validator as a factor at all in upgrade_types, and it does do this enum more sensibly, and the build reflects that because it's 44kb instead 400+. 

There are a couple other smaller wins in this PR that bring it down to ~36kb, but might as well hold off on this one until the mooclet-removal pr is in.

